PR #17, opened 2017-05-14
badger: fix error return

jbowens:
Propagate any errors from lz4.Decode. Looks like the existing code 
would panic with a slice indexing error when extracting the key and
value from the decompressed data if Decode failed.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/17)


manishrjain:
Thanks for the fix.

PR #29, opened 2017-05-16
KV parameters verification.

### LINKED ISSUE #23: Errors in the API
I'm trying badger as a replacement for github.com/syndtr/goleveldb and one thing that hit me immediately is that there are no errors in the badger Go API. Any error will just exit the process through `log.Fatalf`. Is that a design decision or just temporary?

szymonm:
Resolves #23

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/29)

### DIFF
--- a/badger/kv.go
+++ b/badger/kv.go
@@ -24,6 +24,7 @@ import (
 
 	"golang.org/x/net/trace"
 
+	"errors"
 	"github.com/dgraph-io/badger/skl"
 	"github.com/dgraph-io/badger/table"
 	"github.com/dgraph-io/badger/y"
@@ -115,27 +116,20 @@ type KV struct {
 	flushChan chan flushTask // For flushing memtables.
 }
 
-type ErrWrongParameter struct {
-	Msg   string
-	Param string
-}
-
-func (e ErrWrongParameter) Error() string {
-	return fmt.Sprintf("Incorrect parameter %q: %s", e.Param, e.Msg)
-}
+var ErrInvalidDir error = errors.New("Invalid Dir, directory does not exist")
+var ErrInvalidValueLogFileSize error = errors.New("Invalid ValueLogFileSize, must be between 1MB and 1GB")
 
 // NewKV returns a new KV object.
 func NewKV(opt *Options) (out *KV, err error) {
 	dirExists, err := exists(opt.Dir)
 	if err != nil {
-		return nil, y.Wrap(ErrWrongParameter{Msg: err.Error(), Param: "Dir"})
+		return nil, y.Wrapf(err, "Invalid Dir: %q", opt.Dir)
 	}
 	if !dirExists {
-		return nil, y.Wrap(ErrWrongParameter{Msg: "directory does not exist", Param: "Dir"})
+		return nil, ErrInvalidDir
 	}
 	if !(opt.ValueLogFileSize <= 2<<30 && opt.ValueLogFileSize >= 1<<20) {
-		return nil, y.Wrap(ErrWrongParameter{Msg: "must be between 1MB and 1GB",
-			Param: "ValueLogFileSize"})
+		return nil, ErrInvalidValueLogFileSize
 	}
 	out = &KV{
 		imm:       make([]*skl.Skiplist, 0, opt.NumMemtables),
--- a/badger/kv.go
+++ b/badger/kv.go
@@ -35,7 +35,7 @@ var (
 
 // Options are params for creating DB object.
 type Options struct {
-	Dir string // Directory to store the data in.
+	Dir string // Directory to store the data in. Should exist and be writable.
 
 	// The following affect all levels of LSM tree.
 	MaxTableSize        int64 // Each table (or file) is at most this size.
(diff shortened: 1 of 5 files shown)


badger/kv.go
```diff
@@ -116,10 +116,18 @@ type KV struct {
...
-	y.AssertTrue(len(opt.Dir) > 0)
-	y.AssertTrue(opt.ValueLogFileSize <= 2<<30 && opt.ValueLogFileSize >= 1<<20)
-	out := &KV{
+func NewKV(opt *Options) (out *KV, err error) {
+	dirExists, err := exists(opt.Dir)
+	if err != nil {
+		return nil, y.Wrapf(err, "Incorrect Dir parameter")
+	}
+	if !dirExists {
+		return nil, y.Errorf("Directory '%s' does not exist.", opt.Dir)
```
dolmen: Use `%q` instead of `'%s'`

badger/kv.go
```diff
@@ -116,10 +116,18 @@ type KV struct {
...
+	}
+	if !(opt.ValueLogFileSize <= 2<<30 && opt.ValueLogFileSize >= 1<<20) {
+		return nil, y.Errorf("ValueLogFileSize should be between 1MB and 1GB")
```
dolmen: Replace `should` with `must` as you are handling this as a fatal issue.

manishrjain:
Reviewed 5 of 5 files at r1.

---

*[badger/kv.go, line 125 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/29#-KkFQyTT1aZOCb26HOab-r1-125:-KkNBNJQnNpZz3Mj5dRN:bb3nh7d) ([raw file](https://github.com/dgraph-io/badger/blob/a2ae8e78c619d8cadd9518f7bedf1c3480857622/badger/kv.go#L125)):*
<details><summary><i>Previously, dolmen (Olivier Mengué) wrote…</i></summary><blockquote>

Use `%q` instead of `'%s'`
</blockquote></details>

Address the comment above. Also, predefine this error using var, and just reuse them. That way, a user can check the error returned against the var, instead of having to do string matching.

both of these above errors can be ErrInvalidDir.

---

*[badger/kv.go, line 198 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/29#-KkNByZ1sFSEUdV14tEC:-KkNByZ1sFSEUdV14tED:bouxnm5) ([raw file](https://github.com/dgraph-io/badger/blob/a2ae8e78c619d8cadd9518f7bedf1c3480857622/badger/kv.go#L198)):*
> ```Go
> 		return true
> 	}
> 	out.vlog.Replay(vptr, fn)
> ```

Does Replay return any error? If it can, we should handle it here.

In fact, if you look carefully at all the y.Checkf, you should be able to find the ones which are being directly called from the API. Make all those return errors using errors.Wrapf and the likes. I did that just now for value log opening and closing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/29)*

szymonm:
Review status: 2 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[badger/kv.go, line 125 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/29#-KkFQyTT1aZOCb26HOab-r1-125:-KkVNDqiVSN2It9q_rrt:b84bsla) ([raw file](https://github.com/dgraph-io/badger/blob/a2ae8e78c619d8cadd9518f7bedf1c3480857622/badger/kv.go#L125)):*

Address the comment above. Also, predefine this error using var, and just reuse them. That way, a user can check the error returned against the var, instead of having to do string matching.

both of these above errors can be ErrInvalidDir.
</blockquote></details>

Added type instead of var in order to pass the error message up.
It still can be checked without string matching.

---

*[badger/kv.go, line 128 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/29#-KkFQyTT1aZOCb26HOab-r1-128:-KkVNZGmMfrr8I6Rh6PS:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/a2ae8e78c619d8cadd9518f7bedf1c3480857622/badger/kv.go#L128)):*
<details><summary><i>Previously, dolmen (Olivier Mengué) wrote…</i></summary><blockquote>

Replace `should` with `must` as you are handling this as a fatal issue.
</blockquote></details>

Done.

---

*[badger/kv.go, line 198 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/29#-KkNByZ1sFSEUdV14tEC:-KkVN_HuyfH9LZUn7RRK:bklzvef) ([raw file](https://github.com/dgraph-io/badger/blob/a2ae8e78c619d8cadd9518f7bedf1c3480857622/badger/kv.go#L198)):*

Does Replay return any error? If it can, we should handle it here.

In fact, if you look carefully at all the y.Checkf, you should be able to find the ones which are being directly called from the API. Make all those return errors using errors.Wrapf and the likes. I did that just now for value log opening and closing.
</blockquote></details>

Leaving this for you.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/29)*

manishrjain:
:lgtm: one small comment.

---

Reviewed 1 of 3 files at r2, 2 of 2 files at r3.

---

*[badger/kv.go, line 120 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/29#-KkVU8h-0FqHsssj4PUd:-KkVU8h-0FqHsssj4PUe:b-obtiq4) ([raw file](https://github.com/dgraph-io/badger/blob/3e5e1f4ef3d0faf246a08da924da944bee1263b9/badger/kv.go#L120)):*
> ```Go
> 
> var ErrInvalidDir error = errors.New("Invalid Dir, directory does not exist")
> var ErrInvalidValueLogFileSize error = errors.New("Invalid ValueLogFileSize, must be between 1MB and 1GB")
> ```

ErrValueLogSize

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/29#-:-KkVUYwYogi8wk4Y8U7U:b-e1qgtr)*

szymonm:
Merged.

PR #41, opened 2017-05-21
Use y.Printf which honors the verbose option. Fix #33

### LINKED ISSUE #33: Make logging configurable
I noticed that even if I set verbose to false, I get some noise like: "LOG Compact. Iteration to generate one table took: 110.092µs"

Badger is directly calling `fmt.Printf` which is not configurable client side.

Thanks

magic003:
Trying to use y.Printf for verbose information. It honors the verbose option from KV.

For those already using y.Printf. I removed the unnecessary verbose check.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/41)

### DIFF
--- a/badger/kv.go
+++ b/badger/kv.go
@@ -17,7 +17,6 @@
 package badger
 
 import (
-	"fmt"
 	"os"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


manishrjain:
:lgtm: Thanks for the fix. We actually need to do more work here, so that the scope of VerboseMode would be limited to KV store instance, and not be global to the process.

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/41#-:-KkhSaFrO3drBADap33n:bityffv)*

PR #42, opened 2017-05-21
Skip the head key in iterator so that its not returned to the user

### LINKED ISSUE #37: Slice bounds out of range error
Steps to reproduce

1. Compile and run dgraph on branch `badger`.
`./dgraph`

2. Load goldendata using dgraphloader.
`./dgraphloader -r ~/go/src/github.com/dgraph-io/benchmarks/data/goldendata.rdf.gz`

3. As soon as dgraphloader finishes, issue a shutdown command to server.
`curl localhost:8080/admin/shutdown`

4. Start Dgraph again.
`./dgraph`

5. Now try taking a backup.
`curl localhost:8080/admin/backup`

Dgraph panics with error
```
panic: runtime error: slice bounds out of range

goroutine 19899 [running]:
github.com/dgraph-io/dgraph/x.Parse(0xc43331498a, 0x6, 0x6, 0x2)
	/home/pawan/go/src/github.com/dgraph-io/dgraph/x/keys.go:195 +0x236
github.com/dgraph-io/dgraph/worker.backup(0x1, 0xcd16da, 0x6, 0x0, 0x0)
	/home/pawan/go/src/github.com/dgraph-io/dgraph/worker/backup.go:254 +0x85e
github.com/dgraph-io/dgraph/worker.handleBackupForGroup(0x7f6d936a4970, 0xc4200105b0, 0x627d3dd58c69d20, 0xc400000001, 0x0)
	/home/pawan/go/src/github.com/dgraph-io/dgraph/worker/backup.go:330 +0xa83
github.com/dgraph-io/dgraph/worker.BackupOverNetwork.func1(0xc42b4f6d20, 0x7f6d936a4970, 0xc4200105b0, 0x1)
	/home/pawan/go/src/github.com/dgraph-io/dgraph/worker/backup.go:445 +0x4b
created by github.com/dgraph-io/dgraph/worker.BackupOverNetwork
	/home/pawan/go/src/github.com/dgraph-io/dgraph/worker/backup.go:446 +0x138
```

pawanrawal:
Fixes #37

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/42)

### DIFF
--- a/badger/iterator.go
+++ b/badger/iterator.go
@@ -138,9 +138,6 @@ func (it *Iterator) Next() {
 
 	// Set next item to current
 	it.item = it.data.pop()
-	if it.item != nil && bytes.Equal(it.item.Key(), head) {
-		it.item = it.data.pop()
-	}
 
 	// Advance internal iterator until entry is not deleted
 	for it.iitr.Valid() {
@@ -178,6 +175,9 @@ func (it *Iterator) prefetch() {
 	var count int
 	it.item = nil
 	for ; i.Valid(); i.Next() {
+		if bytes.Equal(it.iitr.Key(), head) {
+			continue
+		}
 		if i.Value().Meta&BitDelete > 0 {
 			continue
 		}
--- a/badger/iterator.go
+++ b/badger/iterator.go
@@ -97,10 +97,11 @@ type Iterator struct {
 	kv   *KV
 	iitr y.Iterator
 
-	opt   IteratorOptions
-	item  *KVItem
-	data  list
-	waste list
+	opt      IteratorOptions
+	item     *KVItem
+	data     list
+	waste    list
+	pastHead bool
 }
 
 func (it *Iterator) newItem() *KVItem {
@@ -137,14 +138,15 @@ func (it *Iterator) Next() {
 
 	// Set next item to current
 	it.item = it.data.pop()
-	if bytes.Equal(it.item.Key(), head) {
+	if it.item != nil && bytes.Equal(it.item.Key(), head) {
 		it.item = it.data.pop()
 	}
 
 	// Advance internal iterator until entry is not deleted
 	for it.iitr.Valid() {
 		it.iitr.Next()
-		if bytes.Equal(it.iitr.Key(), head) {
+		if !it.pastHead && bytes.Equal(it.iitr.Key(), head) {
+			it.pastHead = true
 			continue
 		}
 		if it.iitr.Value().Meta&BitDelete == 0 {
@@ -205,6 +207,7 @@ func (it *Iterator) Seek(key []byte) {
 	it.iitr.Seek(key)
 	if bytes.Equal(it.iitr.Key(), head) {
 		it.iitr.Next()
+		it.pastHead = true
 	}
 	it.prefetch()
 }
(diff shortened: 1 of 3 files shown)


manishrjain:
Let me know when this is ready for review.

manishrjain:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[badger/iterator.go, line 208 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/42#-Kki5Jdx5Z2NJmPwNUOh:-Kki5Jdx5Z2NJmPwNUOi:b-nml42q) ([raw file](https://github.com/dgraph-io/badger/blob/0e11f61b4b6945ce977efaed26a795798366ce89/badger/iterator.go#L208)):*
> ```Go
> 	}
> 	it.iitr.Seek(key)
> 	if !it.pastHead && bytes.Equal(it.iitr.Key(), head) {
> ```

We should probably do this for an entire prefix, that is specific to badger. So, we should change the name of this to `!badger!head`; and then skip all keys who start with `!badger!`. `!` occurs earlier than `/`, and is also much less commonly used.

I think to avoid complexity, and to avoid doing micro optimizations (which I suggested), let's just use `bytes.HasPrefix(..)` to determine if they key has this `!badger!` prefix, and if so, skip the keys until you find the first non-such key. Then we can just get rid of this `pastHead` logic and just compare every key. `bytes.HasPrefix` should be able to determine the outcome pretty fast anyway.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/42)*

pawanrawal:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[badger/iterator.go, line 208 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/42#-Kki5Jdx5Z2NJmPwNUOh:-KkiKYRWOwhDFbSzYkCA:b-oh178x) ([raw file](https://github.com/dgraph-io/badger/blob/0e11f61b4b6945ce977efaed26a795798366ce89/badger/iterator.go#L208)):*

We should probably do this for an entire prefix, that is specific to badger. So, we should change the name of this to `!badger!head`; and then skip all keys who start with `!badger!`. `!` occurs earlier than `/`, and is also much less commonly used.

I think to avoid complexity, and to avoid doing micro optimizations (which I suggested), let's just use `bytes.HasPrefix(..)` to determine if they key has this `!badger!` prefix, and if so, skip the keys until you find the first non-such key. Then we can just get rid of this `pastHead` logic and just compare every key. `bytes.HasPrefix` should be able to determine the outcome pretty fast anyway.
</blockquote></details>

Done. That's cleaner and leaves scope for adding new keys later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/42)*

manishrjain:
:lgtm:

---

Reviewed 3 of 3 files at r3.

---

*[badger/iterator.go, line 176 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/42#-KlHPOZEURji7FYa4Oyt:-KlHPOZEURji7FYa4Oyu:b-dwogs5) ([raw file](https://github.com/dgraph-io/badger/blob/7a9e9ff97f2a0528ed96c1526bce73df88c3c0af/badger/iterator.go#L176)):*
> ```Go
> 	it.item = nil
> 	for ; i.Valid(); i.Next() {
> 		if bytes.HasPrefix(it.iitr.Key(), badgerPrefix) {
> ```

Use i instead of it.iitr

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/42#-:-KlHSb1t1djAT3a95dL0:bnfp4nl)*

PR #69, opened 2017-06-20
Add Exists method on KV struct

### LINKED ISSUE #66: Possibility to test existence of a key without reaching Value log
Currently to test if a key exists or not I use `Get` and then test if the Value is nil or not.
Does that mean we hit the value log every time or we only stay in the memtable and LSM ?

zaibon:
fixes #66

Results of the benchmark 
```
BenchmarkExists/WithGet-4         	  200000	      5739 ns/op
BenchmarkExists/WithExists-4      	  300000	      4634 ns/op
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/69)

### DIFF
--- a/badger/doc_test.go
+++ b/badger/doc_test.go
@@ -27,6 +27,7 @@ var d string = "doc"
 func Example() {
 	opt := badger.DefaultOptions
 	opt.Dir = "/tmp"
+	opt.ValueDir = "/tmp"
 	kv, _ := badger.NewKV(&opt)
 
 	key := []byte("hello")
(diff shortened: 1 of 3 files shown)


manishrjain:
:lgtm: Thanks for the PR.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/69#-:-KnMPh7bRSrm3ehhAOz8:bke5l9a)*

PR #70, opened 2017-06-24
Added ValidForPrefix() method on Iterator

celrenheit:
Added a convenience method on Iterator to check for prefix.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/70)

### DIFF
--- a/badger/iterator.go
+++ b/badger/iterator.go
@@ -126,7 +126,7 @@ func (it *Iterator) Valid() bool { return it.item != nil }
 // ValidForPrefix returns false when iteration is done
 // or when the current key is not prefixed by the specified prefix.
 func (it *Iterator) ValidForPrefix(prefix []byte) bool {
-	return it.Valid() && bytes.HasPrefix(it.Item().Key(), prefix)
+	return it.item != nil && bytes.HasPrefix(it.item.key, prefix)
 }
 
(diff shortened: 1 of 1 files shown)


manishrjain:
Reviewed 1 of 1 files at r1.

---

*[badger/iterator.go, line 129 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/70#-Kncn5SiggsnukM8mvbq:-Kncn5SiggsnukM8mvbr:bp2uiam) ([raw file](https://github.com/dgraph-io/badger/blob/ce9ff7317297ed3f6226c69688c030a7f507a5fc/badger/iterator.go#L129)):*
> ```Go
> // or when the current key is not prefixed by the specified prefix.
> func (it *Iterator) ValidForPrefix(prefix []byte) bool {
> 	return it.Valid() && bytes.HasPrefix(it.Item().Key(), prefix)
> ```

You could avoid the indirection, like so. This would be faster when run repeatedly in a loop.

it.item != nil && bytes.HasPrefix(it.item.Key(), prefix)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/70)*

manishrjain:
Thanks for the PR!

PR #99, opened 2017-07-10
Make (*KV).doWrites limit batch size and avoid polling

srh:
Only one do-channel-stuff-and-sleep loop (this one) is trivially convertible because the others hold locks while accessing the channel.  So you wouldn't want to block while doing that.  (But is it actually necessary to hold a lock while waiting on the channel?  That's a strange thing to see, I might investigate that later.)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/99)


manishrjain:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/99#-:-Koe_FBSoV39ZlDUe9QU:bnfp4nl)*

PR #102, opened 2017-07-16
Fix error cleanup in and after OpenTable calls

srh:
- We now close fd inside of OpenTable upon error.

- At many callsites (where we call OpenTable many times in a loop) we
now cleanup resources properly.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/102)

### DIFF
--- a/compact_log_test.go
+++ b/compact_log_test.go
@@ -183,6 +183,7 @@ func TestOverlappingKeyRangeError(t *testing.T) {
 
 	f = buildTestTable(t, "l", 2)
 	t2, err := table.OpenTable(f, table.MemoryMap)
+	require.NoError(t, err)
 	defer t2.DecrRef()
 	done = lh0.tryAddLevel0Table(t2)
 	require.Equal(t, true, done)
--- a/kv.go
+++ b/kv.go
@@ -953,12 +953,11 @@ func (s *KV) flushMemtable(lc *y.LevelCloser) error {
 		}
 
 		tbl, err := table.OpenTable(fd, s.opt.MapTablesTo)
-		defer tbl.DecrRef()
-
 		if err != nil {
 			s.elog.Printf("ERROR while opening table: %v", err)
 			return err
 		}
+		defer tbl.DecrRef()
 		s.lc.addLevel0Table(tbl) // This will incrRef again.
 
 		// Update s.imm. Need a lock.
--- a/levels.go
+++ b/levels.go
@@ -96,12 +96,14 @@ func newLevelsController(kv *KV) (*levelsController, error) {
 		fname := table.NewFilename(fileID, kv.opt.Dir)
 		fd, err := y.OpenSyncedFile(fname, true)
 		if err != nil {
+			closeAllUnmodifiedTables(tables)
 			return nil, errors.Wrapf(err, "Opening file: %q", fname)
 		}
 
 		t, err := table.OpenTable(fd, kv.opt.MapTablesTo)
 		if err != nil {
-			return nil, errors.Wrapf(err, "Opening table: %v", fd)
+			closeAllUnmodifiedTables(tables)
+			return nil, errors.Wrapf(err, "Opening table: %q", fname)
 		}
 
 		// Check metadata for level information.
@@ -121,18 +123,42 @@ func newLevelsController(kv *KV) (*levelsController, error) {
 	for i, tbls := range tables {
 		s.levels[i].initTables(tbls)
 	}
-	//	s.debugPrintMore()
 
 	// Make sure key ranges do not overlap etc.
 	if err := s.validate(); err != nil {
+		_ = s.cleanupLevels()
 		return nil, errors.Wrap(err, "Level validation")
 	}
 
 	// Create new compact log.
 	if err := s.clog.init(clogName); err != nil {
+		_ = s.cleanupLevels()
 		return nil, errors.Wrap(err, "Compaction Log")
 	}
 	return s, nil
+}
+
+// Closes the tables, for cleanup in newLevelsController.  (We Close() instead of using DecrRef()
+// because that would delete the underlying files.)  We ignore errors when closing the table, which
+// is OK because we haven't modified the table.  (It's an LSM tree, but we do hypothetically modify
+// the file with SetMetadata.)  (Even then, we open the files with O_DSYNC, so the only plausible
+// error on Close() would have been from writing file metadata.)
+func closeAllUnmodifiedTables(tables [][]*table.Table) {
+	for _, tableSlice := range tables {
+		for _, table := range tableSlice {
+			_ = table.Close()
+		}
+	}
+}
+
+func (s *levelsController) cleanupLevels() error {
+	var firstErr error
+	for _, l := range s.levels {
+		if err := l.close(); err != nil && firstErr == nil {
+			firstErr = err
+		}
+	}
+	return firstErr
 }
 
 func (s *levelsController) startCompact(lc *y.LevelCloser) {
@@ -242,10 +268,10 @@ func (s *levelsController) compactBuildTables(
 			}
 			y.Check(builder.Add(it.Key(), it.Value()))
 		}
-		if builder.Empty() {
-			builder.Close()
-			continue
-		}
+		// It was true that it.Valid() at least once in the loop above, which means we
+		// called Add() at least once, and builder is not Empty().
+		y.AssertTrue(!builder.Empty())
+
 		cd.elog.LazyPrintf("LOG Compact. Iteration to generate one table took: %v\n", time.Since(timeStart))
 
 		y.AssertTruef(newID <= newIDMax, "%d %d", newID, newIDMax)
@@ -276,10 +302,23 @@ func (s *levelsController) compactBuildTables(
 	}
 
 	// Wait for all table builders to finish.
+	var firstErr error
 	for x := 0; x < i; x++ {
-		if err := <-che; err != nil {
-			return nil, func() error { return errors.Wrapf(err, "While running compaction for: %+v", c) }
+		if err := <-che; err != nil && firstErr == nil {
+			firstErr = err
 		}
+	}
+
+	if firstErr != nil {
+		// An error happened.  Delete all the newly created table files (by calling DecrRef
+		// -- we're the only holders of a ref).
+		for _, table := range newTables[:i] {
+			if table != nil {
+				_ = table.DecrRef()
+			}
+		}
+		errorReturn := errors.Wrapf(firstErr, "While running compaction for: %+v", c)
+		return nil, func() error { return errorReturn }
 	}
 
 	out := newTables[:i]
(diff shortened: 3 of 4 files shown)


srh:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[levels.go, line 313 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpC6VQ_jcYdDXLrrNiN:-KpC6VQ_jcYdDXLrrNiO:b-gdakt0) ([raw file](https://github.com/dgraph-io/badger/blob/e64a65a8e03ec1cd4a91011e13c10277bef1f21e/levels.go#L313)):*
> ```Go
> 		for _, table := range newTables[:i] {
> 			if table != nil {
> 				_ = table.Close()
> ```

When cleaning up here I chose to close the tables.  Instead we could DecrRef and delete the newly created files, or otherwise directly act to delete the newly created files.  I don't yet have enough of a "big picture" view of Badger to know what is right here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/102)*

srh:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[levels.go, line 269 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpChgIqA8sHGmZzwPkD:-KpChgIqA8sHGmZzwPkE:b-pmdrpd) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L269)):*
> ```Go
> 			y.Check(builder.Add(it.Key(), it.Value()))
> 		}
> 		// it.Valid() at least once in the loop above, hence we called .Add, hence
> ```

This "builder.Empty()" condition can never happen.  That's good news, because if it did happen, we would have needed to push a nil error onto `che` so that  the loop below ("Wait for all table builders to finish") doesn't hang.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/102)*

manishrjain:
This one is a tricky change. Would be hard to test it as well, given it's hard to reproduce FS failures. But, at least do this. We have a populate script in badger-bench. Run it with 250M keys and 128B values. It would consume 40GB. That should ensure that at least this works well for the normal case.

---

Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[levels.go, line 313 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpC6VQ_jcYdDXLrrNiN:-KpHKceCul9vRI1J_8Vx:b8ty7vx) ([raw file](https://github.com/dgraph-io/badger/blob/e64a65a8e03ec1cd4a91011e13c10277bef1f21e/levels.go#L313)):*

When cleaning up here I chose to close the tables.  Instead we could DecrRef and delete the newly created files, or otherwise directly act to delete the newly created files.  I don't yet have enough of a "big picture" view of Badger to know what is right here.
</blockquote></details>

You do need to call DecrRef here because when tables are opened, they have at least one ref. This way, we'll delete them using the logic already in place (truncate fanciness and all).

---

*[levels.go, line 147 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpHHlMa98zWmGb6QCJa:-KpHHlMa98zWmGb6QCJb:bpih97o) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L147)):*
> ```Go
> 	for _, tableSlice := range tables {
> 		for _, table := range tableSlice {
> 			_ = table.Close()
> ```

My first thought was: If you close the table, but someone is holding a reference to it, this would cause issues.

But, I realize that this is only called on tables whose ref is not yet being held by anyone. If that's the case, can you add a comment about it. I think the "unmodified" part is largely irrelevant because all our writes are synced to disk with every `write` system call because of O_DSYNC flag. So, closing shouldn't make a difference (apart from the number of open file descriptors).

---

*[levels.go, line 152 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpHGLg7zVm3AqMUxNQz:-KpHGLg7zVm3AqMUxNR-:b-ag5ynb) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L152)):*
> ```Go
> }
> 
> func cleanupLevels(s *levelsController) error {
> ```

Make it a member of levelsController.

---

*[levels.go, line 155 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpHGUED9r3ExnGZdGWm:-KpHGUED9r3ExnGZdGWn:b13i0te) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L155)):*
> ```Go
> 	var firstErr error
> 	for _, l := range s.levels {
> 		if err := l.close(); err != nil && firstErr == nil {
> ```

Shouldn't matter if it's first or last. Can simplify the if statement though:

`if err := l.close(); err != nil { rerr = err }`

---

*[levels.go, line 269 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpChgIqA8sHGmZzwPkD:-KpHH6HiMRbE89yj5MHl:bmszpo6) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L269)):*

This "builder.Empty()" condition can never happen.  That's good news, because if it did happen, we would have needed to push a nil error onto `che` so that  the loop below ("Wait for all table builders to finish") doesn't hang.
</blockquote></details>

Good point. Can you fix up the grammar of comment statement though?

---

*[levels.go, line 271 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpHHEPK9U3-UnBgZbd1:-KpHHEPK9U3-UnBgZbd2:b-98j2tn) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L271)):*
> ```Go
> 		// it.Valid() at least once in the loop above, hence we called .Add, hence
> 		// !builder.Empty.
> 		y.AssertTruef(!builder.Empty(), "compactBuildTables builder cannot be empty")
> ```

No need for truef. They're costly. Just use y.AssertTrue.

---

*[levels.go, line 305 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpHK3k6_H3ADRi4KbCo:-KpHK3k6_H3ADRi4KbCp:br0a99p) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L305)):*
> ```Go
> 	var firstErr error
> 	for x := 0; x < i; x++ {
> 		if err := <-che; err != nil && firstErr == nil {
> ```

Again, no need for first error (can be the last one). Can simplify if statement.

---

*[levels.go, line 574 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpHJamkGUl3-aMNNlF_:-KpHJaml_CWaixj6RIZb:bxnlu3u) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L574)):*
> ```Go
> 		return errors.Wrap(cleanupErr, "levelsController.Close")
> 	}
> 	return err
> ```

errors.Wrap here as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/102#-:-KpHMaF9h7gM8XlQmp2v:b5mq9tg)*

srh:
Currently running the populate script.  54.24M keys so far.

---

Review status: 3 of 4 files reviewed at latest revision, 8 unresolved discussions.

---

*[levels.go, line 313 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpC6VQ_jcYdDXLrrNiN:-KpHjQ8V-5kdPunm1Pza:b-rur6um) ([raw file](https://github.com/dgraph-io/badger/blob/e64a65a8e03ec1cd4a91011e13c10277bef1f21e/levels.go#L313)):*

You do need to call DecrRef here because when tables are opened, they have at least one ref. This way, we'll delete them using the logic already in place (truncate fanciness and all).
</blockquote></details>

Okay, so the decision here is that we do want to delete those files.

---

*[levels.go, line 147 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpHHlMa98zWmGb6QCJa:-KpHlyKuenm_QWmKb1ux:bhube7z) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L147)):*

My first thought was: If you close the table, but someone is holding a reference to it, this would cause issues.

But, I realize that this is only called on tables whose ref is not yet being held by anyone. If that's the case, can you add a comment about it. I think the "unmodified" part is largely irrelevant because all our writes are synced to disk with every `write` system call because of O_DSYNC flag. So, closing shouldn't make a difference (apart from the number of open file descriptors).
</blockquote></details>

Right.  The only plausible error is when writing last modified time, or in the case of some really janky file system.  We should still say unmodified here -- I want to warn people off from blindly using it -- because we might not use O_DSYNC in the future (we might explicitly fdatasync, or under some user option, not sync at all).

---

*[levels.go, line 152 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpHGLg7zVm3AqMUxNQz:-KpHs48qM-W0lGMWlVGk:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L152)):*

Make it a member of levelsController.
</blockquote></details>

Done.

---

*[levels.go, line 155 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpHGUED9r3ExnGZdGWm:-KpHnhAtJkYj7IqPLqbI:b-25fbb1) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L155)):*

Shouldn't matter if it's first or last. Can simplify the if statement though:

`if err := l.close(); err != nil { rerr = err }`
</blockquote></details>

I'm under the impression that the first is more likely to have useful info, while later errors might be of a more generic nature, or be the result of invalid state resulting from under-tested error handling after the first error.  As a general rule.  That's why I've been returning first errors.

---

*[levels.go, line 269 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpChgIqA8sHGmZzwPkD:-KpHs2VMRt09pmpdujHV:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L269)):*

Good point. Can you fix up the grammar of comment statement though?
</blockquote></details>

Done.

---

*[levels.go, line 271 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpHHEPK9U3-UnBgZbd1:-KpHrvJupPXMLdjJdr0Y:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L271)):*

No need for truef. They're costly. Just use y.AssertTrue.
</blockquote></details>

Done.

---

*[levels.go, line 574 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/102#-KpHJamkGUl3-aMNNlF_:-KpHs0Jd1A1bdvt_eess:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/d713a31eca23c470375776a8eca775b77957f138/levels.go#L574)):*

errors.Wrap here as well.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/102#-:-KpHrq7mNr881I50GAMr:b-apli3t)*

manishrjain:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/102#-:-KpIUC1J-I1fEhZ2ItG5:bnfp4nl)*

PR #110, opened 2017-07-20
Check Pid of process written to LOCK file

pawanrawal:
This would ensure that if `dgraph` crashes or if we interrupt the `dgraphloader` then we can restart without having to delete the LOCK file.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/110)

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -17,12 +17,17 @@
 package badger
 
 import (
+	"bytes"
 	"fmt"
+	"io"
 	"log"
 	"math"
 	"os"
 	"path/filepath"
+	"strconv"
+	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"golang.org/x/net/trace"
@@ -146,7 +151,7 @@ func NewKV(opt *Options) (out *KV, err error) {
 			return nil, ErrInvalidDir
 		}
 	}
-	if err := createLockFile(filepath.Join(opt.Dir, lockFile)); err != nil {
+	if err := updateLockFile(filepath.Join(opt.Dir, lockFile)); err != nil {
 		return nil, err
 	}
 	if !(opt.ValueLogFileSize <= 2<<30 && opt.ValueLogFileSize >= 1<<20) {
@@ -333,13 +338,24 @@ const (
 	lockFile = "LOCK"
 )
 
-// Opens a file, errors if it exists, and writes the process id to the file
-func createLockFile(path string) error {
-	f, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0666)
+func processExists(p string) bool {
+	pid, err := strconv.Atoi(p)
 	if err != nil {
-		return errors.Wrap(err, "cannot create pid lock file")
+		log.Fatal(err)
 	}
-	_, err = fmt.Fprintf(f, "%d\n", os.Getpid())
+	process, err := os.FindProcess(pid)
+	if err != nil {
+		return true
+	}
+	// On Unix systems FindProcess always succeeds even if the process doesn't exist.
+	// Sending 0 signal doesn't send a signal but can be used to check for the existence
+	// of a process.
+	err = process.Signal(syscall.Signal(0))
+	return err == nil
+}
+
+func writeAndClose(f *os.File) error {
+	_, err := fmt.Fprintf(f, "%d\n", os.Getpid())
 	closeErr := f.Close()
 	if err != nil {
 		return errors.Wrap(err, "cannot write to pid lock file")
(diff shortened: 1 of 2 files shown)


manishrjain:
I have some comments. I'll leave it to @srh for final approval. Also, understand, why are we doing this? Is this important? What if rocksDB crashes, does it remove the lock file?

---

Reviewed 2 of 2 files at r1.

---

*[kv.go, line 341 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/110#-KpTnxLIMwHSB73mPYyl:-KpTnxLJMeWYbhvVOLe2:bj34vpb) ([raw file](https://github.com/dgraph-io/badger/blob/ea432eadef0b719613279658e21dd72c8dbe77a6/kv.go#L341)):*
> ```Go
> )
> 
> func processExists(p string) (bool, error) {
> ```

Why error? Just return a bool. Pass or fail.

---

*[kv.go, line 377 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/110#-KpToFfBGbnqmU8zDVG0:-KpToFfBGbnqmU8zDVG1:bpnhyp8) ([raw file](https://github.com/dgraph-io/badger/blob/ea432eadef0b719613279658e21dd72c8dbe77a6/kv.go#L377)):*
> ```Go
> 	// It already exists, lets check if the process for the PID exists. If it does then error out,
> 	// else overwrite the PID.
> 	f, err := os.OpenFile(path, os.O_RDWR, 0666)
> ```

ioutil.ReadFile?

---

*[kv.go, line 403 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/110#-KpToU_OoNshDtD_8KVi:-KpToU_OoNshDtD_8KVj:bnnmxwi) ([raw file](https://github.com/dgraph-io/badger/blob/ea432eadef0b719613279658e21dd72c8dbe77a6/kv.go#L403)):*
> ```Go
> // Opens a file, errors if it exists, and writes the process id to the file
> func createLockFile(path string) error {
> 	f, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE, 0666)
> ```

These permissions aren't the same as the one initially implemented.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/110#-:-KpTolAGczRKNKuNfCol:b-skoinh)*

pawanrawal:
RocksDB holds a lock on the file. It doesn't write anything to the file. We can do that too, it would simplify the code. Though I am not sure if Go standard library supports that directly.

This is important so that users don't have to manually remove the file in case of a crash or say when you interrupt the `dgraphloader`.  

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/110#-:-KpTqToF5AKb--Gf8B_d:beyz4zw)*

manishrjain:
And what happens in case of a crash? Does the file stay or go?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/110#-:-KpTt_XiCc9YEdS1wghG:b-cceoqn)*

pawanrawal:
It stays (I tried with a `kill -9`), just that no process is holding lock on it so it can be locked again by some other process.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/110#-:-KpTuO-Pm2AmcfLHFp3B:b-of3u60)*

manishrjain:
Try to simulate what they're doing, and keep the code changes small.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/110#-:-KpTwFcnC0a9IORN9PlT:b-onlmy0)*

srh:
I think we should just flock the directory.  Which does work.  You'd have to dip into unix.Flock.

---

---

*[kv.go, line 371 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/110#-KpTw_nnKDRzK5Muves5:-KpTw_nnKDRzK5Muves6:bh086ku) ([raw file](https://github.com/dgraph-io/badger/blob/ea432eadef0b719613279658e21dd72c8dbe77a6/kv.go#L371)):*
> ```Go
> func updateLockFile(path string) error {
> 	// If LOCK file doesn't exist, then we create it and write PID to it.
> 	if _, err := os.Stat(path); os.IsNotExist(err) {
> ```

If two processes start at the same time, both of them might hit this case where the file does not exist, and then try to create the lock file without O_EXCL, and then write to the file.

Also, if the file exists, we also have that problem -- two processes can read the file, determine no process with that pid exists, and then write their pid to the file.  I think the right solution would be to use flock, but see my comment about flocking the directory.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/110#-:-KpTwUOx0rt6o8sl9dvU:brejao)*

pawanrawal:
#111 will take care of this.

PR #130, opened 2017-07-26
Make cas counter a 64-bit incrementing counter

srh:
Makes the cas counter a 64-bit incrementing counter.  It's actually an accurate and precise count of how many values we put into the value log.

If I didn't have @janardhan1993's recent userMeta change to look at it'd be a lot more difficult and error-prone, so I added some constants, deduped and moved ValueStruct encoding/decoding logic to one place to make similar modifications easier in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/130)

### DIFF
--- a/compact_log_test.go
+++ b/compact_log_test.go
@@ -138,7 +138,7 @@ func buildTable(t *testing.T, keyValues [][]string) *os.File {
 	})
 	for i, kv := range keyValues {
 		y.AssertTrue(len(kv) == 2)
-		err := b.Add([]byte(kv[0]), y.ValueStruct{[]byte(kv[1]), 'A', uint16(i), 0})
+		err := b.Add([]byte(kv[0]), y.ValueStruct{[]byte(kv[1]), 'A', 0, uint16(i)})
 		if t != nil {
 			require.NoError(t, err)
 		} else {
--- a/compact_log_test.go
+++ b/compact_log_test.go
@@ -138,7 +138,7 @@ func buildTable(t *testing.T, keyValues [][]string) *os.File {
 	})
 	for i, kv := range keyValues {
 		y.AssertTrue(len(kv) == 2)
-		err := b.Add([]byte(kv[0]), y.ValueStruct{[]byte(kv[1]), 'A', 0, uint16(i)})
+		err := b.Add([]byte(kv[0]), y.ValueStruct{[]byte(kv[1]), 'A', 0, uint64(i)})
 		if t != nil {
 			require.NoError(t, err)
 		} else {
--- a/iterator.go
+++ b/iterator.go
@@ -32,7 +32,7 @@ type KVItem struct {
 	meta       byte
 	userMeta   byte
 	val        []byte
-	casCounter uint16
+	casCounter uint64
 	slice      *y.Slice
 	next       *KVItem
 }
@@ -51,7 +51,7 @@ func (item *KVItem) Value() []byte {
 }
 
 // Counter returns the CAS counter associated with the value.
-func (item *KVItem) Counter() uint16 {
+func (item *KVItem) Counter() uint64 {
 	return item.casCounter
 }
 
--- a/kv.go
+++ b/kv.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"golang.org/x/net/trace"
@@ -120,8 +121,7 @@ func (opt *Options) estimateSize(entry *Entry) int {
 // KV provides the various functions required to interact with Badger.
 // KV is thread-safe.
 type KV struct {
-	// Guards list of inmemory tables, and vptr, casAsOfVptr, not individual reads and writes.
-	sync.RWMutex
+	sync.RWMutex // Guards list of inmemory tables, not individual reads and writes.
 
 	dirLockGuard *y.DirectoryLockGuard
 	// nil if Dir and ValueDir are the same
@@ -135,11 +135,12 @@ type KV struct {
 	lc        *levelsController
 	vlog      valueLog
 	vptr      valuePointer
-	casAtvptr uint64 // The last-written cas counter as of the write at vptr.
 	writeCh   chan *request
 	flushChan chan flushTask // For flushing memtables.
 
-	// Incremented serially in the write loop.
+	// Incremented in the non-concurrently accessed write loop.  But also incremented outside.
+	// So we use an atomic op.
+	// TODO: Don't use an atomic op.
 	lastUsedCasCounter uint64
 }
 
@@ -519,20 +520,16 @@ func (s *KV) Exists(key []byte) (bool, error) {
 	return true, nil
 }
 
-func (s *KV) updateOffset(ptrs []valuePointer, casAtvptr uint64) {
+func (s *KV) updateOffset(ptrs []valuePointer) {
 	ptr := ptrs[len(ptrs)-1]
 	s.Lock()
 	defer s.Unlock()
-
 	if s.vptr.Fid < ptr.Fid {
 		s.vptr = ptr
-		s.casAtvptr = casAtvptr
 	} else if s.vptr.Offset < ptr.Offset {
 		s.vptr = ptr
-		s.casAtvptr = casAtvptr
 	} else if s.vptr.Fid == ptr.Fid && s.vptr.Offset == ptr.Offset && s.vptr.Len < ptr.Len {
 		s.vptr = ptr
-		s.casAtvptr = casAtvptr
 	}
 }
 
@@ -598,6 +595,19 @@ func (s *KV) writeToLSM(b *request) error {
 	return nil
 }
 
+// newCASCounter is called serially by only one goroutine.
+// TODO: Actually, we also call it when setting the !badger!head entry.
+func (s *KV) newCASCounter() uint64 {
+	return atomic.AddUint64(&s.lastUsedCasCounter, 1)
+}
+
+// newCASCounters generates a set of unique CAS counters -- the interval [x, x + howMany) where x
+// is the return value.
+func (s *KV) newCASCounters(howMany uint64) uint64 {
+	last := atomic.AddUint64(&s.lastUsedCasCounter, howMany)
+	return last - howMany + 1
+}
+
 // writeRequests is called serially by only one goroutine.
 func (s *KV) writeRequests(reqs []*request) error {
 	if len(reqs) == 0 {
@@ -616,11 +626,10 @@ func (s *KV) writeRequests(reqs []*request) error {
 	// CAS counter for all operations has to go onto value log. Otherwise, if it is just in memtable for
 	// a long time, and following CAS operations use that as a check, when replaying, we will think that
 	// these CAS operations should fail, when they are actually valid.
-
 	for _, req := range reqs {
-		for _, e := range req.Entries {
-			s.lastUsedCasCounter++
-			e.casCounter = s.lastUsedCasCounter
+		counterBase := s.newCASCounters(uint64(len(req.Entries)))
+		for i, e := range req.Entries {
+			e.casCounter = counterBase + uint64(i)
 		}
 	}
 	err := s.vlog.write(reqs)
@@ -651,7 +660,7 @@ func (s *KV) writeRequests(reqs []*request) error {
 			done(err)
 			return errors.Wrap(err, "writeRequests")
 		}
-		s.updateOffset(b.Ptrs, b.Entries[len(b.Entries)-1].casCounter)
+		s.updateOffset(b.Ptrs)
 	}
 	done(nil)
 	s.elog.Printf("%d entries written", count)
@@ -999,12 +1008,17 @@ func (s *KV) flushMemtable(lc *y.LevelCloser) error {
 			offset := make([]byte, 10)
 			s.Lock() // For vptr and casAsOfVptr.
 			s.vptr.Encode(offset)
-			casAsOfVptr := s.casAtvptr
 			s.Unlock()
-			// We write the CAS counter in !badger!head because we use it at startup
-			// when replaying the value log and reconstructing the last used CAS
-			// counter.
-			ft.mt.Put(head, y.ValueStruct{Value: offset, CASCounter: casAsOfVptr})
+			// CAS counter is needed and is desirable -- it's the first value log entry
+			// we replay, so to speak, perhaps the only, and we use it to re-initialize
+			// the CAS counter.
+			//
+			// In the write loop, the s.vptr value gets updated _after_ the values are
+			// written to the value log.  This means we might store a !badger!head with
+			// a later CAS counter (because it's accessed atomically and not tied to
+			// the vptr value by any mechanism, and then replay value log entries with
+			// a smaller CAS counter.
+			ft.mt.Put(head, y.ValueStruct{Value: offset, CASCounter: s.newCASCounter()})
 		}
 		fileID, _ := s.lc.reserveFileIDs(1)
 		fd, err := y.OpenSyncedFile(table.NewFilename(fileID, s.opt.Dir), true)
--- a/kv.go
+++ b/kv.go
@@ -982,7 +982,10 @@ func (s *KV) flushMemtable(lc *y.LevelCloser) error {
 			s.Lock() // For vptr.
 			s.vptr.Encode(offset)
 			s.Unlock()
-			ft.mt.Put(head, y.ValueStruct{Value: offset}) // casCounter not needed.
+			// CAS counter is needed and is desirable -- it's the first value log entry
+			// we reply, perhaps the only, and we use it to re-initialize the CAS
+			// counter.
+			ft.mt.Put(head, y.ValueStruct{Value: offset, CASCounter: newCASCounter()})
 		}
 		fileID, _ := s.lc.reserveFileIDs(1)
 		fd, err := y.OpenSyncedFile(table.NewFilename(fileID, s.opt.Dir), true)
--- a/kv.go
+++ b/kv.go
@@ -595,10 +595,9 @@ func (s *KV) writeToLSM(b *request) error {
 	return nil
 }
 
-// newCASCounter is called serially by only one goroutine.
-// TODO: Actually, we also call it when setting the !badger!head entry.
-func (s *KV) newCASCounter() uint64 {
-	return atomic.AddUint64(&s.lastUsedCasCounter, 1)
+// lastCASCounter returns the last-used cas counter.
+func (s *KV) lastCASCounter() uint64 {
+	return atomic.LoadUint64(&s.lastUsedCasCounter)
 }
 
 // newCASCounters generates a set of unique CAS counters -- the interval [x, x + howMany) where x
@@ -623,9 +622,13 @@ func (s *KV) writeRequests(reqs []*request) error {
 
 	s.elog.Printf("writeRequests called. Writing to value log")
 
-	// CAS counter for all operations has to go onto value log. Otherwise, if it is just in memtable for
-	// a long time, and following CAS operations use that as a check, when replaying, we will think that
-	// these CAS operations should fail, when they are actually valid.
+	// CAS counter for all operations has to go onto value log. Otherwise, if it is just in
+	// memtable for a long time, and following CAS operations use that as a check, when
+	// replaying, we will think that these CAS operations should fail, when they are actually
+	// valid.
+
+	// There is code (in flushMemtable) whose correctness depends on us generating CAS Counter
+	// values _before_ we modify s.vptr here.
 	for _, req := range reqs {
 		counterBase := s.newCASCounters(uint64(len(req.Entries)))
 		for i, e := range req.Entries {
@@ -1013,12 +1016,11 @@ func (s *KV) flushMemtable(lc *y.LevelCloser) error {
 			// we replay, so to speak, perhaps the only, and we use it to re-initialize
 			// the CAS counter.
 			//
-			// In the write loop, the s.vptr value gets updated _after_ the values are
-			// written to the value log.  This means we might store a !badger!head with
-			// a later CAS counter (because it's accessed atomically and not tied to
-			// the vptr value by any mechanism, and then replay value log entries with
-			// a smaller CAS counter.
-			ft.mt.Put(head, y.ValueStruct{Value: offset, CASCounter: s.newCASCounter()})
+			// The write loop generates CAS counter values _before_ it sets vptr.  It
+			// is crucial that we read the cas counter here _after_ reading vptr.  That
+			// way, our value here is guaranteed to be >= the CASCounter values written
+			// before vptr (because they don't get replayed).
+			ft.mt.Put(head, y.ValueStruct{Value: offset, CASCounter: s.lastCASCounter()})
 		}
 		fileID, _ := s.lc.reserveFileIDs(1)
 		fd, err := y.OpenSyncedFile(table.NewFilename(fileID, s.opt.Dir), true)
--- a/kv.go
+++ b/kv.go
@@ -596,6 +596,13 @@ func (s *KV) newCASCounter() uint64 {
 	return atomic.AddUint64(&s.lastUsedCasCounter, 1)
 }
 
+// newCASCounters generates a set of unique CAS counters -- the interval [x, x + howMany) where x
+// is the return value.
+func (s *KV) newCASCounters(howMany uint64) uint64 {
+	last := atomic.AddUint64(&s.lastUsedCasCounter, howMany)
+	return last - howMany + 1
+}
+
 // writeRequests is called serially by only one goroutine.
 func (s *KV) writeRequests(reqs []*request) error {
 	if len(reqs) == 0 {
@@ -615,8 +622,9 @@ func (s *KV) writeRequests(reqs []*request) error {
 	// a long time, and following CAS operations use that as a check, when replaying, we will think that
 	// these CAS operations should fail, when they are actually valid.
 	for _, req := range reqs {
-		for _, e := range req.Entries {
-			e.casCounter = s.newCASCounter()
+		counterBase := s.newCASCounters(uint64(len(req.Entries)))
+		for i, e := range req.Entries {
+			e.casCounter = counterBase + uint64(i)
 		}
 	}
 	err := s.vlog.write(reqs)
--- a/kv.go
+++ b/kv.go
@@ -1007,6 +1007,12 @@ func (s *KV) flushMemtable(lc *y.LevelCloser) error {
 			// CAS counter is needed and is desirable -- it's the first value log entry
 			// we replay, so to speak, perhaps the only, and we use it to re-initialize
 			// the CAS counter.
+			//
+			// In the write loop, the s.vptr value gets updated _after_ the values are
+			// written to the value log.  This means we might store a !badger!head with
+			// a later CAS counter (because it's accessed atomically and not tied to
+			// the vptr value by any mechanism, and then replay value log entries with
+			// a smaller CAS counter.
 			ft.mt.Put(head, y.ValueStruct{Value: offset, CASCounter: s.newCASCounter()})
 		}
 		fileID, _ := s.lc.reserveFileIDs(1)
--- a/kv.go
+++ b/kv.go
@@ -997,19 +997,13 @@ func (s *KV) flushMemtable(lc *y.LevelCloser) error {
 		if ft.vptr.Fid > 0 || ft.vptr.Offset > 0 || ft.vptr.Len > 0 {
 			s.elog.Printf("Storing offset: %+v\n", ft.vptr)
 			offset := make([]byte, 10)
-			s.Lock() // For vptr.
+			s.Lock() // For vptr and casAsOfVptr.
 			s.vptr.Encode(offset)
 			casAsOfVptr := s.casAsOfVptr
 			s.Unlock()
-			// CAS counter is needed and is desirable -- it's the first value log entry
-			// we replay, so to speak, perhaps the only, and we use it to re-initialize
-			// the CAS counter.
-			//
-			// In the write loop, the s.vptr value gets updated _after_ the values are
-			// written to the value log.  This means we might store a !badger!head with
-			// a later CAS counter (because it's accessed atomically and not tied to
-			// the vptr value by any mechanism, and then replay value log entries with
-			// a smaller CAS counter.
+			// We write the CAS counter in !badger!head because we use it at startup
+			// when replaying the value log and reconstructing the last used CAS
+			// counter.
 			ft.mt.Put(head, y.ValueStruct{Value: offset, CASCounter: casAsOfVptr})
 		}
 		fileID, _ := s.lc.reserveFileIDs(1)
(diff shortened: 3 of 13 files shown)


srh:
Review status: 0 of 13 files reviewed at latest revision, 5 unresolved discussions.

---

*[kv.go, line 521 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0IqE-dDXjFq6KxJeK:-Kq0IqE-dDXjFq6KxJeL:baubgxc) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L521)):*
> ```Go
> 	s.Lock()
> 	defer s.Unlock()
> 	// TODO: We should always hit a case where we assign, right?
> ```

We should always hit a case where we assign, right?  It must be.

We're also missing a "s.vptr.Fid == ptr.Fid &&" in the first `else if` below.  So that should be added.  Or these conditionals should be part of an assertion.

---

*[util.go, line 139 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0J9cTcJtWafLPa3n8:-Kq0J9cTcJtWafLPa3n9:bp6hlt0) ([raw file](https://github.com/dgraph-io/badger/blob/d10e10c39b052c70a7b5c767a9f3e9e2266484ef/util.go#L139)):*
> ```Go
> }
> 
> // mod65535 mods by 65535 fast.
> ```

As a general FYI, I know neither of you wrote this, but the Go compiler will optimize modulo-by-constant operations automatically.

---

*[value.go, line 111 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0LvOviYPElG1OQNVQ:-Kq0LvOviYPElG1OQNVR:b7c8qd9) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/value.go#L111)):*
> ```Go
> // Therefore, the kv pair is only valid for the duration of fn call.
> func (f *logFile) iterate(offset uint32, fn logEntry) error {
> 	_, err := f.fd.Seek(int64(offset), io.SeekStart)
> ```

(io.SeekStart is defined to be 0, this doesn't change anything)

---

*[value.go, line 583 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0M1kMAAGeoIwrfu4a:-Kq0M1kMAAGeoIwrfu4b:b-qnedpn) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/value.go#L583)):*
> ```Go
> 	// Seek to the end to start writing.
> 	var err error
> 	// TODO: Why is there no l.RLock() here?
> ```

A question I have

---

*[y/iterator.go, line 38 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0Nr72IdQQjFCaOKHN:-Kq0Nr72IdQQjFCaOKHO:b-k8orh9) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/y/iterator.go#L38)):*
> ```Go
> 	Meta       byte
> 	UserMeta   byte
> 	CASCounter uint64
> ```

I reordered these two fields.  The reason is, everything was already acting as if these fields are in this Meta/UserMeta/CASCounter order (struct initializers that used the field names, other logic, etc) -- and it's also the three fields' serialization order.  And decreasing the struct size by 8 bytes doesn't hurt.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

manishrjain:
Reviewed 13 of 13 files at r1.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[kv.go, line 115 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0W-E3UZEHSjQtVGRc:-Kq0W-E3UZEHSjQtVGRd:b-vfv9q2) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L115)):*
> ```Go
> func (opt *Options) estimateSize(entry *Entry) int {
> 	if len(entry.Value) < opt.ValueThreshold {
> 		// 4 is for cas + meta
> ```

Remove comment.

---

*[kv.go, line 139 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0WUbbGEe8xn2F8HRW:-Kq0WUbbGEe8xn2F8HRX:b9js9yl) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L139)):*
> ```Go
> 	vlog        valueLog
> 	vptr        valuePointer
> 	casAsOfVptr uint64 // The last-written cas counter as of the write at vptr.
> ```

casAtvptr // I'd keep vptr small case, because of how the variable is named.

Also, do we need it (still reading the code)?

---

*[kv.go, line 143 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0Wfy7KvU7hIK_L_Vj:-Kq0Wfy7KvU7hIK_L_Vk:b-1ql2xv) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L143)):*
> ```Go
> 	flushChan   chan flushTask // For flushing memtables.
> 
> 	// Incremented/accessed in the non-concurrently accessed write loop.
> ```

// Incremented serially in the write loop

---

*[kv.go, line 521 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0IqE-dDXjFq6KxJeK:-Kq0XA0Ry8Exy0tmPbMV:bfv99fu) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L521)):*

We should always hit a case where we assign, right?  It must be.

We're also missing a "s.vptr.Fid == ptr.Fid &&" in the first `else if` below.  So that should be added.  Or these conditionals should be part of an assertion.
</blockquote></details>

`ptr.Fid` can't be lower than `s.vptr.Fid`. Fid and Offset are monotonically increasing. Can add a cheap y.AssertTrue if you want.

---

*[kv.go, line 524 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0XTWLXnawCWr2jkUR:-Kq0XTWLXnawCWr2jkUS:bz3ei0k) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L524)):*
> ```Go
> 	if s.vptr.Fid < ptr.Fid {
> 		s.vptr = ptr
> 		s.casAsOfVptr = casAsOfVptr
> ```

This information is duplicate. We can find the CAS by reading `s.vptr` in the LSM or the value log. Either would give us that information.

---

*[kv.go, line 597 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0XfVUumGgTwoKbxHk:-Kq0XfVUumGgTwoKbxHl:b-vrgrf9) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L597)):*
> ```Go
> 
> // newCASCounter generates a new unique CAS counter.  Is never zero.
> func (s *KV) newCASCounter() uint64 {
> ```

No need for a function call, if this is called from exactly one location. Previously the logic was intricate due to randomization, now it's simple enough to be called inline.

---

*[kv.go, line 1007 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0Y1dsCZvj_O1hEVTX:-Kq0Y1dsCZvj_O1hEVTY:baeip99) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L1007)):*
> ```Go
> 			// when replaying the value log and reconstructing the last used CAS
> 			// counter.
> 			ft.mt.Put(head, y.ValueStruct{Value: offset, CASCounter: casAsOfVptr})
> ```

Yeah, all this logic might not be required, if we just read the vptr and retrieve the CAS on a restart.

---

*[util.go, line 139 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0J9cTcJtWafLPa3n8:-Kq0YDbUS-dGkh5ai1pb:birh66d) ([raw file](https://github.com/dgraph-io/badger/blob/d10e10c39b052c70a7b5c767a9f3e9e2266484ef/util.go#L139)):*

As a general FYI, I know neither of you wrote this, but the Go compiler will optimize modulo-by-constant operations automatically.
</blockquote></details>

OK. I see.

---

*[value.go, line 111 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0LvOviYPElG1OQNVQ:-Kq0YLTiulo6mX78Ba8L:b2nhktp) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/value.go#L111)):*

(io.SeekStart is defined to be 0, this doesn't change anything)
</blockquote></details>

OK.

---

*[value.go, line 583 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0M1kMAAGeoIwrfu4a:-Kq0YlziFB02ZsHYOmot:bgvmt9r) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/value.go#L583)):*

A question I have
</blockquote></details>

I guess because it's only called during initialization, which is single threaded. Maybe run `go test --race`, to see if this is an issue.

---

*[y/iterator.go, line 28 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0_rXBCrZnEqi6Flte:-Kq0_rXBCrZnEqi6Fltf:b-kwdttu) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/y/iterator.go#L28)):*
> ```Go
> 
> const (
> 	ValueMetaOffset     = 0
> ```

Do these consts need to be exposed?

---

*[y/iterator.go, line 38 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0Nr72IdQQjFCaOKHN:-Kq0aGvYkaDTbAIA38z-:b2nhktp) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/y/iterator.go#L38)):*

I reordered these two fields.  The reason is, everything was already acting as if these fields are in this Meta/UserMeta/CASCounter order (struct initializers that used the field names, other logic, etc) -- and it's also the three fields' serialization order.  And decreasing the struct size by 8 bytes doesn't hurt.
</blockquote></details>

OK.

---

*[y/iterator.go, line 52 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0a2Xs3cAtAZM6IF2t:-Kq0a2Xs3cAtAZM6IF2u:b-xbe6bx) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/y/iterator.go#L52)):*
> ```Go
> // DecodeEntireSlice uses the length of the slice to infer the length of the Value field.
> func (v *ValueStruct) DecodeEntireSlice(b []byte) {
> 	v.Value = b[ValueValueOffset:]
> ```

If all the logic for encoding and decoding is right here, and pretty simple, then I question the need for having all these consts around. They only making the code harder to read.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

janardhan1993:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[kv.go, line 247 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0h8QnqdtYi4gHt9OU:-Kq0h8QnqdtYi4gHt9OV:bo86lb6) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L247)):*
> ```Go
> 		}
> 		first = false
> 		out.lastUsedCasCounter = e.casCounter
> ```

while replaying cascounter might not always be in increasing order due to value log gc.
Might be better to take max of current value and what's there in value log.

---

*[kv.go, line 1007 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0Y1dsCZvj_O1hEVTX:-Kq0ikw4Ny7ezZiDB4Rx:b-dwplcd) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L1007)):*

Yeah, all this logic might not be required, if we just read the vptr and retrieve the CAS on a restart.
</blockquote></details>

Again, we might overwrite head with lesser casvalue.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

srh:
Review status: 11 of 13 files reviewed at latest revision, 11 unresolved discussions.

---

*[kv.go, line 115 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0W-E3UZEHSjQtVGRc:-Kq0mJy_eoKPGZmcR7RE:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L115)):*

Remove comment.
</blockquote></details>

Done.

---

*[kv.go, line 139 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0WUbbGEe8xn2F8HRW:-Kq0mKI3kSKD8RLOEhRP:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L139)):*

casAtvptr // I'd keep vptr small case, because of how the variable is named.

Also, do we need it (still reading the code)?
</blockquote></details>

Done.

---

*[kv.go, line 143 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0Wfy7KvU7hIK_L_Vj:-Kq0mKv4ttwwYyTcLA9Q:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L143)):*

// Incremented serially in the write loop
</blockquote></details>

Done.

---

*[kv.go, line 247 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0h8QnqdtYi4gHt9OU:-Kq0mLqGbNZHdoGDhvV8:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L247)):*
<details><summary><i>Previously, janardhan1993 (Janardhan Reddy) wrote…</i></summary><blockquote>

while replaying cascounter might not always be in increasing order due to value log gc.
Might be better to take max of current value and what's there in value log.
</blockquote></details>

Done.

---

*[kv.go, line 521 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0IqE-dDXjFq6KxJeK:-Kq0mMeeY-3dwlq9Xj5N:blrqj5z) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L521)):*

`ptr.Fid` can't be lower than `s.vptr.Fid`. Fid and Offset are monotonically increasing. Can add a cheap y.AssertTrue if you want.
</blockquote></details>

I added an assertion and it failed.  The reason I think, is in (*valueLog) write() we don't initialize valuePointer if the value is below the threshold and SyncWrites is off.  I'll investigate that further outside the context of this PR.  There's probably a cleaner way to pass the value log offset to !badger!head anyway.

---

*[kv.go, line 524 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0XTWLXnawCWr2jkUR:-Kq0mNR7BF4vPyNabOYX:b-ql5br6) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L524)):*

This information is duplicate. We can find the CAS by reading `s.vptr` in the LSM or the value log. Either would give us that information.
</blockquote></details>

(Did alternate solution.)

---

*[kv.go, line 597 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0XfVUumGgTwoKbxHk:-Kq0mTTmwXI8WUpDdsK4:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L597)):*

No need for a function call, if this is called from exactly one location. Previously the logic was intricate due to randomization, now it's simple enough to be called inline.
</blockquote></details>

Done.

---

*[kv.go, line 1007 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0Y1dsCZvj_O1hEVTX:-Kq0mWwMIouqBysAmSGS:b-mof8xz) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L1007)):*
<details><summary><i>Previously, janardhan1993 (Janardhan Reddy) wrote…</i></summary><blockquote>

Again, we might overwrite head with lesser casvalue.
</blockquote></details>

Did the right thing.

---

*[y/iterator.go, line 28 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0_rXBCrZnEqi6Flte:-Kq0vUZHphkKIgYxh2yr:b32w12) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/y/iterator.go#L28)):*

Do these consts need to be exposed?
</blockquote></details>

No

---

*[y/iterator.go, line 52 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0a2Xs3cAtAZM6IF2t:-Kq0vSowZoGKVy10qJJJ:b-m2at79) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/y/iterator.go#L52)):*

If all the logic for encoding and decoding is right here, and pretty simple, then I question the need for having all these consts around. They only making the code harder to read.
</blockquote></details>

I can move them closer to usage.  They have a real reason to exist -- you can find all users of MetaSize, UserMetaSize, CasSize when you need to change those values, look at what their users do, and consider the consequences.  So we shouldn't hardcode the numbers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

manishrjain:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[kv.go, line 1007 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq0Y1dsCZvj_O1hEVTX:-Kq11GSKoqhNLm6ZxmnR:b4fycw5) ([raw file](https://github.com/dgraph-io/badger/blob/e8938e4ff2ee52a7cb0bc2df40b052ae0d6e19ca/kv.go#L1007)):*

Did the right thing.
</blockquote></details>

Use the lastUsedCasCounter here.

---

*[kv.go, line 138 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12DUEwnMMRkg-Y2v1:-Kq12DUEwnMMRkg-Y2v2:bg8cfkz) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L138)):*
> ```Go
> 	vlog      valueLog
> 	vptr      valuePointer
> 	casAtvptr uint64 // The last-written cas counter as of the write at vptr.
> ```

casAtvptr might not be required. When we flushmemtable, at that time you can just pick the latest value of lastUsedCasCounter, which would be the most accurate number.

Edge case: In some crash + data loss, we might not use some of these cas counter values, but that should be alright.

---

*[kv.go, line 143 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12eAirUSGYOCkfdx0:-Kq12eAirUSGYOCkfdx1:bea3ajb) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L143)):*
> ```Go
> 
> 	// Incremented serially in the write loop.
> 	lastUsedCasCounter uint64
> ```

This can be renamed to globalCasCounter, or just casCounter.

---

*[kv.go, line 522 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12YpBiR0Oy5RgFkZb:-Kq12YpBiR0Oy5RgFkZc:b-jmzj7b) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L522)):*
> ```Go
> }
> 
> func (s *KV) updateOffset(ptrs []valuePointer, casAtvptr uint64) {
> ```

casAtvptr doesn't need to be sent here. See comment above about using lastUsedCasCounter.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

srh:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[kv.go, line 138 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12DUEwnMMRkg-Y2v1:-Kq14dHVaAAv79DIHQHw:b-ep8nwu) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L138)):*

casAtvptr might not be required. When we flushmemtable, at that time you can just pick the latest value of lastUsedCasCounter, which would be the most accurate number.

Edge case: In some crash + data loss, we might not use some of these cas counter values, but that should be alright.
</blockquote></details>

That was my first implementation and it's correct behavior but it requires atomic variable access.

---

*[kv.go, line 143 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12eAirUSGYOCkfdx0:-Kq14iwo-yowZ0q-Gl0c:bc5ndlj) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L143)):*

This can be renamed to globalCasCounter, or just casCounter.
</blockquote></details>

The name has to answer the question, is it the last used or the next-to-use.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

manishrjain:

---

*[kv.go, line 138 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12DUEwnMMRkg-Y2v1:-Kq15IneZLbjoRDUz3W0:b-6r7r0t) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L138)):*

That was my first implementation and it's correct behavior but it requires atomic variable access.
</blockquote></details>

Atomic should be alright. Maintaining one variable is better than maintaining two. Plus, less code changes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

srh:

---

*[kv.go, line 138 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12DUEwnMMRkg-Y2v1:-Kq15wCp_gbAVc8XeXHW:bm5rhv3) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L138)):*

Atomic should be alright. Maintaining one variable is better than maintaining two. Plus, less code changes.
</blockquote></details>

It's not better.  Right now you're writing the actual greatest cas value written to the value log before that point.  Now you're writing some other value which is greater than that.  Anybody reading the code has to stop and think of the consequences and make sure they didn't screw up.

For example I didn't think realize until right this moment that the global cas counter must be accessed after we read the vptr behind RLock().  It's much better to have explicit data flow.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

manishrjain:

---

*[kv.go, line 138 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12DUEwnMMRkg-Y2v1:-Kq1CdQfOhS8-2Ms4_yC:b-1d45to) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L138)):*

It's not better.  Right now you're writing the actual greatest cas value written to the value log before that point.  Now you're writing some other value which is greater than that.  Anybody reading the code has to stop and think of the consequences and make sure they didn't screw up.

For example I didn't think realize until right this moment that the global cas counter must be accessed after we read the vptr behind RLock().  It's much better to have explicit data flow.
</blockquote></details>

We should be writing the max CAS counter which we have used so far. The flaw here is to associate that operation with the operation of `vptr`. Those are two different concepts altogether. Because you're uniting the two, it's confusing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

manishrjain:

---

*[kv.go, line 138 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12DUEwnMMRkg-Y2v1:-Kq1CwhPRLS3ClyxjgQ6:b-wkrymm) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L138)):*

We should be writing the max CAS counter which we have used so far. The flaw here is to associate that operation with the operation of `vptr`. Those are two different concepts altogether. Because you're uniting the two, it's confusing.
</blockquote></details>

You can for e.g., store this in a separate key from HEAD. Call it something else, like `!badger!maxcas`. That would be easier to understand.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

srh:

---

*[kv.go, line 138 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12DUEwnMMRkg-Y2v1:-Kq1VqnW6GR5KyjwQCeS:b-43c3m5) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L138)):*

You can for e.g., store this in a separate key from HEAD. Call it something else, like `!badger!maxcas`. That would be easier to understand.
</blockquote></details>

The max cas counter is intrinsically linked to the vptr as written in !badger!head because that's the point at which we start replaying the value log, which we use to recompute the cas counter upon startup.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

manishrjain:

---

*[kv.go, line 138 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12DUEwnMMRkg-Y2v1:-Kq1fbJpnK17Vua_lr21:b-j3t0yg) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L138)):*

The max cas counter is intrinsically linked to the vptr as written in !badger!head because that's the point at which we start replaying the value log, which we use to recompute the cas counter upon startup.
</blockquote></details>

But, unlike the vptr, we take a max of CAS over all the entries. The other way to think about this, is that we just need to max CAS used, period. This is different from the problem of vptr, which has to only store the pointer which the memtable contains entries of. It's not trying to do a max.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

srh:
Review status: 12 of 13 files reviewed at latest revision, 4 unresolved discussions.

---

*[kv.go, line 138 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq12DUEwnMMRkg-Y2v1:-Kq4yYpCD81WreJzVCoI:b-jvucib) ([raw file](https://github.com/dgraph-io/badger/blob/e3129f72b8ebf1ac2fdc5379cc7673590c266b2d/kv.go#L138)):*

But, unlike the vptr, we take a max of CAS over all the entries. The other way to think about this, is that we just need to max CAS used, period. This is different from the problem of vptr, which has to only store the pointer which the memtable contains entries of. It's not trying to do a max.
</blockquote></details>

We're writing the max cas used as of some vptr value.  Or now that I've reverted back to the atomic op implementation, some cas value >= that, and that of all values which won't get replayed in the value log.  It's not the max cas used, because other operations are accessing the cas atomic immediately after.  It's a cas value >= that of those used before the vptr in !badger!head, and <= later values written to the value log.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

manishrjain:
:lgtm: A few small nitpicks. Other than that, looks good. Thanks for your patience working through this.

---

Reviewed 1 of 1 files at r3.

---

*[kv.go, line 141 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq5f9uKuWWilHtvIHbT:-Kq5f9uKuWWilHtvIHbU:b7ru6s7) ([raw file](https://github.com/dgraph-io/badger/blob/47b03a4d20fc5938bba28e18857351ba083bfca1/kv.go#L141)):*
> ```Go
> 	flushChan chan flushTask // For flushing memtables.
> 
> 	// Incremented in the non-concurrently accessed write loop.  But also incremented outside.
> ```

But also "accessed" outside.

---

*[value.go, line 590 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq5fwOxVq9NbW9QKNub:-Kq5fwOxVq9NbW9QKNuc:b-chdqxn) ([raw file](https://github.com/dgraph-io/badger/blob/47b03a4d20fc5938bba28e18857351ba083bfca1/value.go#L590)):*
> ```Go
> 
> type request struct {
> 	// Input values
> ```

The comments can be in the same line as the code.

Entries []*Entry // Input
Ptrs []valuePointer // Output

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130#-:-Kq5g7kci7HExM7RHt_0:bqzv529)*

srh:
*[value.go, line 590 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/130#-Kq5fwOxVq9NbW9QKNub:-Kq5ioG6bd7M23DQSCu8:b-vrirnn) ([raw file](https://github.com/dgraph-io/badger/blob/47b03a4d20fc5938bba28e18857351ba083bfca1/value.go#L590)):*

The comments can be in the same line as the code.

Entries []*Entry // Input
Ptrs []valuePointer // Output
</blockquote></details>

I made the output comment more clear it was referring to all the subsequent fields.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/130)*

srh:
This is in master as of 6abc548a063c75e1bbdffbd4853788429411cf96.

PR #134, opened 2017-07-30
Avoid starting value log replay from too far ahead

srh:
This change fixes a dataloss bug should the process running Badger crash or be killed.

The bug is that we write s.vptr to !badger!head when we should actually write ft.vptr.

We want to write a vptr value that is less than or equal to the last value log offset for the particular memtable we're writing.  That's ft.vptr.  The value `s.vptr` could be overwritten as we're writing to later memtables.

For example, suppose that when `writeRequests` calls `ensureRoomForWrite`, we send a flush task to `flushChan`.  That flushes the preceding (full) s.mt (call it M), along with its corresponding s.vptr (call it X) value.  Then `writeRequests` writes to the new `s.mt` skiplist and *after that* calls `updateOffset` which updates `s.vptr` to the new value Y.  (Thus `s.vptr` is always less than or equal to the last vlog pointer written to `s.mt`.  And Y >= X.)  Suppose Y > X.

The function `flushMemtable` might not read from `flushChan` until after we've written the value Y to `s.vptr`.  Then !badger!head points to Y, and the LSM is written for all vlog entries up to and including X.

That means the writes in the interval (X, Y] would be missed if we replay the value log.

So writes we've acknowledged might be lost, and we might wake up in an inconsistent state by replaying entries after Y.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/134)

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -1018,7 +1018,6 @@ func (s *KV) flushMemtable(lc *y.LevelCloser) error {
 			return nil
 		}
 
-		// TODO: Can this ever be zero?  That'd just be nuts.
 		if !ft.vptr.IsZero() {
 			s.elog.Printf("Storing offset: %+v\n", ft.vptr)
 			offset := make([]byte, 10)
--- a/kv.go
+++ b/kv.go
@@ -533,12 +533,14 @@ func (s *KV) updateOffset(ptrs []valuePointer) {
 			break
 		}
 	}
+	if ptr.IsZero() {
+		return
+	}
 
 	s.Lock()
 	defer s.Unlock()
-	if s.vptr.Less(ptr) {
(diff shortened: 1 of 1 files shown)


manishrjain:
LGTM. Really nice catch! Looks like I introduced this bug early on.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/134#-:-KqMVLPnq1fLxKaT6MVk:b9l7lu4)*

srh:
In master as of 37f7ab3612b7af6579f2b5c8a801cf443334e52e

PR #140, opened 2017-08-02
Manifest file

srh:
Adds a manifest file, which keeps info about value log and table files.

Generally, the main changes are:

- Removed file metadata
- Removed compact log
- Added manifest

Two changes that were not done, but which we might want to do in the future:

- Make the manifest hold basic metadata like smallest, biggest keys (to improve startup times), and file size (for safety).
- Error handling in compaction.  We ignored an error before -- and we still ignore that error.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/140)


srh:
Review status: 0 of 15 files reviewed at latest revision, 15 unresolved discussions.

---

*[level_handler.go, line 175 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_M-uO-rnTAuY2f1XV:-Kq_M-uP9mu8RIMNGn8D:b-vgohr9) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/level_handler.go#L175)):*
> ```Go
> 	s.Lock()
> 	defer s.Unlock()
> 	if len(s.tables) >= s.kv.opt.NumLevelZeroTablesStall {
> ```

An incidental bugfix.  As far as I can tell this comparison was off-by-one.

---

*[levels.go, line 41 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_N6Pa09uA7XYOxSgh:-Kq_N6Pa09uA7XYOxSgi:bg9xfsv) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L41)):*
> ```Go
> 	kv     *KV
> 
> 	nextFileID uint64 // Atomic
> ```

Renamed, is the same thing as maxFileID.  Being particular about the meaning of "max" here.

---

*[levels.go, line 67 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_MGSRGIERHl678nHx:-Kq_MGSRGIERHl678nHy:bgqlb28) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L67)):*
> ```Go
> 			kv.elog.Printf("Table file %d not referenced in MANIFEST\n", id)
> 			filename := table.NewFilename(id, kv.opt.Dir)
> 			err := os.Remove(filename)
> ```

We delete files not found in the manifest.  I was thinking, maybe we should move them to some lost+found.  But 

---

*[levels.go, line 193 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_Mt2DaMtfVQ7XD68o:-Kq_Mt2DaMtfVQ7XD68p:bgwqtgg) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L193)):*
> ```Go
> 			prios := s.pickCompactLevels()
> 			for _, p := range prios {
> 				// TODO: Handle error.
> ```

I made doCompact return errors which it (or its callees) ignored before.  Right now the error is still unhandled here.

---

*[levels.go, line 455 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_NR39-klnzSR9s00C:-Kq_NR39-klnzSR9s00D:b3nqk81) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L455)):*
> ```Go
> }
> 
> func (s *levelsController) reserveCompactionFileIDs(def *compactDef) (uint64, uint64) {
> ```

Moved from compact_log.go.  (Original function returned a `*compaction` with a slice of these file IDs named `toInsert`.)

---

*[levels.go, line 498 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_NqFfDs693JyMTaFj:-Kq_NqFfDs693JyMTaFk:b-lake5n) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L498)):*
> ```Go
> 			return err
> 		}
> ```

The comments here aren't describing anything new.

---

*[levels.go, line 524 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_O7cLMPz7rt46f5bu:-Kq_O7cLMPz7rt46f5bv:bnzd9ip) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L524)):*
> ```Go
> 		// This compaction couldn't be done successfully.
> 		cd.elog.LazyPrintf("\tLOG Compact FAILED with error: %+v: %+v", err, cd)
> 		return err
> ```

This is where we ignored the error before -- which we now return up the stack a bit before ignoring.

---

*[levels.go, line 649 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_OSR1CBjaCcEasVoB:-Kq_OSR1CBjaCcEasVoC:bb3l1l3) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L649)):*
> ```Go
> // get returns the found value if any. If not found, we return nil.
> func (s *levelsController) get(key []byte) (y.ValueStruct, error) {
> 	// It's important that we iterate the levels from 0 on upward.  The reason is, if we iterated
> ```

This comment doesn't explain anything new.

---

*[manifest.go, line 116 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_Qji2V3tEagPEBJur:-Kq_Qji2V3tEagPEBJus:bl4hb9a) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/manifest.go#L116)):*
> ```Go
> }
> 
> type manifestChangeSet struct {
> ```

Generally speaking a manifest file is a sequence of manifestChangeSets.  They atomically apply a sequence of manifestChanges, in order.

I should add a comment to this effect, somewhere here.

---

*[manifest.go, line 353 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_QLThjJLrBVxDXVdS:-Kq_QLThjJLrBVxDXVdT:b-moopx0) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/manifest.go#L353)):*
> ```Go
> }
> 
> func (mcs *manifestChangeSet) Encode(w *bytes.Buffer) {
> ```

There's a lot of binary encoding/decoding logic in this file and I think, maybe there might be a better way to do this in Go.  Or at least a way with less effort.  What would you do?

---

*[manifest_test.go, line 61 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_OcBD4VTrPpxMgTR4:-Kq_OcBD4VTrPpxMgTR5:b-sfhotd) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/manifest_test.go#L61)):*
> ```Go
> }
> 
> func TestManifestBasic(t *testing.T) {
> ```

Stuff below here is moved/renamed from compact_log tests.

---

*[manifest_test.go, line 192 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_Ompw4TkCR3Fdadpd:-Kq_OmpxEHZb8EyJZ0w_:b-rabfmu) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/manifest_test.go#L192)):*
> ```Go
> }
> 
> func TestManifestRewrite(t *testing.T) {
> ```

Except for this which specifically tests the rewrite mechanism of the manifest.

---

*[value.go, line 596 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_PF3SO-yv-ObQx6d7:-Kq_PF3SO-yv-ObQx6d8:bus953l) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/value.go#L596)):*
> ```Go
> func (l *valueLog) Open(kv *KV, opt *Options, mf *manifest) error {
> 	l.dirPath = opt.ValueDir
> 	l.opt = *opt
> ```

I had to move l.kv above openOrCreateFiles to get things working -- and moved l.opt assignment too -- then investigating the performance slowdown is how I found out about the syncWrites bug.

---

*[value_test.go, line 287 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_PbTr7U7L3HJjvVw6:-Kq_PbTr7U7L3HJjvVw7:bjl0fo7) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/value_test.go#L287)):*
> ```Go
> 				y.Check(err)
> 				defer os.RemoveAll(dir)
> 				var mf manifest
> ```

A bug.  Should be bf := createManifest(...).  I'm lucky this test doesn't get a nil pointer dereference.

---

*[y/y.go, line 38 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_Pu0fNJVsT_NSk1Y5:-Kq_Pu0fNJVsT_NSk1Y6:b6mjs6h) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/y/y.go#L38)):*
> ```Go
> )
> 
> // OpenExistingSyncedFile opens an existing file, errors if it doesn't exist.
> ```

I moved every use of OpenSyncedFile off to these specific OpenExisting... and Create... functions that I could.  The only use of OpenSyncedFile left is for the MANIFEST file -- which is what we want.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/140)*

manishrjain:
Reviewed 15 of 15 files at r1.
Review status: all files reviewed at latest revision, 28 unresolved discussions.

---

*[kv.go, line 236 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqb5sVjr4RMsfz9-ZwT:-Kqb5sVjr4RMsfz9-ZwU:b-jwskkw) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/kv.go#L236)):*
> ```Go
> 	go out.flushMemtable(lc) // Need levels controller to be up.
> 
> 	if err = out.vlog.Open(out, &opt, &manifest); err != nil {
> ```

Value log doesn't need to have manifest.

---

*[level_handler.go, line 106 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqb6cjaTVSaDHrlBmz8:-Kqb6cjaTVSaDHrlBmz9:bbkgtho) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/level_handler.go#L106)):*
> ```Go
> // You must call decr() to delete the old tables _after_ writing the update to the manifest.
> func (s *levelHandler) replaceTables(
> 	newTables []*table.Table) error {
> ```

Can fit into last line.

---

*[level_handler.go, line 148 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqb6ydTumpxiPcktH4V:-Kqb6ydTumpxiPcktH4W:bx7wqo0) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/level_handler.go#L148)):*
> ```Go
> 
> 	s.Unlock() // s.Unlock before we DecrRef tables -- that can be slow.
> ```

Remove vert space.

---

*[level_handler.go, line 175 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_M-uO-rnTAuY2f1XV:-Kqb7KXcjWqK0XcHRu_N:b32w0x) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/level_handler.go#L175)):*

An incidental bugfix.  As far as I can tell this comparison was off-by-one.
</blockquote></details>

OK

---

*[levels.go, line 41 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_N6Pa09uA7XYOxSgh:-Kqb7RATcwOM2aIW1ThB:b32w0x) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L41)):*

Renamed, is the same thing as maxFileID.  Being particular about the meaning of "max" here.
</blockquote></details>

OK

---

*[levels.go, line 54 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqb7m3RDfESszoS4JLc:-Kqb7m3RDfESszoS4JLd:b-xjluas) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L54)):*
> ```Go
> )
> 
> func revertToManifest(kv *KV, mf *manifest, idMap map[uint64]struct{}) error {
> ```

Add a comment about what does idMap contain? And how did we generate it.

---

*[levels.go, line 67 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_MGSRGIERHl678nHx:-Kqb7txxCob9nBpYT1iI:bnee6di) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L67)):*

We delete files not found in the manifest.  I was thinking, maybe we should move them to some lost+found.  But 
</blockquote></details>

LGTM. Deletion look OK to me. If users tell us some specific use case where deletion is not the right thing, we can reconsider then.

---

*[levels.go, line 68 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqb883IHBijiJXDZZhJ:-Kqb883IHBijiJXDZZhK:brya7tf) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L68)):*
> ```Go
> 			filename := table.NewFilename(id, kv.opt.Dir)
> 			err := os.Remove(filename)
> 			if err != nil {
> ```

if err := os.Remove(..); err != nil {

---

*[levels.go, line 193 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_Mt2DaMtfVQ7XD68o:-Kqb9n37C8WYZ8yfTWRy:b-v1416c) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L193)):*

I made doCompact return errors which it (or its callees) ignored before.  Right now the error is still unhandled here.
</blockquote></details>

Maybe print them using log.Printf?

---

*[levels.go, line 330 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbBrKOddHBwgT0h_QV:-KqbBrKOddHBwgT0h_QW:b-d39mt) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L330)):*
> ```Go
> 
> 	out := newTables[:i]
> ```

remove vertical space.

---

*[levels.go, line 455 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_NR39-klnzSR9s00C:-KqbCPyVb65dFDIBh57C:balzzbz) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L455)):*

Moved from compact_log.go.  (Original function returned a `*compaction` with a slice of these file IDs named `toInsert`.)
</blockquote></details>

Do we still need to estimate how many tables we'll be generating in advance? Instead, could we just use an atomic counter, and every time we need a new file, it can just give us a new id within compactBuildTables.

We can then write the new files + proposed deletions using `buildChangeSet` to manifest. And then delete the old ones.

Simplifies the code.

---

*[levels.go, line 485 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbDE9KyJzCjo5UtF3O:-KqbDE9KyJzCjo5UtF3P:bunjwzh) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L485)):*
> ```Go
> 		tbl := cd.top[0]
> 
> 		// We write to the manifest _before_ we delete files (and after we created files)
> ```

full stop at the end.

---

*[levels.go, line 487 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbDMZy8kP5iW3ZtVrf:-KqbDMZy8kP5iW3ZtVrg:bo0bz1n) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L487)):*
> We have to delete the table from one level before adding it to the next.

We add to the next level before deleting them in this level. So, the above is contradictory.

---

*[levels.go, line 499 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbDokZgNr5aMkUqHXg:-KqbDokZgNr5aMkUqHXh:b-mutwud) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L499)):*
> ```Go
> 		}
> 
> 		// We have to add to nextLevel before we remove from thisLevel -- in the opposite order
> ```

s/in the opposite order/otherwise, it would cause a bug for a moment where reads would see ...

---

*[levels.go, line 528 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbAvMEIufm-oys49GN:-KqbAvMEIufm-oys49GO:bqseyk) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L528)):*
> ```Go
> 	defer func() {
> 		if decErr := decr(); err == nil {
> 			err = decErr
> ```

Add a comment `// only assign to err, if it's not nil.`

Might sound obvious, but the typical pattern of `if err := foo(); err != nil` isn't followed here, which causes initial confusion.

---

*[levels.go, line 599 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbEgRIM3DtO8N3aEl5:-KqbEgRIM3DtO8N3aEl6:byolkp8) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L599)):*
> ```Go
> 
> func (s *levelsController) addLevel0Table(t *table.Table) error {
> 	// We want to update the manifest _before_ we actually add the table to memory.
> ```

add the table to level 0 (which is on disk).

---

*[manifest.go, line 38 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbFf_OSUJlHQjc2IB8:-KqbFf_OSUJlHQjc2IB9:b-21no8i) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/manifest.go#L38)):*
> ```Go
> 	levels        []levelManifest
> 	tables        map[uint64]tableManifest
> 	valueLogFiles map[uint64]struct{}
> ```

value log files are not required. Even if we have extras lying around, the vlog GC would take care of them.

---

*[manifest.go, line 353 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_QLThjJLrBVxDXVdS:-KqbG7hq2sEN5q8emOuj:b-w6xnv9) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/manifest.go#L353)):*

There's a lot of binary encoding/decoding logic in this file and I think, maybe there might be a better way to do this in Go.  Or at least a way with less effort.  What would you do?
</blockquote></details>

Yeah, this is too much code.

I'd say use protobufs as much as you can. They'd cut down on the amount of custom encoding we need to do and maintain by hand. Also, would make this code more legible.

The state machinery can all go to protobufs. It's only the deltas that you are appending -- even they can be encoded by prepending the length of byte slice, followed by the proto byte slice; which would be a lot simpler logic.

---

*[value.go, line 316 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbGdfTG7F0s6pVXEFD:-KqbGdfTG7F0s6pVXEFE:b7ok2j8) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/value.go#L316)):*
> ```Go
> 	f.fd.Close() // close file previous to remove it
> 
> 	vlog.kv.manifest.addChanges(manifestChangeSet{
> ```

No need. Here and below.

---

*[y/y.go, line 38 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_Pu0fNJVsT_NSk1Y5:-KqbGwkVzfusiiNbWcMU:b-6cdi75) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/y/y.go#L38)):*

I moved every use of OpenSyncedFile off to these specific OpenExisting... and Create... functions that I could.  The only use of OpenSyncedFile left is for the MANIFEST file -- which is what we want.
</blockquote></details>

I assume you handle the case where there's no MANIFEST file (fresh new dir).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/140)*

manishrjain:
Looked like a big scary change, but it's done nicely!

I'd cut down on the manifest encoding/decoding significantly by finding ways to use protobufs. And I don't think value log needs to be tracked in manifest.

---

Review status: all files reviewed at latest revision, 25 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/140#-:-KqbH5A1FAo0gcqzgtGn:b-h1wxme)*

srh:
Review status: 7 of 14 files reviewed at latest revision, 25 unresolved discussions.

---

*[kv.go, line 236 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqb5sVjr4RMsfz9-ZwT:-KqeFhtdm9ox3TrKVpSn:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/kv.go#L236)):*

Value log doesn't need to have manifest.
</blockquote></details>

Done.

---

*[level_handler.go, line 106 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqb6cjaTVSaDHrlBmz8:-Kqe55xsszXjyza7fOe6:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/level_handler.go#L106)):*

Can fit into last line.
</blockquote></details>

Done.

---

*[level_handler.go, line 148 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqb6ydTumpxiPcktH4V:-Kqe55bjDhtW0rAWXHBy:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/level_handler.go#L148)):*

Remove vert space.
</blockquote></details>

Done.

---

*[levels.go, line 54 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqb7m3RDfESszoS4JLc:-Kqe5dwGd5uQzcM1yBd-:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L54)):*

Add a comment about what does idMap contain? And how did we generate it.
</blockquote></details>

Done.

---

*[levels.go, line 68 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqb883IHBijiJXDZZhJ:-Kqe5jloyrIrWb6ghBAN:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L68)):*

if err := os.Remove(..); err != nil {
</blockquote></details>

Done.

---

*[levels.go, line 193 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_Mt2DaMtfVQ7XD68o:-KqeK8VPdzX_lO9Ge2p_:b-y5yw24) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L193)):*

Maybe print them using log.Printf?
</blockquote></details>

With elog.LazyPrintf, apparently.  Did the logging in doCompact (moved some logging from its callee).

---

*[levels.go, line 330 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbBrKOddHBwgT0h_QV:-Kqe5sqJ6PSzqpUTLi5z:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L330)):*

remove vertical space.
</blockquote></details>

Done.

---

*[levels.go, line 455 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_NR39-klnzSR9s00C:-KqeKKjjPtNM4xpmrCOR:bmftk89) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L455)):*

Do we still need to estimate how many tables we'll be generating in advance? Instead, could we just use an atomic counter, and every time we need a new file, it can just give us a new id within compactBuildTables.

We can then write the new files + proposed deletions using `buildChangeSet` to manifest. And then delete the old ones.

Simplifies the code.
</blockquote></details>

Main reason I didn't is that we end up allocating a slice up-front and having multiple goroutines write to an element (concurrently).  But we could clean up the code by having a slice of pointers.

---

*[levels.go, line 485 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbDE9KyJzCjo5UtF3O:-Kqe5zVg6VOFQiicwPxc:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L485)):*

full stop at the end.
</blockquote></details>

Done.

---

*[levels.go, line 487 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbDMZy8kP5iW3ZtVrf:-Kqe7NMukjTuOShijLiX:b-cailir) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L487)):*

> We have to delete the table from one level before adding it to the next.

We add to the next level before deleting them in this level. So, the above is contradictory.
</blockquote></details>

Made the comment a bit less confusing.

---

*[levels.go, line 499 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbDokZgNr5aMkUqHXg:-Kqe7wSax8QzKWrnb8Dy:b6xx9ab) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L499)):*

s/in the opposite order/otherwise, it would cause a bug for a moment where reads would see ...
</blockquote></details>

Done, reworded a bit differently.

---

*[levels.go, line 528 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbAvMEIufm-oys49GN:-Kqe8KPAv3GiqXcjJU19:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L528)):*

Add a comment `// only assign to err, if it's not nil.`

Might sound obvious, but the typical pattern of `if err := foo(); err != nil` isn't followed here, which causes initial confusion.
</blockquote></details>

Done.

---

*[levels.go, line 599 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbEgRIM3DtO8N3aEl5:-Kqe9EsN67gDpofxtTXM:b72r6zl) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/levels.go#L599)):*

add the table to level 0 (which is on disk).
</blockquote></details>

I'm actually referring to the in-memory data structures.  Edited comment to clarify this.

---

*[manifest.go, line 38 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbFf_OSUJlHQjc2IB8:-KqeFtyjh0PRq3Alb4Qy:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/manifest.go#L38)):*

value log files are not required. Even if we have extras lying around, the vlog GC would take care of them.
</blockquote></details>

Done.

---

*[manifest.go, line 353 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_QLThjJLrBVxDXVdS:-KqeFuwIeydF1gqxpA41:b-dx0sfx) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/manifest.go#L353)):*

Yeah, this is too much code.

I'd say use protobufs as much as you can. They'd cut down on the amount of custom encoding we need to do and maintain by hand. Also, would make this code more legible.

The state machinery can all go to protobufs. It's only the deltas that you are appending -- even they can be encoded by prepending the length of byte slice, followed by the proto byte slice; which would be a lot simpler logic.
</blockquote></details>

We don't use protobufs anywhere else in Badger, and I think we shouldn't use them here.  One reason is it would be our only protobuf dependency.  In general, in most places in Badger I think we won't want to use protobufs.  Also, with the value log removal we have much less encoding logic here.

---

*[value.go, line 316 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqbGdfTG7F0s6pVXEFD:-KqeGsP_07wDnjl7lB4Y:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/value.go#L316)):*

No need. Here and below.
</blockquote></details>

Done.

---

*[y/y.go, line 38 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_Pu0fNJVsT_NSk1Y5:-Kqe9qJnJEzGgq4n-BCr:b-xko1ui) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/y/y.go#L38)):*

I assume you handle the case where there's no MANIFEST file (fresh new dir).
</blockquote></details>

Right, it uses OpenSyncedFile, which creates an empty one.  (Which is correct and ideal behavior.)

---

*[value_test.go, line 287 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_PbTr7U7L3HJjvVw6:-KqeGx1KvXD5w4l01fMk:baq8v7w) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/value_test.go#L287)):*

A bug.  Should be bf := createManifest(...).  I'm lucky this test doesn't get a nil pointer dereference.
</blockquote></details>

Moot now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/140)*

manishrjain:
I still think we should cut down the encoding / decoding effort in manifest for the sake of future readability. You could use gob encoder, if you want to avoid protobufs. But, we should use something -- instead of hand coding all this. Rest looks good.

---

Reviewed 8 of 8 files at r2.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[levels.go, line 265 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqf52-GdqL_Kmy94z0C:-Kqf52-GdqL_Kmy94z0D:bbvb1k) ([raw file](https://github.com/dgraph-io/badger/blob/081842718168d1623563822cca99b9aefeca1210/levels.go#L265)):*
> ```Go
> 
> 	// Start generating new tables.
> 	newTables := []**table.Table{}
> ```

Why pointer to pointer? It's not required.

var newTables []*table.Table, or newTables := make([]*table.Table)

In the goroutine, you can pass in the index, i, extend the array; which you can use to write the generated table to.

---

*[levels.go, line 300 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqf6LqwCtoGmK2aO6lO:-Kqf6Lqx7HMAy3gzYiyq:b9vxad8) ([raw file](https://github.com/dgraph-io/badger/blob/081842718168d1623563822cca99b9aefeca1210/levels.go#L300)):*
> ```Go
> 			}
> 
> 			*tablePtr, err = table.OpenTable(fd, s.kv.opt.MapTablesTo)
> ```

This is buggy, because the tablePtr would have changed by the time this instruction happens. So, better to pass in index (the same way as before) and still do the assign table to a slice index thing.

---

*[manifest.go, line 353 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kq_QLThjJLrBVxDXVdS:-Kqf7OwOEl01MiQKWK_1:b6bva8k) ([raw file](https://github.com/dgraph-io/badger/blob/4e9f97d61c0285409d7b313613c21dd90205a7fd/manifest.go#L353)):*

We don't use protobufs anywhere else in Badger, and I think we shouldn't use them here.  One reason is it would be our only protobuf dependency.  In general, in most places in Badger I think we won't want to use protobufs.  Also, with the value log removal we have much less encoding logic here.
</blockquote></details>

I think adding a dep is okay, if it's going to cut down our code significantly. The problem I see with this encoding here, is that once written, nobody else would want to touch it -- it's menial and cumbersome; which is what protos are supposed to tackle.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/140#-:-Kqf8EqHIbYuyVo6tdoF:b-ji8ny7)*

srh:
OK, it now uses protobufs.

---

Review status: 11 of 17 files reviewed at latest revision, 11 unresolved discussions.

---

*[levels.go, line 265 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqf52-GdqL_Kmy94z0C:-Kqf93z7d0SaMEzTEDoU:ba43rad) ([raw file](https://github.com/dgraph-io/badger/blob/081842718168d1623563822cca99b9aefeca1210/levels.go#L265)):*

Why pointer to pointer? It's not required.

var newTables []*table.Table, or newTables := make([]*table.Table)

In the goroutine, you can pass in the index, i, extend the array; which you can use to write the generated table to.
</blockquote></details>

You can't reallocate the slice while other goroutines index into it.  They'll race.

---

*[levels.go, line 300 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqf6LqwCtoGmK2aO6lO:-Kqf9Gyqr5LnAqyu8s_W:ba0fd9) ([raw file](https://github.com/dgraph-io/badger/blob/081842718168d1623563822cca99b9aefeca1210/levels.go#L300)):*

This is buggy, because the tablePtr would have changed by the time this instruction happens. So, better to pass in index (the same way as before) and still do the assign table to a slice index thing.
</blockquote></details>

tablePtr is lexically scoped, so it gets captured by the closure.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/140#-:-KqfM6okgWTFKOBTyS9y:bbg82z4)*

manishrjain:
Thanks. Almost there. I think we can get rid of the double pointers -- and use a generously pre allocated slice, because the compact code path isn't performance critical.

---

Reviewed 6 of 6 files at r3.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[levels.go, line 265 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqf52-GdqL_Kmy94z0C:-KqfZNB3agW9PCTamC6d:b-8zqcqe) ([raw file](https://github.com/dgraph-io/badger/blob/081842718168d1623563822cca99b9aefeca1210/levels.go#L265)):*

You can't reallocate the slice while other goroutines index into it.  They'll race.
</blockquote></details>

Then just pre-allocate a slice of 20 (or even 50, doesn't matter). We would never be writing that many tables. And doing extra allocation here is OK, because this code path isn't performance critical.

Double pointers are a C/C++ thing -- Go shouldn't be using them.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/140#-:-Kqf_Yi6mJwb9OIo-FBX:bqfnl75)*

srh:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[levels.go, line 265 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqf52-GdqL_Kmy94z0C:-KqffDzNuOiGggFJFngj:bapw7p) ([raw file](https://github.com/dgraph-io/badger/blob/081842718168d1623563822cca99b9aefeca1210/levels.go#L265)):*

Then just pre-allocate a slice of 20 (or even 50, doesn't matter). We would never be writing that many tables. And doing extra allocation here is OK, because this code path isn't performance critical.

Double pointers are a C/C++ thing -- Go shouldn't be using them.
</blockquote></details>

I don't get how there is a limit to the number of overlapping tables in the lower level.  We pick a table from thisLevel, then see how many overlap it in nextLevel.  For example, if key/values were generally inserted in increasing order, but then one day a handful of older key/values were updated, we could have one trailing table which percolates down and overlaps thousands of tables at the bottom level.

Or maybe I'm missing something, but that's still making assumptions about the value of LevelSizeMultiplier.

I don't see what is wrong with having pointers to pointers.  The values we're using are of type `*table.Table`, and we need to provide to each goroutine a place to write such a value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/140)*

srh:
Review status: 16 of 17 files reviewed at latest revision, 10 unresolved discussions.

---

*[levels.go, line 332 at r4](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqfpC0TkesFfv4GvfQt:-KqfpC0TkesFfv4GvfQu:babq7vi) ([raw file](https://github.com/dgraph-io/badger/blob/74ab43b117c848bbfb0c590e71aff53662700010/levels.go#L332)):*
> ```Go
> 		// -- we're the only holders of a ref).
> 		for j := 0; j < i; j++ {
> 			if newTables[j] != nil {
> ```

There was a bug here -- at least one of the values would be nil.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/140)*

srh:
Review status: 16 of 17 files reviewed at latest revision, 10 unresolved discussions.

---

*[levels.go, line 265 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/140#-Kqf52-GdqL_Kmy94z0C:-KqfpIVP5kJJjp8aA-U9:biqrmbg) ([raw file](https://github.com/dgraph-io/badger/blob/081842718168d1623563822cca99b9aefeca1210/levels.go#L265)):*

I don't get how there is a limit to the number of overlapping tables in the lower level.  We pick a table from thisLevel, then see how many overlap it in nextLevel.  For example, if key/values were generally inserted in increasing order, but then one day a handful of older key/values were updated, we could have one trailing table which percolates down and overlaps thousands of tables at the bottom level.

Or maybe I'm missing something, but that's still making assumptions about the value of LevelSizeMultiplier.

I don't see what is wrong with having pointers to pointers.  The values we're using are of type `*table.Table`, and we need to provide to each goroutine a place to write such a value.
</blockquote></details>

Now it's using a chan.  Combined it with the error chan.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/140)*

manishrjain:
:lgtm: Great work!

---

Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[levels.go, line 344 at r4](https://reviewable.io:443/reviews/dgraph-io/badger/140#-KqjkKDiZz8UXb_r0INd:-KqjkKDiZz8UXb_r0INe:b-xd2km2) ([raw file](https://github.com/dgraph-io/badger/blob/74ab43b117c848bbfb0c590e71aff53662700010/levels.go#L344)):*
> ```Go
> 	})
> 
> 	out := newTables[:i]
> ```

This out assignment and limiting to i isn't required. Because len(newTables) == i.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/140#-:-KqjkU9x_xfvvgaSK-VA:bolbkdq)*

srh:
This is in master as of f3d51527f1d3a5eb4cd91de20abbed195510d444.

PR #148, opened 2017-08-07
Add Checksum to value log entries

peterstace:
Changes:

* There is now a 32 bit CRC checksum at the end of each value log entry. So the format is now `[header][key][value][crc32]`.
* It's checked during value log replay. If an mismatch is found, the value log is truncated at the end of the last good entry.
* It's also checked during a normal value log read.
* When doing the value log replay, we also truncate the log if a partial entry is found (i.e. we reach EOF while in the middle of an entry).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/148)

### DIFF
--- a/value.go
+++ b/value.go
@@ -180,10 +180,13 @@ func (f *logFile) iterate(offset uint32, fn logEntry) error {
 		}
 		crc := binary.BigEndian.Uint32(crcBuf[:])
 		if crc != hash.Sum32() {
-			// TODO(peter): What to do when CRC doesn't match?
-			// 1. Return error.
-			// 2. Break out of loop (returning nil).
-			// Do we need to erase the remainder of the log file? Otherwise it would be broken 'forever'.
+			f.Lock()
+			if err := f.fd.Truncate(int64(recordOffset)); err != nil {
+				f.Unlock()
+				return err
+			}
+			f.Unlock()
+			break
 		}
 
 		var vp valuePointer
--- a/value.go
+++ b/value.go
@@ -138,14 +138,18 @@ func (f *logFile) iterate(offset uint32, fn logEntry) error {
 	k := make([]byte, 1<<10)
 	v := make([]byte, 1<<20)
 
+	truncate := false
 	recordOffset := offset
 	for {
 
 		hash := crc32.New(entryHashTable)
 		tee := io.TeeReader(reader, hash)
 
-		if err = read(tee, hbuf[:]); err == io.EOF {
-			break
+		if err = read(tee, hbuf[:]); err != nil {
+			if err == io.EOF {
+				break
+			}
+			return err
 		}
 
 		var e Entry
@@ -164,6 +168,10 @@ func (f *logFile) iterate(offset uint32, fn logEntry) error {
 		e.Value = v[:vl]
 
 		if err = read(tee, e.Key); err != nil {
+			if err == io.EOF {
+				truncate = true
+				break
+			}
 			return err
 		}
 		e.Meta = h.meta
@@ -171,21 +179,24 @@ func (f *logFile) iterate(offset uint32, fn logEntry) error {
 		e.casCounter = h.casCounter
 		e.CASCounterCheck = h.casCounterCheck
 		if err = read(tee, e.Value); err != nil {
+			if err == io.EOF {
+				truncate = true
+				break
+			}
 			return err
 		}
 
 		var crcBuf [entryHashSize]byte
 		if err = read(reader, crcBuf[:]); err != nil {
+			if err == io.EOF {
+				truncate = true
+				break
+			}
 			return err
 		}
 		crc := binary.BigEndian.Uint32(crcBuf[:])
 		if crc != hash.Sum32() {
-			f.Lock()
-			if err := f.fd.Truncate(int64(recordOffset)); err != nil {
-				f.Unlock()
-				return err
-			}
-			f.Unlock()
+			truncate = true
 			break
 		}
 
@@ -204,6 +215,16 @@ func (f *logFile) iterate(offset uint32, fn logEntry) error {
 			return y.Wrap(err)
 		}
 	}
+
+	if truncate {
+		f.Lock()
+		if err := f.fd.Truncate(int64(recordOffset)); err != nil {
+			f.Unlock()
+			return err
+		}
+		f.Unlock()
+	}
+
 	return nil
 }
 
--- a/value.go
+++ b/value.go
@@ -105,7 +105,7 @@ func (lf *logFile) sync() error {
 
 var errStop = errors.New("Stop iteration")
 
-var entryHashTable = crc32.MakeTable(crc32.Koopman)
+var entryHashTable = crc32.MakeTable(crc32.Castagnoli)
 
 const entryHashSize = 4
 
--- a/value.go
+++ b/value.go
@@ -104,7 +104,6 @@ func (lf *logFile) sync() error {
 }
 
 var errStop = errors.New("Stop iteration")
-
 var entryHashTable = crc32.MakeTable(crc32.Castagnoli)
 
 const entryHashSize = 4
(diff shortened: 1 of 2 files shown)


srh:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[value.go, line 150 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/148#-KqyowsRhj2UbudyXFkZ:-KqyowsRhj2UbudyXFk_:b-5i3mwy) ([raw file](https://github.com/dgraph-io/badger/blob/c13f2441a32288acdc0945b159df8cdf3920ec16/value.go#L150)):*
> ```Go
> 		if err = read(tee, hbuf[:]); err != nil {
> 			if err == io.EOF {
> 				break
> ```

This call to the local function read() could read a few bytes and then return with io.EOF.  The result is that we have to truncate the file.  (And the bug is that we don't.)  We could use io.ReadFull instead, or we could always truncate, to address this.

---

*[value_test.go, line 277 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/148#-Kqyrt_-ZXko1OkU9hJ6:-Kqyrt_-ZXko1OkU9hJ7:b-l2rq6c) ([raw file](https://github.com/dgraph-io/badger/blob/c13f2441a32288acdc0945b159df8cdf3920ec16/value_test.go#L277)):*
> ```Go
> }
> 
> func TestPatialAppendToValueLog(t *testing.T) {
> ```

Partial

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/148)*

peterstace:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[value.go, line 150 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/148#-KqyowsRhj2UbudyXFkZ:-KqzCtOfik8iFjCk9Zph:bfb6y6i) ([raw file](https://github.com/dgraph-io/badger/blob/c13f2441a32288acdc0945b159df8cdf3920ec16/value.go#L150)):*

This call to the local function read() could read a few bytes and then return with io.EOF.  The result is that we have to truncate the file.  (And the bug is that we don't.)  We could use io.ReadFull instead, or we could always truncate, to address this.
</blockquote></details>

Fixed. I just got rid of the `read` function entirely handle the `io.EOF` and `io.UnexpectedEOF` errors appropriately in each case (only different for the header).

---

*[value_test.go, line 277 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/148#-Kqyrt_-ZXko1OkU9hJ6:-KqzCjx2O6G4-_w9nbFU:b-886gsx) ([raw file](https://github.com/dgraph-io/badger/blob/c13f2441a32288acdc0945b159df8cdf3920ec16/value_test.go#L277)):*

Partial
</blockquote></details>

Fixed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/148)*

srh:
Everything else looks correct to my eyes.  My main other concern is performance.

The general question is how much should we care -- disks are slow, right?  I'm not so sure about the future though, what with NVM devices.

Going by this:

https://golang.org/src/hash/crc32/crc32_amd64.go

It looks like Golang supports the amd64 CRC32 instruction if we use Castagnoli instead of Koopman.  This would make it about as fast as using Adler32 or Fletcher checksums on that platform.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/148#-:-KqzJ2VIzsmfOYCgrNYL:bpdpcpj)*

peterstace:
The whole code path for Castagnoli seems more specialised (short cuts in a few places), so using that instead.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/148#-:-KqzYUYe-n_Gbc0sHkIM:byccpnc)*

srh:
:lgtm:

---

Reviewed 1 of 2 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/148#-:-Kqz_1_Ox_Cwlnz8bLPL:bnfp4nl)*

manishrjain:
:lgtm:

---

Reviewed 1 of 2 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[value.go, line 107 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/148#-KqzaRu0P2Klx6Tx3-HS:-KqzaRu0P2Klx6Tx3-HT:b-yemun) ([raw file](https://github.com/dgraph-io/badger/blob/71fa0539b19f6c8418c2dd202177f260ffe7f766/value.go#L107)):*
> ```Go
> 
> var errStop = errors.New("Stop iteration")
> ```

Remove vertical space here.

---

*[value.go, line 194 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/148#-Kqzc9hw9vKDNTimAlFF:-Kqzc9hw9vKDNTimAlFG:b-2jo5yg) ([raw file](https://github.com/dgraph-io/badger/blob/71fa0539b19f6c8418c2dd202177f260ffe7f766/value.go#L194)):*
> ```Go
> 		var vp valuePointer
> 
> 		recordOffset += uint32(hlen + kl + vl + entryHashSize)
> ```

Duplication here. Just make sure that there's no special reasoning why they should be two separate calculations.

---

*[value.go, line 209 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/148#-Kqzci5LMyXm4mbyzhRk:-Kqzci5LMyXm4mbyzhRl:b7z690v) ([raw file](https://github.com/dgraph-io/badger/blob/71fa0539b19f6c8418c2dd202177f260ffe7f766/value.go#L209)):*
> ```Go
> 
> 	if truncate {
> 		f.Lock()
> ```

No need for lock acquisition.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/148#-:-KqzdfXi3OIcgxEw-jFh:bnfp4nl)*

peterstace:
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[value.go, line 107 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/148#-KqzaRu0P2Klx6Tx3-HS:-Kqzs4Qt-RKfTZN3KC_H:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/71fa0539b19f6c8418c2dd202177f260ffe7f766/value.go#L107)):*

Remove vertical space here.
</blockquote></details>

Done.

---

*[value.go, line 194 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/148#-Kqzc9hw9vKDNTimAlFF:-Kqzs6qlA4kgvxfWYgSF:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/71fa0539b19f6c8418c2dd202177f260ffe7f766/value.go#L194)):*

Duplication here. Just make sure that there's no special reasoning why they should be two separate calculations.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/148)*

peterstace:
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[value.go, line 209 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/148#-Kqzci5LMyXm4mbyzhRk:-Kqzu2sWAGQhpW2osIiP:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/71fa0539b19f6c8418c2dd202177f260ffe7f766/value.go#L209)):*

No need for lock acquisition.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/148)*

PR #152, opened 2017-08-10
Fix condition for value log GC threshold

peterstace:
Documentation on the config option:

```
// Run value log garbage collection if we can reclaim at least this much space. This is a ratio.
```

Since the behaviour is difficult for the end user to see, I think it makes more sense to change the behaviour rather than the documentation (user probably didn't notice there was a problem).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/152)


manishrjain:
Lgtm.

PR #175, opened 2017-08-22
Add FillItem method.

### LINKED ISSUE #166: Allow filling values selectively during key-only iteration
Currently, Badger spawns a new goroutine and asynchronously reads values during iteration. In my application, I am reading a large number of k/v pairs, where the values are small (<100 byte), and every value is read almost immediately after advancing the iterator. Turning the read into a synchronous read (by calling it.kv.fillItem in Iterator.Fill directly), I could reduce wall time from 33s to 30s, and CPU usage from 460% to 120%. The reduction in runtime is due to the various overheads of using goroutines: memory allocations, locking and scheduling. The 33s and 30s figures are for the entire runtime of the application, which processes the values read. I can provide cleaner numbers if required.

In addition to my scenario, I can think of another one where the current behavior would be problematic: huge values with sparse reads during iteration. In that scenario, values would be large, and only a small subset of iterated items would have their value read. In the current implementation, this would lead to reading a lot of data unnecessarily, and to Next blocking until the (uninteresting) value has been read fully.

To address both scenarios at once, I propose a change to the semantics of the FetchValues option. Instead of requiring FetchValues == true for any reading of values to occur, it should only cause asynchronous reading. If FetchValues == false, a call to KVItem.Value should cause a synchronous read of the data.

**Edit:** I realize that prefetching interacts with this, both affecting the measured times and possibly requiring special consideration when making this change.

deepakjois:
Fixes #166.

We now expose a public FillItem method of KVItem struct which allows for filling values selectively during key-only iteration.

A few caveats here:

1. This requires a valid item returned during a Badger iteration. If the item is not valid, the behavior is undefined.

2. I chose acquire a lock before calling the private `fillitem` method. This may not be strictly necessary. But I wanted to avoid a misbehaving caller which may call this method even from a key-value iteration. Calling this method from an ongoing key-value iteration can interact with the prefetching.

Please let me know your thoughts, @manishrjain @srh

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/175)

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -175,6 +175,7 @@ func (it *Iterator) fill(item *KVItem) {
 	item.casCounter = vs.CASCounter
 	item.key = y.Safecopy(item.key, it.iitr.Key())
 	item.vptr = y.Safecopy(item.vptr, vs.Value)
+	item.val = nil
 	if it.opt.FetchValues {
 		item.wg.Add(1)
 		go func() {
--- a/kv.go
+++ b/kv.go
@@ -454,6 +454,27 @@ func (s *KV) getMemTables() ([]*skl.Skiplist, func()) {
 	}
 }
 
+// FillItem populates item with a value.
+//
+// item must be a valid KVItem returned by Badger during iteration. This method
+// could be used to fetch values explicitly during a key-only iteration
+// (FetchValues is set to false). It is useful for example, if values are
+// required for some keys only.
+//
+// This method should not be called when iteration is performed with
+// FetchValues set to true, as it will cause additional copying.
+//
+// Multiple calls to this method will result in multiple copies from the value
+// log. It is the caller’s responsibility to make sure they don’t call this
+// method more than once.
+func (s *KV) FillItem(item *KVItem) error {
+	// Wait for any pending fill operations to finish.
+	item.wg.Wait()
+	item.wg.Add(1)
+	defer item.wg.Done()
+	return s.fillItem(item)
+}
+
 func (s *KV) fillItem(item *KVItem) error {
 	if item.meta == 0 && item.vptr == nil {
 		item.val = nil // key not found
--- a/kv_test.go
+++ b/kv_test.go
@@ -873,3 +873,52 @@ func TestIteratorPrefetchSize(t *testing.T) {
 		require.Equal(t, 100, c)
 	}
 }
+
+func TestFillItem(t *testing.T) {
+	dir, err := ioutil.TempDir("", "badger")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+	kv, err := NewKV(getTestOptions(dir))
+	require.NoError(t, err)
+	defer kv.Close()
+
+	bkey := func(i int) []byte {
+		return []byte(fmt.Sprintf("%09d", i))
+	}
+	bval := func(i int) []byte {
+		return []byte(fmt.Sprintf("%025d", i))
+	}
+
+	n := 1000
+	for i := 0; i < n; i++ {
+		kv.Set(bkey(i), bval(i), 0)
+	}
+	opt := IteratorOptions{}
+	opt.FetchValues = false
+
+	it := kv.NewIterator(opt)
+	var count int
+	for it.Rewind(); it.Valid(); it.Next() {
+		item := it.Item()
+		val := item.Value()
+		require.Nil(t, val) // Should initially be nil
+		kv.FillItem(item)
+		val = item.Value()
+		require.EqualValues(t, bval(count), string(val))
+		count++
+	}
+	require.EqualValues(t, n, count)
+
+	opt.FetchValues = true
+	it = kv.NewIterator(opt)
+	count = 0
+	for it.Rewind(); it.Valid(); it.Next() {
+		item := it.Item()
+		val1 := item.Value()
+		require.EqualValues(t, bval(count), string(val1))
+		kv.FillItem(item)
+		val2 := item.Value()
+		require.EqualValues(t, val1, val2)
+		count++
+	}
+}


deepakjois:
I think this might be cleaner if we refactor the code a little bit, to keep the waitgroup logic in a single place.

srh:
I don't follow -- what lock do you acquire?

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/175#-:-Ks82BXflv6I3lsDyn2a:bazs5e1)*

deepakjois:
I mean I use the waitgroup methods `wg.Wait` to wait for any pending fill to finish, and `wg.Add` before I call the fillItem method. It seems necessary to do this now that we make FillItem public, as we dont control how and when it is called.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/175#-:-Ks83yLAgh-Ma1TmSpqo:b-j8witw)*

manishrjain:
Looks good to me, but do add unit tests and verify that this behaves as expected and doesn't add any deadlocks etc. Leave it to @srh for the final approval.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/175#-:-Ks8Tz2QkXpNVChoIjR8:bv6fsop)*

deepakjois:
I will add a test.

srh:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[iterator.go, line 179 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/175#-KsBYqCo2PiB2lTOeBQT:-KsBYqCo2PiB2lTOeBQU:b-q68swf) ([raw file](https://github.com/dgraph-io/badger/blob/d27560fea8184a94fb403692a66e3fb5e9bf5a4f/iterator.go#L179)):*
> ```Go
> 	item.vptr = y.Safecopy(item.vptr, vs.Value)
> 	if it.opt.FetchValues {
> 		item.wg.Wait() // wait for any pending fill
> ```

This should be unnecessary, right?  We only call `fill` on items for which the waitgroup's counter is zero.  They're returned by `newItem`, and it's true of both items in `it.waste` and newly constructed items that `wg`'s counter must be zero.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/175)*

deepakjois:
This is now ready for another review.
* @srh, addressed your comment.
* Made another change (as a separate commit) to reset items from waste list before giving them back.

srh:
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[iterator.go, line 122 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/175#-KsGAfNST9OD4JCkDPr4:-KsGAfNTW2TwfLplBnQD:b1483ta) ([raw file](https://github.com/dgraph-io/badger/blob/7a33ad4ae6bde7158d4af219a24b4d7a86c2b1fa/iterator.go#L122)):*
> ```Go
> 		item = &KVItem{slice: new(y.Slice)}
> 	}
> 	// Clean up old values
> ```

We deliberately reuse the slices pointed at by key and vptr, in `fill`, using `Safecopy`.  So setting the fields to nil will have the effect of making us reallocate the slices and also re-initialize their content to zero.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/175)*

deepakjois:
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[iterator.go, line 122 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/175#-KsGAfNST9OD4JCkDPr4:-KsH2cUJo5UF_5B2YDSx:b-5dpdht) ([raw file](https://github.com/dgraph-io/badger/blob/7a33ad4ae6bde7158d4af219a24b4d7a86c2b1fa/iterator.go#L122)):*

We deliberately reuse the slices pointed at by key and vptr, in `fill`, using `Safecopy`.  So setting the fields to nil will have the effect of making us reallocate the slices and also re-initialize their content to zero.
</blockquote></details>

Yes, I missed that. I should not set item.key and item.vptr to nil. But setting everything else to nil should be fine. In fact, item.val _needs_ to be set to nil.

Maybe it is better if I just set item.val explicitly to nil in the Iterator.fill method below. I think it will be a cleaner code change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/175)*

deepakjois:
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[iterator.go, line 122 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/175#-KsGAfNST9OD4JCkDPr4:-KsH4hYixksZ_wwK3kq8:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/7a33ad4ae6bde7158d4af219a24b4d7a86c2b1fa/iterator.go#L122)):*
<details><summary><i>Previously, deepakjois (Deepak Jois) wrote…</i></summary><blockquote>

Yes, I missed that. I should not set item.key and item.vptr to nil. But setting everything else to nil should be fine. In fact, item.val _needs_ to be set to nil.

Maybe it is better if I just set item.val explicitly to nil in the Iterator.fill method below. I think it will be a cleaner code change.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/175)*

srh:
:lgtm:

---

Reviewed 1 of 2 files at r1, 1 of 2 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/175#-:-KsMBRzSrGoPTxtVyYra:bnfp4nl)*

PR #177, opened 2017-08-23
Fix: use write lock in valueLog.rewrite()

gavriva:
Hi guys! I think it is a bug, valueLog.rewrite() modifies files array so it must use write lock.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/177)

### DIFF
--- a/value.go
+++ b/value.go
@@ -312,7 +312,7 @@ func (vlog *valueLog) rewrite(f *logFile) error {
 	elog.Printf("Removing fid: %d", f.fid)
 	// Entries written to LSM. Remove the older file now.
 	{
-		vlog.RLock()
+		vlog.Lock()
 		idx := sort.Search(len(vlog.files), func(idx int) bool {
 			return vlog.files[idx].fid >= f.fid
 		})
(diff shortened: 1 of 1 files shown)


srh:
Thanks for noticing that.  It looks like we have another mistake in that code too -- we can return an error without unlocking the value log.  Fortunately the error ought to be logically impossible (if the DB directory isn't tampered with).

Also, a few lines below that code, we call `f.fd.Close()` on the old value log file.  What if something's using the `valueLog` object (to read from the descriptor) while that happens?  You would think there would be some sort of refcount, the way `table.Table` has.

I'm working on something else right now so I'll get back to this PR later.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/177#-:-KsGCyyogdXSBKxK_ZXU:b-1asgbj)*

manishrjain:
:lgtm: Nice catch! Thanks for the fix.

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/177#-:-KsbpyR55SikMJMygrFC:b-bh132o)*

Issue #178, opened 2017-08-23
Deadlock in tests with race detector.


gavriva:
Hi guys!

when I run `go test -race` in badger's directory it hangs periodically.
[output.txt](https://github.com/dgraph-io/badger/files/1245573/output.txt)

srh:
Thank you for the report.  We'll get on this soon.

srh:
Working on this.  I've seen occasional hangs too.

srh:
This was reproduced and it looks like another problem with getting compaction to happen. To reproduce I had to add artificial load to my machine by making a few busy-waiting processes (or just run Slack in one browser tab).

srh:
I got it to trigger with some helpful printfs.

In addLevel0Table at the place where we sleep 10ms, we see

Level 0 totalSize: 48201, level 1 totalSize: 125040, level 1 maxTotalSize 131072

At the same time, pickCompactLevels returns an empty slice.  And there are zero active compactions, and in pickCompactLevels `s.cstatus.overlapsWith(0, infRange)` is observed to be false, but also `s.levels[0].numTables() >= s.kv.opt.NumLevelZeroTables` is false, which means we don't have a level 0 compaction.  Level 1 doesn't compact because its size is less than maxTotalSize.

The architectural problem here is that the logic in `addLevel0Table` needs to exactly replicate the logic in `pickCompactLevels` while it sleeps within a loop.  But it's not.

srh:
A fix is implemented in sam/178; I'll let it run a bit before submitting a PR.

manishrjain:
So, Level 0 is below the max size allowed (256MB), also level 0 tablets are less than NumLevelZeroTables. Similarly, level 1 is below its size as well, and that leads to no compactions running. So, what's the issue here?

srh:
addLevel0Table wouldn't progress unless level 0 was completely empty (`s.levels[0].getTotalSize() == 0`), under the assumption that sometime between level 0 being emptied and getting another table, that line of code would get to run.

manishrjain:
Nice catch!

PR #188, opened 2017-08-31
Fix valueLog/logFile locking

srh:
This fixes problems referred to by the comment at https://github.com/dgraph-io/badger/pull/177#issuecomment-324483634

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/188)

### DIFF
--- a/value.go
+++ b/value.go
@@ -65,8 +65,11 @@ const (
 )
 
 type logFile struct {
-	lock   sync.RWMutex
-	path   string
+	path string
+	// This is a lock on the file descriptor's value and the file's existence.  Use shared
+	// ownership when reading/writing the file, use exclusive ownership to open/close the
+	// descriptor or remove the file.
+	fdLock sync.RWMutex
 	fd     *os.File
 	fid    uint16
 	offset uint32
@@ -109,8 +112,8 @@ func (lf *logFile) doneWriting() error {
 	//
 	// If there's a benefit to reopening the file read-only, it might be on Windows.  I don't know
 	// what the benefit is.  Consider keeping the file read-write, or use fcntl to change permissions.
-	lf.lock.Lock()
-	defer lf.lock.Unlock()
+	lf.fdLock.Lock()
+	defer lf.fdLock.Unlock()
 	if err := lf.fd.Close(); err != nil {
 		return errors.Wrapf(err, "Unable to close value log: %q", lf.path)
 	}
@@ -337,10 +340,10 @@ func (vlog *valueLog) rewrite(f *logFile) error {
 	}
 
 	// Exclusively lock the file so that there are no readers before closing/destroying it
-	f.lock.Lock()
+	f.fdLock.Lock()
 	rem := vlog.fpath(f.fid)
 	f.fd.Close() // close file previous to remove it
-	f.lock.Unlock()
+	f.fdLock.Unlock()
 
 	elog.Printf("Removing %s", rem)
 	return os.Remove(rem)
@@ -636,13 +639,13 @@ func (vlog *valueLog) sync() error {
 		return nil
 	}
 	curlf := vlog.files[len(vlog.files)-1]
-	curlf.lock.RLock()
+	curlf.fdLock.RLock()
 	vlog.filesLock.RUnlock()
 
 	dirSyncCh := make(chan error)
 	go func() { dirSyncCh <- syncDir(vlog.opt.ValueDir) }()
 	err := curlf.sync()
-	curlf.lock.RUnlock()
+	curlf.fdLock.RUnlock()
 	dirSyncErr := <-dirSyncCh
 	if err != nil {
 		err = dirSyncErr
@@ -737,7 +740,7 @@ func (vlog *valueLog) getFileRLocked(fid uint16) (*logFile, error) {
 		return nil, ErrCorrupt
 	}
 	ret := vlog.files[idx]
-	ret.lock.RLock()
+	ret.fdLock.RLock()
 	return ret, nil
 }
 
@@ -747,7 +750,7 @@ func (vlog *valueLog) Read(p valuePointer, s *y.Slice) (e Entry, err error) {
 	if err != nil {
 		return e, err
 	}
-	defer lf.lock.RUnlock()
+	defer lf.fdLock.RUnlock()
 
 	if s == nil {
 		s = new(y.Slice)
--- a/value.go
+++ b/value.go
@@ -89,10 +89,8 @@ func (lf *logFile) openReadOnly() error {
 	return nil
 }
 
+// lf must be RLocked (or exclusively locked) if you call this.
 func (lf *logFile) read(buf []byte, offset int64) error {
-	lf.lock.RLock()
-	defer lf.lock.RUnlock()
-
 	nbr, err := lf.fd.ReadAt(buf, offset)
 	y.NumReads.Add(1)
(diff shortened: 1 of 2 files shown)


srh:
On a general note, I gave the lock fields actual names. It is possible to use the "go to all uses" feature of VS Code and maybe some other IDE's if the field has a name.

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[value.go, line 101 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/188#-KsuUcOOf14CKpXenZeJ:-KsuUcOPHhYIQj0APnIm:bdm9wf8) ([raw file](https://github.com/dgraph-io/badger/blob/a7b4b8b8da7d4fb15da98d148f5181f0562bbd35/value.go#L101)):*
> ```Go
> 
> func (lf *logFile) doneWriting() error {
> 	// Sync before acquiring lock.  (We call this from write() and thus know we have shared access
> ```

Independent of the other changes is our moving this lock acquisition down.

---

*[value.go, line 111 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/188#-KsuUgceO4rc1V_X60Ep:-KsuUgceO4rc1V_X60Eq:b-jbdupf) ([raw file](https://github.com/dgraph-io/badger/blob/a7b4b8b8da7d4fb15da98d148f5181f0562bbd35/value.go#L111)):*
> ```Go
> 	//
> 	// If there's a benefit to reopening the file read-only, it might be on Windows.  I don't know
> 	// what the benefit is.  Consider keeping the file read-write, or use fcntl to change permissions.
> ```

^^^ I was wondering about this.  Is there a reason we re-open the file, other than some safety of having it open read-only?

---

*[value.go, line 332 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/188#-KsuUkDA-bjvQJkiK5Uk:-KsuUkDA-bjvQJkiK5Ul:bzk578) ([raw file](https://github.com/dgraph-io/badger/blob/a7b4b8b8da7d4fb15da98d148f5181f0562bbd35/value.go#L332)):*
> ```Go
> 		})
> 		if idx == len(vlog.files) || vlog.files[idx].fid != f.fid {
> 			vlog.filesLock.Unlock()
> ```

Independent of the other changes is this fix to 

---

*[value.go, line 639 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/188#-KsuVTKcrWdBDDlsEorW:-KsuVTKcrWdBDDlsEorX:brya07u) ([raw file](https://github.com/dgraph-io/badger/blob/a7b4b8b8da7d4fb15da98d148f5181f0562bbd35/value.go#L639)):*
> ```Go
> 	}
> 	curlf := vlog.files[len(vlog.files)-1]
> 	curlf.lock.RLock()
> ```

The other main change is that readers must get shared access to the logFile before releasing shared access to the valueLog.  This is as was discussed in Slack.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/188#-:-KsuV0kfvcToO4wF4U8G:bwb8fc5)*

manishrjain:
Adding lock fields seems an unnecessary change. A ctrl+f could have given you that? Because all the users are in the same file. Other than that, :lgtm_strong: 

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[value.go, line 111 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/188#-KsuUgceO4rc1V_X60Ep:-KsuiIO4K7mQp9cc6asx:b-ilnu84) ([raw file](https://github.com/dgraph-io/badger/blob/a7b4b8b8da7d4fb15da98d148f5181f0562bbd35/value.go#L111)):*

^^^ I was wondering about this.  Is there a reason we re-open the file, other than some safety of having it open read-only?
</blockquote></details>

It's cleaner to make a file supposed to be immutable read only. Blocking a few reads / writes for this small duration, is a small price to pay for this.

---

*[value.go, line 332 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/188#-KsuUkDA-bjvQJkiK5Uk:-KsulILMtixqsJLUit1u:b32w0x) ([raw file](https://github.com/dgraph-io/badger/blob/a7b4b8b8da7d4fb15da98d148f5181f0562bbd35/value.go#L332)):*

Independent of the other changes is this fix to 
</blockquote></details>

OK

---

*[value.go, line 639 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/188#-KsuVTKcrWdBDDlsEorW:-Ksulqe6I-F1XJ0WrgMS:b2nhktp) ([raw file](https://github.com/dgraph-io/badger/blob/a7b4b8b8da7d4fb15da98d148f5181f0562bbd35/value.go#L639)):*

The other main change is that readers must get shared access to the logFile before releasing shared access to the valueLog.  This is as was discussed in Slack.
</blockquote></details>

OK.

---

*[value.go, line 733 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/188#-Ksum9puo6N6v1IelugG:-Ksum9puo6N6v1IelugH:b-wspfj) ([raw file](https://github.com/dgraph-io/badger/blob/a7b4b8b8da7d4fb15da98d148f5181f0562bbd35/value.go#L733)):*
> ```Go
> 	defer vlog.filesLock.RUnlock()
> 
> 	idx := sort.Search(len(vlog.files), func(idx int) bool {
> ```

Unrelated to this change, but caught my eye. This seems expensive to do for every read call. We should switch to a map instead of a list of files.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/188#-:-Ksuj0XATRCMoFG1FvEQ:bc5wfs)*

srh:
How do you know all the users are in the same file?  The first iteration of this review was broken because I had missed stuff by trying to find all uses of the locks manually somehow.  Also the lock for the value log isn't locking the entire object, and for logFile it's basically a lock on the file descriptor, not the whole structure either.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/188#-:-KsuoUJ8WsPofw4TvP_4:b-cvqrsl)*

srh:

---

*[value.go, line 733 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/188#-Ksum9puo6N6v1IelugG:-KsuplrwdyiL3pd3noaP:b2s2c8n) ([raw file](https://github.com/dgraph-io/badger/blob/a7b4b8b8da7d4fb15da98d148f5181f0562bbd35/value.go#L733)):*

Unrelated to this change, but caught my eye. This seems expensive to do for every read call. We should switch to a map instead of a list of files.
</blockquote></details>

Made issue https://github.com/dgraph-io/badger/issues/189 for this. Working on it now.  (Generally speaking I'm doing cleanup work while working on https://github.com/dgraph-io/badger/issues/178.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/188)*

manishrjain:
> Also the lock for the value log isn't locking the entire object, and for logFile it's basically a lock on the file descriptor, not the whole structure either.

Fair point.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/188#-:-KsuylG84bLdzwXg0fFl:b-bp8dx9)*

srh:
In master as of edbce1ba8ec4e07ff2d881291d314598512a6a0c.

Issue #189, opened 2017-09-01
Make valueLog::files field be a map


srh:
Right now it's a sorted array and the result is we binary search through it, using sort.Search.  A map would probably be better.  The only time we take advantage of `files` being an array is, I think, when we select a file at random for the purpose of GC.

PR #192, opened 2017-09-01
Replace valueLog::files with valueLog::filesMap

srh:
Connected to #189.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/192)

### DIFF
--- a/value.go
+++ b/value.go
@@ -588,14 +588,14 @@ func (vlog *valueLog) Close() error {
 	return err
 }
 
-// sortedFiles returns the filesMap sorted by fid.  Assumes we have shared access to filesMap
-func (vlog *valueLog) sortedFiles() []*logFile {
-	ret := make([]*logFile, 0, len(vlog.filesMap))
-	for _, f := range vlog.filesMap {
-		ret = append(ret, f)
+// sortedFids returns the file id's, sorted.  Assumes we have shared access to filesMap.
+func (vlog *valueLog) sortedFids() []uint16 {
+	ret := make([]uint16, 0, len(vlog.filesMap))
+	for fid := range vlog.filesMap {
+		ret = append(ret, fid)
 	}
 	sort.Slice(ret, func(i, j int) bool {
-		return ret[i].fid < ret[j].fid
+		return ret[i] < ret[j]
 	})
 	return ret
 }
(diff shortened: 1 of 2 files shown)


manishrjain:
:lgtm: Another tangential comment, unrelated to this change.

---

Reviewed 2 of 2 files at r1.

---

*[value.go, line 478 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/192#-Kt-Xs4mfKqvU5ASwVpy:-Kt-Xs4mfKqvU5ASwVpz:bx3m0zn) ([raw file](https://github.com/dgraph-io/badger/blob/587559f370d9390eabfa21165b631ff90a09d3b6/value.go#L478)):*
> ```Go
> 	// guards our view of which files exist
> 	filesLock sync.RWMutex
> 	filesMap  map[uint16]*logFile
> ```

Some time ago, we debated whether we should use a 16-bit int or a 32-bit int for logfiles. I think the 16-bit int is still a contentious decision in my mind. We should switch it over to 32-bit to be on the safe side. Because our initial value log size depends upon the number of writes, not the total amount of data.

An extra 2 bytes is probably OK, with the assurance that we won't run out of numbers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/192#-:-Kt-YZu64UaBsRLZ0sfG:b-hwjhjv)*

srh:

---

*[value.go, line 478 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/192#-Kt-Xs4mfKqvU5ASwVpy:-Kt8m2EtDNixd5HrlRPA:b9s6irk) ([raw file](https://github.com/dgraph-io/badger/blob/587559f370d9390eabfa21165b631ff90a09d3b6/value.go#L478)):*

Some time ago, we debated whether we should use a 16-bit int or a 32-bit int for logfiles. I think the 16-bit int is still a contentious decision in my mind. We should switch it over to 32-bit to be on the safe side. Because our initial value log size depends upon the number of writes, not the total amount of data.

An extra 2 bytes is probably OK, with the assurance that we won't run out of numbers.
</blockquote></details>

I completely agree, because 65 TB of writes is not much. That was going to be next :)  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/192)*

srh:
In master as of 2a6fc673dc60d17d7bcb75c95f5d4b661658b7a8.

PR #196, opened 2017-09-05
Update format string for file path to use 6 chars

deepakjois:
I ran into an issue with this change when I tried to run benchmarks against an older dataset. The filenames did not correspond, and Badger crashed with an error.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/196)

### DIFF
--- a/value.go
+++ b/value.go
@@ -490,7 +490,7 @@ type valueLog struct {
 }
 
 func vlogFilePath(dirPath string, fid uint32) string {
-	return fmt.Sprintf("%s%s%09d.vlog", dirPath, string(os.PathSeparator), fid)
+	return fmt.Sprintf("%s%s%06d.vlog", dirPath, string(os.PathSeparator), fid)
 }
 
 func (vlog *valueLog) fpath(fid uint32) string {


srh:
This change should be fine if we're happy having filenames lose their sorting order after creating 1000000 vlog files.

But your older pre-existing Badger directory still won't work properly because file id values into the value log were changed to be 4 bytes, which changed their serialization format too.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/196#-:-KtIi24HDshCYx9I8CPR:b-bihhm2)*

srh:
So, :lgtm:, though I have no opinion about the decision on whether this change is positive or negative -- the positive is that most users get shorter, more convenient file names.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/196#-:-KtIifvyNtYyogFpOueW:b-qh4wbs)*

srh:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/196)*

deepakjois:
If the older pre-existing Badger directory won’t work properly anyway (which I also observed after hacking in this change in my local copy), I am not that strongly inclined to go ahead with this. There seems to be more value in preserving sorting order.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/196#-:-KtJq3_R0WdLNMlZU-Un:btho4gd)*

manishrjain:
:lgtm: I think 6 digits is good enough for filenames.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/196#-:-KtJyuJgHZp1RrRRxFI5:blnnf1c)*

PR #204, opened 2017-09-07
Minor fixes to mmap reads

deepakjois:
* update prefetch status, even if val is nil

* avoid copying if backing array is the same,
  which can happen if the read happened into
  the slice instead of the mmap reference
  being returned.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/204)


deepakjois:
@manishrjain I am going to merge this because it is really minor. But feel free to provide any comments if you get a change to look.

PR #213, opened 2017-09-11
Removing ineffectal assignment.

deepakjois:
Go Report Card complains about this code, and there does not seem to be
a good reason to execute the loops and increment a value that is not
being used anywhere else.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/213)


deepakjois:
@srh I am going to go ahead an merge this code, because it is not changing anything semantically as far as I can see. Just want to speed the process of getting Badger into go-awesome.

But do point out if you think I missed something here.

srh:
Yep, clearly a good change.

PR #216, opened 2017-09-12
Fix race when exposing writeCh metrics

peterstace:
Just starts the goroutine after the channel is created, rather than before.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/216)


peterstace:
Not needed, using an alternate approach which avoids the problem.

PR #219, opened 2017-09-12
Avoid copying values during pre-fetching.

deepakjois:
Instead of copying the values, we just discard the mmap-ed slice
returned. This should make sure the data has been loaded from disk to
RAM, and is ready when we actually need the value later in the
iteration.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/219)

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -18,6 +18,7 @@ package badger
 
 import (
 	"bytes"
+	"io/ioutil"
 	"sync"
 
 	"github.com/dgraph-io/badger/y"
@@ -25,25 +26,17 @@ import (
 
 type prefetchStatus uint8
 
-const (
-	empty prefetchStatus = iota
-	prefetched
-)
-
 // KVItem is returned during iteration. Both the Key() and Value() output is only valid until
 // iterator.Next() is called.
 type KVItem struct {
 	status     prefetchStatus
-	err        error
 	wg         sync.WaitGroup
 	kv         *KV
 	key        []byte
 	vptr       []byte
 	meta       byte
 	userMeta   byte
-	val        []byte
 	casCounter uint64
-	slice      *y.Slice
 	next       *KVItem
 }
 
@@ -62,12 +55,6 @@ func (item *KVItem) Key() []byte {
 // append to this slice; it would result in a panic.
 func (item *KVItem) Value(consumer func([]byte) error) error {
 	item.wg.Wait()
-	if item.status == prefetched {
-		if item.err != nil {
-			return item.err
-		}
-		return consumer(item.val)
-	}
 	return item.kv.yieldItemValue(item, consumer)
 }
 
@@ -83,18 +70,11 @@ func (item *KVItem) hasValue() bool {
 	return true
 }
 
-func (item *KVItem) prefetchValue() {
-	item.err = item.kv.yieldItemValue(item, func(val []byte) error {
-		if val == nil {
-			item.status = prefetched
-			return nil
-		}
-
-		buf := item.slice.Resize(len(val))
-		copy(buf, val)
-		item.val = buf
-		item.status = prefetched
-		return nil
+func (item *KVItem) primeValueMmap() {
+	// We just prime the mmap to load data from file to RAM
+	_ = item.kv.yieldItemValue(item, func(val []byte) error {
+		_, err := ioutil.Discard.Write(val)
+		return err
 	})
 }
 
@@ -187,7 +167,7 @@ type Iterator struct {
 func (it *Iterator) newItem() *KVItem {
 	item := it.waste.pop()
 	if item == nil {
-		item = &KVItem{slice: new(y.Slice), kv: it.kv}
+		item = &KVItem{kv: it.kv}
 	}
 	return item
 }
@@ -245,12 +225,10 @@ func (it *Iterator) fill(item *KVItem) {
 	item.casCounter = vs.CASCounter
 	item.key = y.Safecopy(item.key, it.iitr.Key())
 	item.vptr = y.Safecopy(item.vptr, vs.Value)
-	item.val = nil
 	if it.opt.PrefetchValues {
 		item.wg.Add(1)
 		go func() {
-			// FIXME we are not handling errors here.
-			item.prefetchValue()
+			item.primeValueMmap()
 			item.wg.Done()
 		}()
 	}
--- a/kv.go
+++ b/kv.go
@@ -404,23 +404,13 @@ func (s *KV) yieldItemValue(item *KVItem, consumer func([]byte) error) error {
 		return consumer(nil)
 	}
 
-	if item.slice == nil {
-		item.slice = new(y.Slice)
-	}
-
 	if (item.meta & BitValuePointer) == 0 {
-		val := item.slice.Resize(len(item.vptr))
-		copy(val, item.vptr)
-		return consumer(val)
+		return consumer(item.vptr)
 	}
 
 	var vp valuePointer
 	vp.Decode(item.vptr)
-	err := s.vlog.Read(vp, consumer)
-	if err != nil {
-		return err
-	}
-	return nil
+	return s.vlog.Read(vp, consumer)
 }
 
 // get returns the value in memtable or disk for given key.


manishrjain:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[iterator.go, line 59 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/219#-KtowDw6AJ4HPTaA7ecp:-KtowDw7u02ulYru_Y1-:b-g8sp1q) ([raw file](https://github.com/dgraph-io/badger/blob/efa2c3ff6cfe585fc136a63c448d7b3352f49f25/iterator.go#L59)):*
> ```Go
> // append to this slice; it would result in a panic.
> func (item *KVItem) Value(consumer func([]byte) error) error {
> 	item.wg.Wait()
> ```

item.wg.Wait() is probably also not required. Can remove it from `fill`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/219)*

manishrjain:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[iterator.go, line 75 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/219#-Ktowh_MUKQoXM66Y32G:-Ktowh_MUKQoXM66Y32H:b3jt2ei) ([raw file](https://github.com/dgraph-io/badger/blob/efa2c3ff6cfe585fc136a63c448d7b3352f49f25/iterator.go#L75)):*
> ```Go
> }
> 
> func (item *KVItem) prefetchValue() {
> ```

Might rename this to `valuePrimer()`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/219)*

deepakjois:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[iterator.go, line 59 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/219#-KtowDw6AJ4HPTaA7ecp:-Ktp1iymV2eRL8YEqBXY:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/efa2c3ff6cfe585fc136a63c448d7b3352f49f25/iterator.go#L59)):*

item.wg.Wait() is probably also not required. Can remove it from `fill`.
</blockquote></details>

Done.

---

*[iterator.go, line 75 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/219#-Ktowh_MUKQoXM66Y32G:-Ktp1jr1RqoBwH2Uhgdk:brvsm5s) ([raw file](https://github.com/dgraph-io/badger/blob/efa2c3ff6cfe585fc136a63c448d7b3352f49f25/iterator.go#L75)):*

Might rename this to `valuePrimer()`.
</blockquote></details>

renamed it to primeValueMmap

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/219)*

manishrjain:
So, the word on this is that it's slow?

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/219#-:-KtsYrmhRclgZk8s4lZZ:b-lufrxx)*

srh:
If we avoid copying values and return mmap'd slices, then we can never do value log GC, period.  We have to either copy values, or make the user tell us when they're done using slices, or never do value log GC.

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/219#-:-KttPIaUjNY128c9cPlB:b1ae2p7)*

manishrjain:
@srh The idea wasn't to return mmaped slice, but just to access that mmaped portion that we know we'll need soon. That way, unix can bring the needed pages into memory, before we need them.

srh:
Oh okay.  Now I get it.  Saw some copying being removed and freaked out.

Performance problems might be because the code wasn't actually causing a preload to happen -- see my comment about Discard below.

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[iterator.go, line 76 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/219#-Ku2brLisPnbGD_zIbcm:-Ku2brLisPnbGD_zIbcn:b-mzlthr) ([raw file](https://github.com/dgraph-io/badger/blob/50cc3938ceebccde0ec450e9333c5a3dd3510e9b/iterator.go#L76)):*
> ```Go
> 	// We just prime the mmap to load data from file to RAM
> 	_ = item.kv.yieldItemValue(item, func(val []byte) error {
> 		_, err := ioutil.Discard.Write(val)
> ```

The goal here is to actually force the slice, which is mmaped memory, to be loaded from disk, right?  And Discard.Write is supposed to accomplish this?

It won't -- Discard.Write doesn't actually touch the bytes in the slice `val`.  This statement has no effect.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/219#-:-Ku2cKy19CBmH5I_kAwm:b913dto)*

manishrjain:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[iterator.go, line 76 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/219#-Ku2brLisPnbGD_zIbcm:-Ku2d6Uck2Aa6R389g-4:ba4fc2w) ([raw file](https://github.com/dgraph-io/badger/blob/50cc3938ceebccde0ec450e9333c5a3dd3510e9b/iterator.go#L76)):*

The goal here is to actually force the slice, which is mmaped memory, to be loaded from disk, right?  And Discard.Write is supposed to accomplish this?

It won't -- Discard.Write doesn't actually touch the bytes in the slice `val`.  This statement has no effect.
</blockquote></details>

True, but we do read the initial part of the total slice to read the header, then pass the rest to this function. So, only if value is large would the mmap not be touched. I suppose if we do this:

var b byte
for i := range val {
  b = val[i] 
}

That might ensure that it would touch the slice in all cases, and that the compiler doesn't optimize the code away.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/219)*

srh:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[iterator.go, line 76 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/219#-Ku2brLisPnbGD_zIbcm:-Ku2eVXQAjU2BWWYpS6T:b4f4k05) ([raw file](https://github.com/dgraph-io/badger/blob/50cc3938ceebccde0ec450e9333c5a3dd3510e9b/iterator.go#L76)):*

True, but we do read the initial part of the total slice to read the header, then pass the rest to this function. So, only if value is large would the mmap not be touched. I suppose if we do this:

var b byte
for i := range val {
  b = val[i] 
}

That might ensure that it would touch the slice in all cases, and that the compiler doesn't optimize the code away.
</blockquote></details>

The compiler would likely optimize that code away in latest Go or later, if they're using SSA now and doing basic optimizations.

A downside to that is since we've already madvised the whole memory map that access is random, it won't, I believe, read-ahead any pages, which means the behavior might be to fetch the pages of the value one page at a time.  It might be proper behavior for vlog::Read to call madvise on the pages holding the value, with MADV_WILLNEED.  I don't know how "sticky" that advice is.

(I'd make a wild guess we should use MADV_NORMAL for value log files instead of MADV_RANDOM because the kernel will tune the read-ahead window based on past hit rate.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/219)*

deepakjois:
Should be just close this?

manishrjain:
Yeah. Copying has a real value for us now. In a long-running transaction, doing value copies releases the read locks on the log files quicker, allowing things to run smoothly. So, we're not taking the copying out at all now.

PR #220, opened 2017-09-12
Explicit check and/or handle errors whereever possible.

### LINKED ISSUE #218: Audit code for any unhandled errors (found by errcheck)
`errcheck` returns a few lines of code where we are not handing errors. We should audit this, and make sure we are not missing out on handling any errors that ought to be handled. Even in the case where we are ignoring errors, it might be worthwhile to make it explicit.

```
$ errcheck  | grep -v "_test"
dir_unix.go:55:10:      f.Close()
dir_unix.go:65:10:      f.Close()
iterator.go:210:15:     it.iitr.Close()
kv.go:173:22:   go out.flushMemtable(out.closers.memtable) // Need levels controller to be up.
kv.go:1031:18:  defer iter.Close()
kv.go:1099:14:  tbl.DecrRef()                  // Releases our ref.
level_handler.go:249:17:        defer it.Close()
levels.go:279:16:       defer it.Close() // Important to close the iterator to do ref counting.
levels.go:350:25:       newTables[j].DecrRef()
manifest.go:234:11:     fp.Close()
manifest.go:243:11:     fp.Close()
manifest.go:247:11:     fp.Close()
manifest.go:264:11:     fp.Close()
manifest.go:268:11:     fp.Close()
util.go:54:40:  func (s *KV) validate() { s.lc.validate() }
value.go:384:10:        y.Munmap(f.fmap)
value.go:385:12:        f.fd.Close() // close file previous to remove it
value.go:424:12:        hash.Write(headerEnc[:])
value.go:427:12:        hash.Write(e.Key)
value.go:430:12:        hash.Write(e.Value)
value.go:887:16:        vlog.doRunGC()
```

deepakjois:
As a convention, we should try and explicitly check for errors. If we
don’t wish to handle an error, we can just ignore it using the ‘_’
character (we should be prudent with ignoring the error, and only do it
where it is safe).

This change fixes a bunch of code flagged by the errcheck utility. We
ignore some cases, for e.g. when a function is called using the defer
keyword.

Fixes #218.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/220)

### DIFF
--- a/dir_unix.go
+++ b/dir_unix.go
@@ -52,7 +52,7 @@ func AcquireDirectoryLock(dirPath string, pidFileName string) (*DirectoryLockGua
 	}
 	err = unix.Flock(int(f.Fd()), unix.LOCK_EX|unix.LOCK_NB)
 	if err != nil {
-		f.Close()
+		_ = f.Close()
 		return nil, errors.Wrapf(err,
 			"Cannot acquire directory lock on %q.  Another process is using this Badger database.",
 			dirPath)
@@ -62,7 +62,7 @@ func AcquireDirectoryLock(dirPath string, pidFileName string) (*DirectoryLockGua
 	// directory.
 	err = ioutil.WriteFile(absPidFilePath, []byte(fmt.Sprintf("%d\n", os.Getpid())), 0666)
 	if err != nil {
-		f.Close()
+		_ = f.Close()
 		return nil, errors.Wrapf(err,
 			"Cannot write pid file %q", absPidFilePath)
 	}
--- a/iterator.go
+++ b/iterator.go
@@ -207,7 +207,7 @@ func (it *Iterator) ValidForPrefix(prefix []byte) bool {
 
 // Close would close the iterator. It is important to call this when you're done with iteration.
 func (it *Iterator) Close() {
-	it.iitr.Close()
+	_ = it.iitr.Close()
 }
 
 // Next would advance the iterator by one. Always check it.Valid() after a Next()
--- a/kv.go
+++ b/kv.go
@@ -170,6 +170,7 @@ func NewKV(optParam *Options) (out *KV, err error) {
 	out.lc.startCompact(out.closers.compactors)
 
 	out.closers.memtable = y.NewCloser(1)
+	// FIXME why are we dropping an error here?
 	go out.flushMemtable(out.closers.memtable) // Need levels controller to be up.
 
 	if err = out.vlog.Open(out, &opt); err != nil {
@@ -1103,7 +1104,11 @@ func (s *KV) flushMemtable(lc *y.Closer) error {
 		}
 		// We own a ref on tbl.
 		err = s.lc.addLevel0Table(tbl) // This will incrRef (if we don't error, sure)
-		tbl.DecrRef()                  // Releases our ref.
+		if err != nil {
+			return err
+		}
+
+		err = tbl.DecrRef() // Releases our ref.
 		if err != nil {
 			return err
 		}
--- a/levels.go
+++ b/levels.go
@@ -347,7 +347,7 @@ func (s *levelsController) compactBuildTables(
 		// -- we're the only holders of a ref).
 		for j := 0; j < i; j++ {
 			if newTables[j] != nil {
-				newTables[j].DecrRef()
+				_ = newTables[j].DecrRef()
 			}
 		}
 		errorReturn := errors.Wrapf(firstErr, "While running compaction for: %+v", cd)
--- a/manifest.go
+++ b/manifest.go
@@ -231,7 +231,7 @@ func helpRewrite(dir string, m *Manifest) (*os.File, int, error) {
 
 	changeBuf, err := set.Marshal()
 	if err != nil {
-		fp.Close()
+		_ = fp.Close()
 		return nil, 0, err
 	}
 	var lenCrcBuf [8]byte
@@ -240,11 +240,11 @@ func helpRewrite(dir string, m *Manifest) (*os.File, int, error) {
 	buf = append(buf, lenCrcBuf[:]...)
 	buf = append(buf, changeBuf...)
 	if _, err := fp.Write(buf); err != nil {
-		fp.Close()
+		_ = fp.Close()
 		return nil, 0, err
 	}
 	if err := fp.Sync(); err != nil {
-		fp.Close()
+		_ = fp.Close()
 		return nil, 0, err
 	}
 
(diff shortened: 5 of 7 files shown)


deepakjois:
I tried to use my judgement in most places, but in some places I could not figure out what the best behavior could be, so I left a FIXME.

manishrjain:
Few obvious issues, resolve them before a full review.

---

Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[dir_unix.go, line 55 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/220#-Kts_wi7A84LBjKIQgj2:-Kts_wi7A84LBjKIQgj3:b13p2pd) ([raw file](https://github.com/dgraph-io/badger/blob/8371b486246bc7383056dd6792ca86d1cf76a160/dir_unix.go#L55)):*
> ```Go
> 	err = unix.Flock(int(f.Fd()), unix.LOCK_EX|unix.LOCK_NB)
> 	if err != nil {
> 		_ = f.Close()
> ```

I had this idea that I had mentioned on Slack that we should use an Ignore function. That way, we could use this function to at least log these errors. That'd be better than using underscores to ignore errors.

---

*[kv.go, line 173 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/220#-KtsaDqghyE83x7Wygxg:-KtsaDqghyE83x7Wygxh:bk7q171) ([raw file](https://github.com/dgraph-io/badger/blob/8371b486246bc7383056dd6792ca86d1cf76a160/kv.go#L173)):*
> ```Go
> 
> 	out.closers.memtable = y.NewCloser(1)
> 	// FIXME why are we dropping an error here?
> ```

Use `// TODO` instead of `FIXME` -- not that I have a preference, but we've been using TODOs all along.

---

*[kv.go, line 1107 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/220#-Kts_kecC7IGwi1yAoiQ:-Kts_kecC7IGwi1yAoiR:bt6ydcu) ([raw file](https://github.com/dgraph-io/badger/blob/8371b486246bc7383056dd6792ca86d1cf76a160/kv.go#L1107)):*
> ```Go
> 		// We own a ref on tbl.
> 		err = s.lc.addLevel0Table(tbl) // This will incrRef (if we don't error, sure)
> 		if err != nil {
> ```

Don't change the behavior of the code.

Here even if adding the table to level 0 caused an error, we still want to call decrref. Your change doesn't call decrref on error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/220#-:-KtsaLtVTjFerUw6pkNp:bfbtell)*

srh:
Review status: 0 of 7 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[iterator.go, line 209 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/220#-Kts_t3DwBKIoEjR-4q5:-Kts_t3EkQIPlQub9_8Q:blm199g) ([raw file](https://github.com/dgraph-io/badger/blob/8371b486246bc7383056dd6792ca86d1cf76a160/iterator.go#L209)):*
> ```Go
> 
> // Close would close the iterator. It is important to call this when you're done with iteration.
> func (it *Iterator) Close() {
> ```

It looks reasonable to return the error (and change the signature of this function).

---

*[kv.go, line 173 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/220#-Ktsa2icQkVvEWoiTW9e:-Ktsa2icQkVvEWoiTW9f:b-klxl49) ([raw file](https://github.com/dgraph-io/badger/blob/8371b486246bc7383056dd6792ca86d1cf76a160/kv.go#L173)):*
> ```Go
> 
> 	out.closers.memtable = y.NewCloser(1)
> 	// FIXME why are we dropping an error here?
> ```

There's a FIXME here.

---

*[kv.go, line 1107 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/220#-Ktsa8VVW5pDJrVpKBWc:-Ktsa8VVW5pDJrVpKBWd:b-mbyj18) ([raw file](https://github.com/dgraph-io/badger/blob/8371b486246bc7383056dd6792ca86d1cf76a160/kv.go#L1107)):*
> ```Go
> 		// We own a ref on tbl.
> 		err = s.lc.addLevel0Table(tbl) // This will incrRef (if we don't error, sure)
> 		if err != nil {
> ```

Here we deliberately called DecrRef before handling the error.

---

*[util.go, line 55 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/220#-KtsaU_hdhBRPJfEkXtB:-KtsaU_hdhBRPJfEkXtC:b-mwu6lb) ([raw file](https://github.com/dgraph-io/badger/blob/8371b486246bc7383056dd6792ca86d1cf76a160/util.go#L55)):*
> ```Go
> 
> func (s *KV) validate() {
> 	// FIXME we do not check for error here.
> ```

Looks like we'd return it and handle it in the caller.

---

*[value.go, line 887 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/220#-KtsajjnGQusSLuvvrEf:-KtsajjnGQusSLuvvrEg:b-9m7fjl) ([raw file](https://github.com/dgraph-io/badger/blob/8371b486246bc7383056dd6792ca86d1cf76a160/value.go#L887)):*
> ```Go
> 			return
> 		case <-tick.C:
> 			// FIXME we do not check for here.
> ```

The tricky part is returning the error back up, or somehow aborting the rest of the DB.  Which we should generally want to do.  We have a similar ignored error in compaction.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/220)*

srh:
Review status: 0 of 7 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[value.go, line 385 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/220#-KtsbVIhVZR7wPhoKc3X:-KtsbVIhVZR7wPhoKc3Y:bcbs1bh) ([raw file](https://github.com/dgraph-io/badger/blob/8371b486246bc7383056dd6792ca86d1cf76a160/value.go#L385)):*
> ```Go
> 	rem := vlog.fpath(f.fid)
> 	_ = y.Munmap(f.fmap)
> 	_ = f.fd.Close() // close file previous to remove it
> ```

I think these errors should not be ignored.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/220)*

deepakjois:
This is very out of date. Will submit another PR after doing another `errcheck` pass on repo.

PR #231, opened 2017-09-18
Allocate nodes from the arena.

cardbot:
The skiplist allocates keys and values from the arena, but not nodes.
This change begins allocating nodes from the arena. Links between
nodes at the same level are now offsets into the arena rather than
pointers. The "tower" of links is allocated as extra bytes appended
to the end of the node rather than a separate array.

Together, these changes significantly speed up skiplist writes and
reads. It also reduces overall memory usage (by using next offsets
rather than next pointers), as well as GC pressure (nodes do not
use pointers to reference one another).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/231)

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -93,6 +93,7 @@ func NewKV(optParam *Options) (out *KV, err error) {
 	// Make a copy early and fill in maxBatchSize
 	opt := *optParam
 	opt.maxBatchSize = (15 * opt.MaxTableSize) / 100
+	opt.maxBatchCount = opt.maxBatchSize / int64(skl.MaxNodeSize)
 
 	for _, path := range []string{opt.Dir, opt.ValueDir} {
 		dirExists, err := exists(path)
@@ -697,7 +698,7 @@ func (s *KV) doWrites(lc *y.Closer) {
 
 func (s *KV) sendToWriteCh(entries []*Entry) []*request {
 	var reqs []*request
-	var size int64
+	var count, size int64
 	var b *request
 	var bad []*Entry
 	for _, entry := range entries {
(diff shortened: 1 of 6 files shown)


manishrjain:
Very interesting changes. Particularly impressive benchmarks.

I wonder if there are alternate ways to achieve this skiplist, which doesn't use a node struct, and probably avoids the whole unsafe.Pointer thing.

It's worth asking: If you were to redesign this skiplist, how would you do it such that we can achieve the performance numbers that you have achieved, while also simplifying the code design.

---

Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[skl/skl.go, line 77 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/231#-KuHtc2hp8FUmAOTfzpS:-KuHtc2hp8FUmAOTfzpT:b-14anvo) ([raw file](https://github.com/dgraph-io/badger/blob/8cceeed6bb55aae82eae39b1c9d64e0a4109de19/skl/skl.go#L77)):*
> ```Go
> 	// exponentially. The size is always <= maxHeight. All accesses to elements
> 	// should use CAS operations, with no need to lock.
> 	tower [1]uint32
> ```

If this tower is an array of size 1, then shouldn't it just be `tower uint32`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/231#-:-KuHu0yreExlI94whdtt:b-2pqqw6)*

manishrjain:
Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/231)*

manishrjain:
:lgtm: Second approach sounds good. Let me know if you're making any more changes. Otherwise, this is good to go from my end.

---

Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/231#-:-KuI9sjhzLp6skmwr7oi:b8a33mi)*

manishrjain:
1. Not a problem.

2. Again, not a problem. The memory usage is the same as before. Instead of allocating many objects on heap, we're putting them in one place.

Let me know you think the PR is good. I'll do a quick final review and merge it.

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/231#-:-KuMc_i3T6gPDpXsv7qE:bqkfjmt)*

manishrjain:
:lgtm: Thanks for this change!

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/231#-:-KuMjehqb-z1lvFP4FSk:b-hjzf6)*

Issue #135, opened 2017-07-31
Have a way to create backup of Badger


manishrjain:
Look into how RocksDB is doing things. Use that knowledge to inform our case of how we should be doing backups in Badger. We don't do versioning, but we do write ahead logs, which can be used to ensure that our backup is consistent up to a certain point.

Some rough notes:

• Unsorted iteration in Badger
   ◦ Go over the value log first, then check the keys in LSM.
   ◦ If sync=true, then this should work.
   ◦ If sync=false, then some key-values won’t be in vlog.
   ◦ Then iterate over LSM tree as well.

These say unsorted iteration because sorted iteration might be too slow (verify that using GOMAXPROCS=128, or 256).

We could write this out to just one file. Or, we could write this out to a bunch of files, which might be consistent with how Badger stores things (LSM + value log). If we do the latter, we could potentially open up one Badger store and just write to it -- but then, compactions would slow us down.

OTOH, we should allow a way to do incremental backups as well. How would that work?

srh:
For incremental:

- Our global incrementing CAS counter is generally useful for filtering values / tombstones that were already backed up.
- Tombstone entries need to be kept around at the highest numbered level.  I don't know off the top of my head whether we do that.  (Or: Tombstone entries with CAS counter greater than some threshold -- that of the last successful backup, as informed by the user -- need to be kept around.)
- It will be useful to retain some general knowledge of the max CAS value seen in any given .sst file, so we can skip those.

manishrjain:
> Tombstone entries need to be kept around at the highest numbered level. I don't know off the top of my head whether we do that.

For sync=true, they're written to value log. So, they'd be around. In fact, the whole backup thing works nicely if we can count on value log.

manishrjain:
Had some ideas about making successive backups fast if there are little data changes.

• For the first run, divide up Badger data into key ranges of equal size. Have one file per key range.
• Second run onwards, determine the key ranges from the existing files.
• When Backup starts, divide up the task into these key ranges, one goroutine per KR.
• Before outputting anything, create a checksum for this key range, by doing a fast key-only iteration.
• Compare this against a checksum stored in the KR file.
• If no change, then we’re done for the KR.
• Also, run an estimate for how big the KR is, using EstimateSize while key-only itearation.
• If it exceeds the size, then do a binary split of the KR.
• Write the KR.
• This design ensures that we don’t rewrite every key range every time. We only write data if this key-range had changes since the last backup.
[...]

PR #233, opened 2017-09-18
Badger backup support

srh:
This adds Badger backup support.

What this means is:
1. An API for streaming badger backup items from Badger
2. An API for storing backup items in a backup store.

Connected to #135.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/233)


srh:
Review status: 0 of 17 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[backupstore.go, line 34 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuIKpJb3JxcOJFyT6OC:-KuIKpJb3JxcOJFyT6OD:b-s99ut7) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/backupstore.go#L34)):*
> ```Go
> 	"github.com/pkg/errors"
> )
> ```

There should be some comment explaining the layout of a badger backup store here.

It consists of a directory with a BACKUPMANIFEST file (which we rewrite every time) and a bunch of files "backup-1", "backup-2", etc.  Every new incremental backup adds a new one of these, and rewrites the BACKUPMANIFEST file with all the info about backup files.

---

*[backupstore.go, line 244 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuILOwyNAwNRNKEfjbz:-KuILOwzhpqo7xBB2kXQ:b5gom5s) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/backupstore.go#L244)):*
> ```Go
> // it produces an empty slice (or error).  Keys must be distinct and arrive in increasing order
> // (across all slices).
> func NewBackup(path string, maxCASCounter uint64, producer func() ([]protos.BackupItem, error)) (err error) {
> ```

"maxCASCounter" shouldn't be exposed in the user API as if it were a "CAS counter" value.  It should be called a backup identifier, something like that.

---

*[kv.go, line 390 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuILmDELGT90yY2gdXX:-KuILmDELGT90yY2gdXY:bolkfod) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/kv.go#L390)):*
> ```Go
> 	tables := make([]*skl.Skiplist, 0, n+1)
> 	// We make a copy of the mutable skiplist.
> ```

I really ought to make sure that the write loop doesn't call `s.Lock()` while holding writePendingCh.

---

*[kv_test.go, line 60 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuIM9qGlV-MeZANAQOE:-KuIM9qGlV-MeZANAQOF:bnc03pm) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/kv_test.go#L60)):*
> ```Go
> }
> 
> type tempKV struct {
> ```

An independent change.

---

*[manifest.go, line 71 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuIMObQ2BkgOumiVw-b:-KuIMObQ2BkgOumiVw-c:b108svy) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/manifest.go#L71)):*
> ```Go
> type TableManifest struct {
> 	Level         uint8
> 	MaxCasCounter uint64
> ```

Should be named MaxCASCounter for consistency.

---

*[protos/manifest.proto, line 35 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuIMY0sQ7GmzJSPAWMT:-KuIMY0sQ7GmzJSPAWMU:b108svy) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/protos/manifest.proto#L35)):*
> ```protobuf
>         Operation Op = 2;
>         uint32 Level = 3;  // Only used for CREATE
>         uint64 MaxCasCounter = 4;  // Only used for CREATE
> ```

Should be named MaxCASCounter for consistency.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233)*

srh:
Review status: 0 of 17 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[backupstore.go, line 34 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuIKpJb3JxcOJFyT6OC:-KuLbFm3_X8MxkmLxO2U:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/backupstore.go#L34)):*

There should be some comment explaining the layout of a badger backup store here.

It consists of a directory with a BACKUPMANIFEST file (which we rewrite every time) and a bunch of files "backup-1", "backup-2", etc.  Every new incremental backup adds a new one of these, and rewrites the BACKUPMANIFEST file with all the info about backup files.
</blockquote></details>

Done.

---

*[backupstore.go, line 244 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuILOwyNAwNRNKEfjbz:-KuLbGYkehmc10M14uiZ:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/backupstore.go#L244)):*

"maxCASCounter" shouldn't be exposed in the user API as if it were a "CAS counter" value.  It should be called a backup identifier, something like that.
</blockquote></details>

Done.

---

*[kv.go, line 390 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuILmDELGT90yY2gdXX:-KuLbIYmfmNCv39ZvuUs:b-v7p8wf) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/kv.go#L390)):*

I really ought to make sure that the write loop doesn't call `s.Lock()` while holding writePendingCh.
</blockquote></details>

It was calling `s.Lock()` -- so the lock ordering between the two has now been made clear in comments and the deadlock bug here is fixed.

---

*[manifest.go, line 71 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuIMObQ2BkgOumiVw-b:-KuLbTtmPhlJHRAMPlCc:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/manifest.go#L71)):*

Should be named MaxCASCounter for consistency.
</blockquote></details>

Done.

---

*[protos/manifest.proto, line 35 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuIMY0sQ7GmzJSPAWMT:-KuLbVdUNv8XmCHbd8ym:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/protos/manifest.proto#L35)):*

Should be named MaxCASCounter for consistency.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233)*

manishrjain:
Reviewed 7 of 17 files at r1, 10 of 10 files at r2.
Review status: all files reviewed at latest revision, 34 unresolved discussions, some commit checks failed.

---

*[backupstore.go, line 244 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuILOwyNAwNRNKEfjbz:-KuMUEJoE_dYM1oJEfaf:b613ue7) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/backupstore.go#L244)):*

Done.
</blockquote></details>

100 chars.

And what does backupVersion represent?

Producer can be a channel, no?

---

*[backupstore.go, line 72 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMRfMsXUIex9GFUpXc:-KuMRfMsXUIex9GFUpXd:b-pxa963) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L72)):*
> ```Go
> 
> func backupFileName(id uint64) string {
> 	return fmt.Sprintf("backup-%d", id)
> ```

Integer padding would look nice. "backup-%06d" maybe.

---

*[backupstore.go, line 75 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMaKrNuopmuF4SUM3M:-KuMaKrNuopmuF4SUM3N:b-hunni) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L75)):*
> ```Go
> }
> 
> // RestoreBackup streams all changes in increasing key order to itemCh.
> ```

itemCh?

---

*[backupstore.go, line 76 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMaOjYsyw2Ejg14Tp8:-KuMaOjYsyw2Ejg14Tp9:bg887u3) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L76)):*
> ```Go
> 
> // RestoreBackup streams all changes in increasing key order to itemCh.
> func RestoreBackup(path string, consumer func(protos.BackupItem) error) (err error) {
> ```

Doesn't need to be a public function, if this is made a tool.

How is this different from build KV from backup function in KV?

---

*[backupstore.go, line 101 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMS0eHrE8hbXR5Xtmb:-KuMS0eHrE8hbXR5Xtmc:b-s00so7) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L101)):*
> ```Go
> 
> 	for _, id := range ids {
> 		iter, err := makeBackupFileIterator(path, id, bufferSize)
> ```

Place the helper functions above the first function where they get used at. Helps with the flow of reading.

---

*[backupstore.go, line 137 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMU1CLWAQXry7lE9RV:-KuMU1CLWAQXry7lE9RW:b-lrh85i) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L137)):*
> ```Go
> 	return nil
> }
> ```

Add a comment that this fulfills y.Iterator interface. It wasn't clear upfront.

---

*[backupstore.go, line 138 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMSM_8uzk9o4grfuvH:-KuMSM_8uzk9o4grfuvI:b-fs7bhl) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L138)):*
> ```Go
> }
> 
> type backupFileIterator struct {
> ```

move this to top?

---

*[backupstore.go, line 141 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMT2_-8w16_8RknX0u:-KuMT2_-8w16_8RknX0v:b2ni69m) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L141)):*
> ```Go
> 	fp         *os.File
> 	reader     *bufio.Reader
> 	firstKey   []byte
> ```

key

---

*[backupstore.go, line 142 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMT39XP1n1QglN9jiU:-KuMT39XP1n1QglN9jiV:bhnnr9) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L142)):*
> ```Go
> 	reader     *bufio.Reader
> 	firstKey   []byte
> 	firstValue y.ValueStruct
> ```

val or value. The first prefix isn't needed.

---

*[backupstore.go, line 181 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMSaUSgJRPzhzpt59u:-KuMSaUSgJRPzhzpt59v:b-g0lxl3) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L181)):*
> ```Go
> 		return
> 	}
> 	len, err := binary.ReadUvarint(bit.reader)
> ```

len var overwrites the len function. Use sz instead.

---

*[backupstore.go, line 184 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMSnEkMq8euXBh2r4g:-KuMSnEkMq8euXBh2r4h:bh922mi) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L184)):*
> ```Go
> 	if err != nil {
> 		bit.err = err
> 		bit.firstKey = nil
> ```

If there's error, just setting error should be sufficient, this would make `Valid()` false; without this key and value shouldn't be accessed anyway. So, key and value don't need to be zeroed out.

---

*[backupstore.go, line 222 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMTXsQloadSmAM5Fly:-KuMTXsQloadSmAM5Flz:b-9fuzlf) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L222)):*
> ```Go
> 
> func (bit *backupFileIterator) Valid() bool {
> 	return bit.firstKey != nil
> ```

This can check error instead of key being nil.

---

*[backupstore.go, line 242 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMSP7pHoNEWeWO0fpE:-KuMSP7pHoNEWeWO0fpF:bp7akyt) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L242)):*
> ```Go
> var (
> 	// ErrBackupKeysOutOfOrder means you tried storing a backup without supplying the keys in order.
> 	ErrBackupKeysOutOfOrder = errors.New("backup keys out of order")
> ```

move vars to top?

---

*[backupstore.go, line 289 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMV1USN-x46Nlo4r2r:-KuMV1USN-x46Nlo4r2s:b-woplf5) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L289)):*
> ```Go
> 		if err != nil {
> 			fileName := f.Name()
> 			_ = f.Close()
> ```

Add a function called Ignore which takes an error in y. That's better than assigning errors to underscore.

---

*[backupstore.go, line 299 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMVyxo5hQLL1qmhj1p:-KuMVyxo5hQLL1qmhj1q:bi3jdq) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L299)):*
> ```Go
> 
> 		for _, item := range items {
> 			if bytes.Compare(prevKey, item.Key) >= 0 && !first {
> ```

prevKey would be empty initially, so it can't be greater or equal to item.Key; right?

---

*[backupstore.go, line 306 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMVT3fiJwSWdPCGsOv:-KuMVT3fiJwSWdPCGsOw:b-hm01q5) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L306)):*
> ```Go
> 				return errors.Wrap(err, "NewBackup marshaling")
> 			}
> 			var lenBuf [binary.MaxVarintLen64]byte
> ```

This can be declared outside of the loop.

---

*[backupstore.go, line 323 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMWJtKjjOYIGtPCKgK:-KuMWJtKjjOYIGtPCKgL:bb01d6a) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L323)):*
> ```Go
> 
> 	newBackupID := maxBackupID + 1
> 	err = syncCloseRename(f, f.Name(), path, backupFileName(newBackupID))
> ```

You can join the path and backupFileName and then just send the dst path in one variable.

---

*[backupstore.go, line 330 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMWl4_LTLngYkRB2Cb:-KuMWl4_LTLngYkRB2Cc:bw8dc6g) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L330)):*
> ```Go
> 
> 	status.Backups = append(status.Backups, &protos.BackupStatusItem{
> 		BackupID: newBackupID,
> ```

Not clear about the use of backupid.

---

*[kv.go, line 390 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMXGl2aY7vtCF8UQb8:-KuMXGl2aY7vtCF8UQb9:b-uqkzt0) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L390)):*
> ```Go
> 	tables := make([]*skl.Skiplist, 1)
> 	s.writePendingCh <- struct{}{}
> 	tables[0] = s.mt.NewCopy()
> ```

This is unnecessarily going to block writes. You can just move the memtable to immutable (functions already exist for that), and then use it -- meanwhile a user can continue to do writes in a new memtable.

---

*[kv.go, line 1031 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuM_NfwFdvMomglVjwC:-KuM_NfwFdvMomglVjwD:b-qy8ogw) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1031)):*
> ```Go
> }
> 
> // StreamBackup sends a stream of backup items by calling `f` over and over again, until it's done,
> ```

s/f/consumer

---

*[kv.go, line 1033 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMbDQp_jwi-MZVvvEm:-KuMbDQp_jwi-MZVvvEn:bm50vpr) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1033)):*
> ```Go
> // StreamBackup sends a stream of backup items by calling `f` over and over again, until it's done,
> // at which point it returns.
> func (s *KV) StreamBackup(afterCas uint64, consumer func(protos.BackupItem) error) error {
> ```

Who needs this function?

---

*[kv.go, line 1083 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZkYP_L2m2jlPKEvg:-KuMZkYP_L2m2jlPKEvh:bklod7m) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1083)):*
> ```Go
> // not exist.  `source` supplies BackupItems (in increasing key order, hopefully) until it returns
> // a nil slice (or an error).
> func BuildKVFromBackup(opt *Options, source func() ([]protos.BackupItem, error)) (err error) {
> ```

This should be a tool / binary, instead of a call. You retrieve from a backup only once. And it's an offline process.

---

*[kv.go, line 1107 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMav-WBvnuuSF0A3_h:-KuMav-WBvnuuSF0A3_i:bvjtyyh) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1107)):*
> ```Go
> 	batch := []*Entry{}
> 	for {
> 		items, err := source()
> ```

The flow isn't clear to me. By calling source() you get back a list of backup items and an error? Who passes source to this?

---

*[kv.go, line 1125 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZeoO15Q5BpEJJdmt:-KuMZeoO15Q5BpEJJdmu:bz3dn3x) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1125)):*
> ```Go
> 
> 				if size >= opt.maxBatchSize {
> 					if err := kv.BatchSet(batch); err != nil {
> ```

kv.BatchSetAsync would be faster.

---

*[levels.go, line 325 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZMFOg3Zdq9bjI_Ad:-KuMZMFOg3Zdq9bjI_Ae:byvenkm) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/levels.go#L325)):*
> ```Go
> 			}
> 
> 			tbl, err := table.OpenTable(fd, maxCas, s.kv.opt.TableLoadingMode)
> ```

OpenTable shouldn't be passed the maxCas, it should track it by itself.

---

*[protos/backup.proto, line 26 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMXtumBzAH8GZLqtgg:-KuMXtunit4tHxPLjUzj:bokhgtd) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/protos/backup.proto#L26)):*
> ```protobuf
> message BackupItem {
>         bytes Key = 1;
>         uint64 Version = 2;
> ```

Is this the CAS Counter? If so, maybe call it Counter. So, the naming is consistent throughout the code base.

---

*[protos/backup.proto, line 42 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMY5NP06FIMq7FALhn:-KuMY5NP06FIMq7FALho:b-6bxqgk) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/protos/backup.proto#L42)):*
> ```protobuf
> // incremental backup after that.)
> message BackupStatusItem {
>         uint64 BackupID = 1;
> ```

Same here. If this is Cas counter, then call it Counter. Or, we can just change the naming to something more generic than CAS Counter, a global counter if you will.

---

*[skl/skl.go, line 515 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMXT4zN_7nAlsudp3P:-KuMXT4zN_7nAlsudp3Q:b-s3hmk9) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/skl/skl.go#L515)):*
> ```Go
> 
> // NewCopy copies the skiplist -- creating a new one with refcount 1.
> func (s *Skiplist) NewCopy() *Skiplist {
> ```

Not needed.

---

*[table/table.go, line 331 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMYqj_Pksvvy1GKO09:-KuMYqj_Pksvvy1GKO0A:b-jeioes) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/table/table.go#L331)):*
> ```Go
> 
> // MaxCasCounter is its maximum cas counter value (never zero), or 0 if there are no keys
> func (t *Table) MaxCasCounter() uint64 { return t.maxCasCounter }
> ```

This max cas counter belongs to the table. And it doesn't change once the table is created. So, this should be encoded within the table, instead of keeping this separately in manifest.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233)*

srh:
Review status: 12 of 17 files reviewed at latest revision, 34 unresolved discussions, some commit checks failed.

---

*[backupstore.go, line 244 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuILOwyNAwNRNKEfjbz:-KuMgLMm2kRNLrIT-XxY:bxf3lma) ([raw file](https://github.com/dgraph-io/badger/blob/8109192afb304e204cbde0b2c2f7c05ad578a49c/backupstore.go#L244)):*

100 chars.

And what does backupVersion represent?

Producer can be a channel, no?
</blockquote></details>

The version of the store at the time a backup gets made.  It's the latest cas counter value.

Producer could be a channel, but it shouldn't be.  The user doesn't need to use a channel to supply the backup items.

---

*[backupstore.go, line 72 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMRfMsXUIex9GFUpXc:-KuMudp4ubfGAKLqra8y:bwevvub) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L72)):*

Integer padding would look nice. "backup-%06d" maybe.
</blockquote></details>

I guess.

---

*[backupstore.go, line 75 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMaKrNuopmuF4SUM3M:-KuMff5S8yXdpdEShx0K:b-glsvqd) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L75)):*

itemCh?
</blockquote></details>

consumer used to be a channel.  Comment fixed.

---

*[backupstore.go, line 76 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMaOjYsyw2Ejg14Tp8:-KuMfvyEJzQMSAXDPH0t:b-78engu) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L76)):*

Doesn't need to be a public function, if this is made a tool.

How is this different from build KV from backup function in KV?
</blockquote></details>

Maybe they should be differently named -- renamed it to RetrieveBackup.  One reads from the backup store, the other makes a new KV and writes to it.

---

*[backupstore.go, line 101 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMS0eHrE8hbXR5Xtmb:-KuMr7wp7FWkUc1niYny:b-3n3wh) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L101)):*

Place the helper functions above the first function where they get used at. Helps with the flow of reading.
</blockquote></details>

OK.  Done.

---

*[backupstore.go, line 137 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMU1CLWAQXry7lE9RV:-KuMukHrI3ZSlosxU7jj:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L137)):*

Add a comment that this fulfills y.Iterator interface. It wasn't clear upfront.
</blockquote></details>

Done.

---

*[backupstore.go, line 138 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMSM_8uzk9o4grfuvH:-KuMulQdmtkTVMhMAiQR:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L138)):*

move this to top?
</blockquote></details>

Done.

---

*[backupstore.go, line 141 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMT2_-8w16_8RknX0u:-KuMulgxs8jNNiNC1CIQ:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L141)):*

key
</blockquote></details>

Done.

---

*[backupstore.go, line 142 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMT39XP1n1QglN9jiU:-KuMum5EVbMc8EQBoeMh:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L142)):*

val or value. The first prefix isn't needed.
</blockquote></details>

Done.

---

*[backupstore.go, line 181 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMSaUSgJRPzhzpt59u:-KuMumoJbhcjN88XsbXE:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L181)):*

len var overwrites the len function. Use sz instead.
</blockquote></details>

Done.

---

*[backupstore.go, line 184 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMSnEkMq8euXBh2r4g:-KuMgCGjWofZFrDa9Re8:b4kt7t5) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L184)):*

If there's error, just setting error should be sufficient, this would make `Valid()` false; without this key and value shouldn't be accessed anyway. So, key and value don't need to be zeroed out.
</blockquote></details>

They still should be.

---

*[backupstore.go, line 222 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMTXsQloadSmAM5Fly:-KuMvFL-MGLCr5ZNaUde:be8tqbz) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L222)):*

This can check error instead of key being nil.
</blockquote></details>

It doesn't seem like an iterator's valid before you call Rewind() on it.  Or is it? What's the interface supposed to be?

---

*[backupstore.go, line 242 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMSP7pHoNEWeWO0fpE:-KuMs-HDErBWILyCs-kQ:be4z0w8) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L242)):*

move vars to top?
</blockquote></details>

I guess?  Done.

---

*[backupstore.go, line 289 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMV1USN-x46Nlo4r2r:-KuMgUIvc2Idq7uanKWE:bcgnrlz) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L289)):*

Add a function called Ignore which takes an error in y. That's better than assigning errors to underscore.
</blockquote></details>

We can do that in a separate change -- I assigned to underscore because all the other places we could make a function Ignore for did the same.

---

*[backupstore.go, line 299 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMVyxo5hQLL1qmhj1p:-KuMgg_erd23XKZbkSus:b-u1tymr) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L299)):*

prevKey would be empty initially, so it can't be greater or equal to item.Key; right?
</blockquote></details>

Maybe that should be assigned to.

---

*[backupstore.go, line 306 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMVT3fiJwSWdPCGsOv:-KuMsQ1ZxnzJMyClA54r:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L306)):*

This can be declared outside of the loop.
</blockquote></details>

Done.

---

*[backupstore.go, line 323 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMWJtKjjOYIGtPCKgK:-KuMgu2VGqWy9_zJIOJt:b-ov2ysf) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L323)):*

You can join the path and backupFileName and then just send the dst path in one variable.
</blockquote></details>

We need to sync the directory so they're passed separately.

---

*[backupstore.go, line 330 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMWl4_LTLngYkRB2Cb:-KuMviiemFWpE0yQjvZo:b2aftfu) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L330)):*

Not clear about the use of backupid.
</blockquote></details>

In terms of representing the manifest it's redundant with the array index of the Backups array, but it's the number of the incremental backup.

---

*[kv.go, line 390 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMXGl2aY7vtCF8UQb8:-KuMlNdmuuAyWnhGf2E2:b1k9qiq) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L390)):*

This is unnecessarily going to block writes. You can just move the memtable to immutable (functions already exist for that), and then use it -- meanwhile a user can continue to do writes in a new memtable.
</blockquote></details>

I don't think there's any way to do that very nicely.  I'll look into it a bit more though.

---

*[kv.go, line 1031 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuM_NfwFdvMomglVjwC:-KuMwU1774DjGnj2pViq:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1031)):*

s/f/consumer
</blockquote></details>

Done.

---

*[kv.go, line 1033 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMbDQp_jwi-MZVvvEm:-KuMhHpwFfPOK3FLdkiY:b2fzry8) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1033)):*

Who needs this function?
</blockquote></details>

The user who wants to make a backup.

---

*[kv.go, line 1083 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZkYP_L2m2jlPKEvg:-KuMwVAKHu0WMyO1Lio5:b47vo53) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1083)):*

This should be a tool / binary, instead of a call. You retrieve from a backup only once. And it's an offline process.
</blockquote></details>

True but any tool will need a function like this to use, and if we saw fit to change this to build the KV store using internal API's then it should be part of KV.

---

*[kv.go, line 1107 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMav-WBvnuuSF0A3_h:-KuMke3OO-M3e0OMo-As:bliu92n) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1107)):*

The flow isn't clear to me. By calling source() you get back a list of backup items and an error? Who passes source to this?
</blockquote></details>

The caller would be the tool we create, or any other tool.

---

*[kv.go, line 1125 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZeoO15Q5BpEJJdmt:-KuMjxuxKh1XtYWhR6L-:ba86dam) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1125)):*

kv.BatchSetAsync would be faster.
</blockquote></details>

How so?

---

*[levels.go, line 325 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZMFOg3Zdq9bjI_Ad:-KuMk-SDM4rlXoYRFV7L:bl7bvca) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/levels.go#L325)):*

OpenTable shouldn't be passed the maxCas, it should track it by itself.
</blockquote></details>

The manifest tracks it.  We don't want to read the value from every table at startup.

---

*[protos/backup.proto, line 26 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMXtumBzAH8GZLqtgg:-KuMk8ousMZLyOPCc-th:brtph60) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/protos/backup.proto#L26)):*

Is this the CAS Counter? If so, maybe call it Counter. So, the naming is consistent throughout the code base.
</blockquote></details>

Sounds good, went with Counter.

---

*[protos/backup.proto, line 42 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMY5NP06FIMq7FALhn:-KuMxJeFOHl2bP0dxAQ-:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/protos/backup.proto#L42)):*

Same here. If this is Cas counter, then call it Counter. Or, we can just change the naming to something more generic than CAS Counter, a global counter if you will.
</blockquote></details>

Done.

---

*[table/table.go, line 331 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMYqj_Pksvvy1GKO09:-KuMkEj73dsTls-Ke6Fi:bfnnl4f) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/table/table.go#L331)):*

This max cas counter belongs to the table. And it doesn't change once the table is created. So, this should be encoded within the table, instead of keeping this separately in manifest.
</blockquote></details>

Meta-information about tables should be in the manifest so that we don't have to read every table at startup.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233)*

srh:
Current status: Main outstanding issue is calling NewCopy() on s.mt.  Moving that to s.imm isn't exactly straightforward.

---

Review status: 12 of 17 files reviewed at latest revision, 34 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233#-:-KuN9kikMUEYU95d3Si_:b-p2yp9d)*

manishrjain:
Review status: 12 of 17 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[backupstore.go, line 76 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMaOjYsyw2Ejg14Tp8:-KuO8-xV9JnLClPIKWQw:b-swqz58) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L76)):*

Maybe they should be differently named -- renamed it to RetrieveBackup.  One reads from the backup store, the other makes a new KV and writes to it.
</blockquote></details>

Packaging it into a tool shouldn't be too much code on top. Make it part of this PR. Also, we don't need to expose these as APIs. Users should use the tool to restore.

---

*[backupstore.go, line 184 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMSnEkMq8euXBh2r4g:-KuO8S1ZLWuISCoFHXkG:bb76mu4) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L184)):*

They still should be.
</blockquote></details>

Why?

---

*[backupstore.go, line 222 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMTXsQloadSmAM5Fly:-KuO8_SzXNgXzX1aKLKM:b-gjy9yv) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L222)):*

It doesn't seem like an iterator's valid before you call Rewind() on it.  Or is it? What's the interface supposed to be?
</blockquote></details>

It's not valid until you call Rewind.

---

*[backupstore.go, line 289 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMV1USN-x46Nlo4r2r:-KuO99EiTKmRQ_5KdG0j:b-r5yz13) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L289)):*

We can do that in a separate change -- I assigned to underscore because all the other places we could make a function Ignore for did the same.
</blockquote></details>

Add a TODO in that case.

---

*[backupstore.go, line 299 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMVyxo5hQLL1qmhj1p:-KuO9FLPxMdjmEC-zzN_:bowvl0w) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L299)):*

Maybe that should be assigned to.
</blockquote></details>

I meant first variable isn't required, because prevKey would be empty, making the compare cond work for you.

---

*[backupstore.go, line 330 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMWl4_LTLngYkRB2Cb:-KuO9X5s6dwvujAp01UZ:bwbo17a) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L330)):*

In terms of representing the manifest it's redundant with the array index of the Backups array, but it's the number of the incremental backup.
</blockquote></details>

So, not needed then?

In fact, a repeated Counter is all you need, not even BackupStatusItem?

---

*[backupstore.go, line 186 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuO8iQfZI55j3DcjbW6:-KuO8iQfZI55j3DcjbW7:b-l5pcmt) ([raw file](https://github.com/dgraph-io/badger/blob/63c52ab4ece4b007b1e07f0252a8b469fe647c0b/backupstore.go#L186)):*
> ```Go
> 
> // RetrieveBackup streams all changes from the backup store in increasing key order to consumer.
> func RetrieveBackup(path string, consumer func(protos.BackupItem) error) (err error) {
> ```

Make it private function. In fact, move this to the package which provides the tool.

---

*[kv.go, line 1033 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMbDQp_jwi-MZVvvEm:-KuO9mFUL3kfBNRr32es:brvpu20) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1033)):*

The user who wants to make a backup.
</blockquote></details>

By user, you don't mean end user?

---

*[kv.go, line 1083 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZkYP_L2m2jlPKEvg:-KuO9zA-iW4Hw4iF6DTq:bhmz1bx) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1083)):*

True but any tool will need a function like this to use, and if we saw fit to change this to build the KV store using internal API's then it should be part of KV.
</blockquote></details>

Yeah, these can be used internally, but there's no need to make them be available via public APIs. We can make them public if we decide to later.

---

*[kv.go, line 1107 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMav-WBvnuuSF0A3_h:-KuOAIsMmJHwdcxJmgmh:b-96lh3e) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1107)):*

The caller would be the tool we create, or any other tool.
</blockquote></details>

Let's do it as part of this PR.

---

*[kv.go, line 1125 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZeoO15Q5BpEJJdmt:-KuOAMd1GoZOjcwoTdfZ:b-xgnxg3) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1125)):*

How so?
</blockquote></details>

Is that a rhetorical question?

---

*[levels.go, line 325 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZMFOg3Zdq9bjI_Ad:-KuOAgtwSBdEw3Y2UvRO:b-697n5k) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/levels.go#L325)):*

The manifest tracks it.  We don't want to read the value from every table at startup.
</blockquote></details>

Manifest shouldn't track it. We open up all the tables at startup and parse their indices anyway.

---

*[table/table.go, line 331 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMYqj_Pksvvy1GKO09:-KuOB071wrv3wVG3jGEE:b3i99l0) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/table/table.go#L331)):*

Meta-information about tables should be in the manifest so that we don't have to read every table at startup.
</blockquote></details>

We read every table at startup. Otherwise, we wouldn't know their min, max key boundaries.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233)*

srh:
Review status: 12 of 17 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[backupstore.go, line 184 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMSnEkMq8euXBh2r4g:-KuQ5LDDxlbh4kvbY3Re:bbvepm0) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L184)):*

Why?
</blockquote></details>

Leaving garbage state around leaves behind a higher risk of missing bugs in testing, or in production.  Somebody could call Key() and get an old key.

---

*[backupstore.go, line 299 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMVyxo5hQLL1qmhj1p:-KuQ5u8a3jEutWvuK7VW:b-8lkj0r) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L299)):*

I meant first variable isn't required, because prevKey would be empty, making the compare cond work for you.
</blockquote></details>

My point is, it's empty because I forgot to write prevKey = item.Key in the for loop.

---

*[backupstore.go, line 330 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMWl4_LTLngYkRB2Cb:-KuQ62tQX5s3x__uztxp:b-79z5h1) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L330)):*

So, not needed then?

In fact, a repeated Counter is all you need, not even BackupStatusItem?
</blockquote></details>

If we stick it in a protobuf we get simpler future compatibility.

---

*[kv.go, line 1033 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMbDQp_jwi-MZVvvEm:-KuQ6BIMJ9Y2S1ld3ZlH:b-tlts6o) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1033)):*

By user, you don't mean end user?
</blockquote></details>

I do mean end user.  If you want to make a backup, you call StreamBackup.  Then you stream those items into a backup store.

---

*[kv.go, line 1125 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZeoO15Q5BpEJJdmt:-KuQ6Ou7BXH5btbbQFeL:b-l6lgee) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1125)):*

Is that a rhetorical question?
</blockquote></details>

No.  As far as I can tell it's not faster at all.  Why would it be?

---

*[table/table.go, line 331 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMYqj_Pksvvy1GKO09:-KuQ6bOUi8u3DfNoq7Cj:b-8yq78s) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/table/table.go#L331)):*

We read every table at startup. Otherwise, we wouldn't know their min, max key boundaries.
</blockquote></details>

We shouldn't have to do that either.  The manifest should hold the smallest/biggest key too so that we have flexibility of implementation in the future.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233)*

srh:
Review status: 10 of 20 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[backupstore.go, line 76 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMaOjYsyw2Ejg14Tp8:-KuRYYf0Rg9uT3WV5SDw:b2w45al) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L76)):*

Packaging it into a tool shouldn't be too much code on top. Make it part of this PR. Also, we don't need to expose these as APIs. Users should use the tool to restore.
</blockquote></details>

If Dgraph has some distributed backup system, for example, we're not going to want to shell out to the tool to restore.

---

*[backupstore.go, line 289 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMV1USN-x46Nlo4r2r:-KuRj1JUK2K92wHrNQwI:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L289)):*

Add a TODO in that case.
</blockquote></details>

Done.

---

*[backupstore.go, line 186 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuO8iQfZI55j3DcjbW6:-KuRj9l2Q1Q8lof0VHzV:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/63c52ab4ece4b007b1e07f0252a8b469fe647c0b/backupstore.go#L186)):*

Make it private function. In fact, move this to the package which provides the tool.
</blockquote></details>

Done.

---

*[kv.go, line 390 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMXGl2aY7vtCF8UQb8:-KuRY84zuALI5-sMSosX:bx56w2k) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L390)):*

I don't think there's any way to do that very nicely.  I'll look into it a bit more though.
</blockquote></details>

Took the ensureRoomForWrite loop we do and encapsulated it into pollRoomForWrite.  Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233)*

srh:
I think there's an issue where backup can get inconsistent versions of keys (it can get some keys which are too new), and this also is a problem that hits regular iteration.  So I'm looking into that right now.

---

Review status: 10 of 20 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233#-:-KuRkEI5nlSMiZ6zaY1w:b1d2ynm)*

manishrjain:
Review status: 10 of 20 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[backupstore.go, line 76 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMaOjYsyw2Ejg14Tp8:-KuSUy5cMELUqsK4IQf_:bfr3v3m) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/backupstore.go#L76)):*

If Dgraph has some distributed backup system, for example, we're not going to want to shell out to the tool to restore.
</blockquote></details>

If bash scripting doesn't help, at that point, we can expose these APIs. For now, a tool would do.

---

*[kv.go, line 1033 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMbDQp_jwi-MZVvvEm:-KuSWtNIH_Cnwi3xA_Hk:b-wv6hij) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1033)):*

I do mean end user.  If you want to make a backup, you call StreamBackup.  Then you stream those items into a backup store.
</blockquote></details>

I thought you were creating one with the backupmanifest etc. How is that called?

---

*[kv.go, line 1125 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZeoO15Q5BpEJJdmt:-KuSWDkBH44vGunYjphm:b-q3p8db) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1125)):*

No.  As far as I can tell it's not faster at all.  Why would it be?
</blockquote></details>

Because it won't block. So, if writes to disk are slow, multiple of these batches can be batched up together and written to disk together.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233)*

srh:
Review status: 10 of 20 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[kv.go, line 1125 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMZeoO15Q5BpEJJdmt:-KuSjNRwshJYzewsT_sY:b3iclbt) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1125)):*

Because it won't block. So, if writes to disk are slow, multiple of these batches can be batched up together and written to disk together.
</blockquote></details>

Okay, but that's also what having a large batch accomplishes.  Without that we need to add some code to limit the number of active ops.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233)*

manishrjain:
So, did you solve the inconsistent version of keys issue?

---

Review status: 10 of 20 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233#-:-KuX8lZvBU5wk-5gcrvo:bpza6pf)*

srh:
I got it confirmed with TestIteratorConcurrentWrites in branch sam/iter.

Basically the problem is, we get in-memory memtables first.  Then we unlock, and then we get level-zero memtables.  The merge iterator we use puts in-memory memtables ahead of level-zero in order of precedence.

Suppose s.imm and s.mt are tables A, B, C, D.  Our iterator reads those memtables as being the in-memory tables.  Then, suppose before we get level-zero memtables, we flush the tables A, B, C, D, and the subsequent table E.

Then we'll build an iterator with in-memory: [A, B, C, D], and level zero: [..., A, B, C, D, E].  The result is, according to the iterator, some changes in the new table E get overridden by the previous changes in A, B, C, or D.

I think we have another problem like this from compaction too, but it would happen less frequently.

I think a solution to this is to make iterators acquire a read lock on level K+1 while holding the lock on level K (with level -1 being the in-memory tables).  It might be useful to make the locks be the sort of RWLock you can get your place in line for, but not block, so that we can release the level K lock while acquiring K+1.  Operations that move information downward (from one successive level to the next) would have to acquire the locks for write in the same order, but may release them between.  This way a reader knows information can't pass downward through the locks it holds while it gets its lists of tables.

---

Review status: 10 of 20 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233#-:-KuXM4_UKG4geQ3PinAu:by44cas)*

manishrjain:
Not sure I understand. So, even if we end up compacting [A, B, C, D] with
the next level -- any updates from this set would overwrite (and hence
ignore) any key in the level below. I guess the biggest worry here is that
some new table gets compacted into the next level before we pick up tables
from that level?

Thanks!

Cheers,
Manish

Founder/CEO, Dgraph Labs
+61-487138931
https://calendly.com/mrjn

srh:
In the example with A, B, C, D, and E it's in-memory tables getting written to level zero.  The problem is from the iterator's perspective, it sees

Tables A..D in memory
(off-stage: some in-memory tables get written)
Tables A..E in level zero

For a different example, it could be the case that the iterator sees...

Tables P, Q, R in level 0
(off-stage: a new table gets written to level 0, they get compacted to level 1)
Tables S..Z in level 1, including writes which came after those in P, Q, and R.

The same thing at higher numbered levels is possible, but less likely because it requires later writes to quite miraculously "outrun" the iterator constructor through a serious of compactions but only catch up at a higher numbered levels.

---

Review status: 10 of 20 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233#-:-KuXdg4ER_HZZ8rUL81R:b-wu092f)*

srh:
Review status: 6 of 21 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[kv.go, line 1033 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/233#-KuMbDQp_jwi-MZVvvEm:-Ku_wxMgDhoNYcXxArGL:b-aoz5n1) ([raw file](https://github.com/dgraph-io/badger/blob/56aaa27a39bc757af725551fa68ffa121216a7b7/kv.go#L1033)):*

I thought you were creating one with the backupmanifest etc. How is that called?
</blockquote></details>

Using CreateBackupStore and NewBackup.

---

*[kv.go, line 1013 at r5](https://reviewable.io:443/reviews/dgraph-io/badger/233#-Ku_xJ58k_yZwlxeMV4V:-Ku_xJ58k_yZwlxeMV4W:b-uirqfy) ([raw file](https://github.com/dgraph-io/badger/blob/4d6f8284f8e29f2b7b5c06b32950671c153f55e1/kv.go#L1013)):*
> ```Go
> 
> // BackupStream holds information about a backup.
> type BackupStream struct {
> ```

The interface for this has been changed, because we need to not just stream backup items, we also need to emit info about what the version is of the backup.  The user creates a new BackupStream, and then calls StreamAndClose on that.  (The user may also call Counter() and StartTime().)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233)*

srh:
Review status: 6 of 21 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[iterator.go, line 346 at r5](https://reviewable.io:443/reviews/dgraph-io/badger/233#-Ku_zj8Mt0YR-ltk1N2Z:-Ku_zj8Mt0YR-ltk1N2_:b3ulbsk) ([raw file](https://github.com/dgraph-io/badger/blob/4d6f8284f8e29f2b7b5c06b32950671c153f55e1/iterator.go#L346)):*
> ```Go
> //   itr.Close()
> func (s *KV) NewIterator(opt IteratorOptions) *Iterator {
> 	s.tablesLock.RLock()
> ```

We also use chained lock acquisition for creating other iterators too.  So writes can't slip underneath iterator creation.

---

*[iterator.go, line 365 at r5](https://reviewable.io:443/reviews/dgraph-io/badger/233#-Ku_zEu09u6eClQ6zaPw:-Ku_zEu09u6eClQ6zaPx:bu70s8k) ([raw file](https://github.com/dgraph-io/badger/blob/4d6f8284f8e29f2b7b5c06b32950671c153f55e1/iterator.go#L365)):*
> ```Go
> }
> 
> // newBackupIterator returns an iterator that produces changes in the level controller (but not
> ```

We're now kicking the mutable memtable down instead of copying it.

We also now use this function to get the counter value associated with the backup.

We also acquire the level iterators by chained locking/unlocking from level 0 on upward.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/233)*

deepakjois:
This is out of date. Closing.

PR #235, opened 2017-09-18
make mmap-ing of value log optional

deepakjois:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/235)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,7 +33,7 @@ build_script:
 test_script:
   # Unit tests
   - ps: Add-AppveyorTest "Unit Tests" -Outcome Running
-  - go test -v github.com/dgraph-io/badger/...
+  - go test -vlog_mmap=false -v github.com/dgraph-io/badger/...
   - ps: Update-AppveyorTest "Unit Tests" -Outcome Passed
 
 notifications:
--- a/doc_test.go
+++ b/doc_test.go
@@ -19,7 +19,10 @@ package badger_test
 import (
 	"fmt"
 	"io/ioutil"
+	"runtime"
 	"sync"
+
+	"github.com/dgraph-io/badger/options"
 
 	"github.com/dgraph-io/badger"
 )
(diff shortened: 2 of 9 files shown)


deepakjois:
This also has the side effect of making the Windows build pass (as long as we disable mmap on it)

manishrjain:
I'm not entirely convinced that we should make value log loading be optional. It's not even clear when should a user turn it on and when off. If the problem that we intend to solve is to make this work on Windows, then setting max mmap to a reasonable value is a better solution.

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[doc_test.go, line 122 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/235#-KuMehUmMJlAQlTxPU1c:-KuMehUmMJlAQlTxPU1d:b-ory672) ([raw file](https://github.com/dgraph-io/badger/blob/3112f2a38fd5cb7a4cff522e4a61ad5cd1b595de/doc_test.go#L122)):*
> ```Go
> func ExampleKV_BatchSetAsync() {
> 	opt := badger.DefaultOptions
> 	if runtime.GOOS == "windows" {
> ```

Instead of doing this, why not set the max size of value to the size of the value log file? Then, do an mmap to double the size of value log file. That should avoid all this special handling, and be good for all systems.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/235#-:-KuMfPvuWef6RZ8GOuTf:b-ili4ls)*

deepakjois:
Fixing mmap on Windows is a separate issue. I already tried setting the max size of the value log to a lower value (even as low as 500MB) and it did not work. I am still investigating that. This change does fix windows Appveyor builds, but that is just a side effect and not the real motivation behind this change.

There are at least two good reasons that I can think of, to make mmap-ing optional:

1. We do not really understand mmap behavior fully. All we know is that it makes one random read benchmark significantly faster. For all other benchmarks, the latencies were always higher. Granted that they were negligibly higher (difference being in the order of 100s of nanoseconds), but across multiple benchmarks run on different instances the numbers _always_ were slightly higher.

2. Making mmap optional makes Badger run on more platforms. 32-bit platforms may not be able to run Badger at all if we use mmap as the only option.

Issue #227, opened 2017-09-15
Improve error of key-value limits


peterstace:
Ran into this limit while loading the stackoverflow data set.

Should also return two distinct errors so that it's clear whether it was the key or value that was too big.

Also, would be good to show the prefix of the actual key/value to aid with debugging if this occurs again.

manishrjain:
The value limit should be based on the value log file size. Also see: https://github.com/dgraph-io/badger/issues/212

peterstace:
PR: #237

deepakjois:
@peterstace, can this be closed now?

peterstace:
Yes, closed.

PR #237, opened 2017-09-19
Improve error for key value size limits

peterstace:
We now have two separate errors, one for keys and one for values. The error will contain the first 1KB of the offending key/value, and show what the limit was.

Also, the max value size is now the same as the value log size. It's not checked when replaying the value log, since the setting could have been different for a previous run of badger.

I increased the limit for the key size from 1MB to 16MB (found some data in the graph overflow dataset that possibly violates the 1MB limit).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/237)

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -87,8 +87,8 @@ func exceedsMaxKeySizeError(key []byte) error {
 }
 
 func exceedsMaxValueSizeError(value []byte, maxValueSize int64) error {
-	return errors.Errorf("Value with size %d exceeded ValueLogFileSize (%d). Key:\n%s",
-		len(value), maxValueSize, hex.Dump(value[:1<<10]))
+	return errors.Errorf("Value with size %d exceeded ValueLogFileSize (%dMB). Key:\n%s",
+		len(value), maxValueSize<<20, hex.Dump(value[:1<<10]))
 }
 
 const (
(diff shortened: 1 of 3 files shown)


peterstace:
PR: #227

manishrjain:
:lgtm: Address the comment before submitting.

---

Reviewed 3 of 3 files at r1.

---

*[value.go, line 83 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/237#-KuMyJ7FOCJordjABaBp:-KuMyJ7GB8DYzHY0SedZ:b-em4w8n) ([raw file](https://github.com/dgraph-io/badger/blob/3ebeeadf04fb345069f589559894c4f8ed7d2644/value.go#L83)):*
> ```Go
> )
> 
> const maxKeySize = 16 << 20
> ```

I think 1MB is good for keys. In comparison, our tables are only 64MB each. If we have some data in SO which violates this, we should avoid the exact index.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/237#-:-KuMyf2bxZhLKxTsKrwI:b-2bblxq)*

peterstace:

---

*[value.go, line 83 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/237#-KuMyJ7FOCJordjABaBp:-KuMz6C26lVvE5smxoec:b-7g05q6) ([raw file](https://github.com/dgraph-io/badger/blob/3ebeeadf04fb345069f589559894c4f8ed7d2644/value.go#L83)):*

I think 1MB is good for keys. In comparison, our tables are only 64MB each. If we have some data in SO which violates this, we should avoid the exact index.
</blockquote></details>

Done. Reverted to use 1MB.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/237)*

Issue #239, opened 2017-09-20
Test fails if we set SyncWrites = false


srh:
We have this line in `getTestOptions`

`	opt.SyncWrites = true // Some tests seem to need this to pass.`

Which turns out to be correct.

If we set SyncWrites = false, we get the panic,

```--- FAIL: TestValueBasic (0.87s)
panic: runtime error: slice bounds out of range [recovered]
	panic: runtime error: slice bounds out of range

goroutine 78796 [running]:
testing.tRunner.func1(0xc4201261e0)
	/usr/local/go/src/testing/testing.go:711 +0x2d2
panic(0x7ebfa0, 0xa8ac80)
	/usr/local/go/src/runtime/panic.go:491 +0x283
github.com/dgraph-io/badger.(*header).Decode(...)
	/home/srh/go/src/github.com/dgraph-io/badger/value.go:528
github.com/dgraph-io/badger.valueBytesToEntry(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/srh/go/src/github.com/dgraph-io/badger/value.go:933 +0x1c4
github.com/dgraph-io/badger.TestValueBasic(0xc4201261e0)
	/home/srh/go/src/github.com/dgraph-io/badger/value_test.go:76 +0x882
testing.tRunner(0xc4201261e0, 0x879388)
	/usr/local/go/src/testing/testing.go:746 +0xd0
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:789 +0x2de
```

PR #240, opened 2017-09-20
Make tests pass with SyncWrites = false

srh:
The problem was with the value log tests -- they weren't actually
writing to the value log with !SyncWrites because the values were too
small.  Connected to #239.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/240)


srh:
This is just tests -- merged to master in 8de624fae14f055ad6b1ab5c1b77a87776f2cd31.

PR #241, opened 2017-09-21
Mmap windows fix

### LINKED ISSUE #212: Couple of test failures  due to mmap changes
This seems to be happening because the AppVeyor instance does not have enough virtual address space to map the value log. I am guessing this should not prevent Badger from working on a Windows machine with enough RAM, but I don’t have one around to test.

screenshot from 2017-09-11 11-08-52

deepakjois:
Fixes #212. Please read commit logs for detailed info.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/241)

### DIFF
--- a/kv.go
+++ b/kv.go
@@ -261,7 +261,14 @@ func NewKV(optParam *Options) (out *KV, err error) {
 	if err = out.vlog.Replay(vptr, fn); err != nil {
 		return out, err
 	}
+
 	replayCloser.SignalAndWait() // Wait for replay to be applied first.
+
+	// Mmap writable log
+	lf := out.vlog.filesMap[out.vlog.maxFid]
+	if err = lf.mmap(2 * out.vlog.opt.ValueLogFileSize); err != nil {
+		return out, errors.Wrapf(err, "Unable to mmap RDWR log file")
+	}
 
 	out.writeCh = make(chan *request, kvWriteChCapacity)
 	out.closers.writes = y.NewCloser(1)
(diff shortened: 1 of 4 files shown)


manishrjain:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[value.go, line 165 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/241#-KuXkeUs1YnZhqslYrHI:-KuXkeUs1YnZhqslYrHJ:b216hke) ([raw file](https://github.com/dgraph-io/badger/blob/fc2bfb393afe1a98178a71c09ee069b5a4a8f683/value.go#L165)):*
> ```Go
> 		return errors.Wrapf(err, "Unable to munmap value log: %q", lf.path)
> 	}
> 	if err := lf.fd.Truncate(int64(offset)); err != nil {
> ```

Add a TODO here that we should confirm if we need to sync again.

---

*[value.go, line 282 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/241#-KuXljImOoaFAv3iiHwL:-KuXljImOoaFAv3iiHwM:b16bv03) ([raw file](https://github.com/dgraph-io/badger/blob/fc2bfb393afe1a98178a71c09ee069b5a4a8f683/value.go#L282)):*
> ```Go
> 	}
> 
> 	if truncate {
> ```

Instead of doing this, we can just pass a variable telling us not to truncate during value log GC. That way, only replay can do truncation, when mmap is not set yet.

In fact, just add a condition here that if the file is mmaped, don't truncate.

`if !mmap && truncate { ... }`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/241)*

PR #254, opened 2017-10-04
Transaction: Change Item.Value api

manishrjain:
- Make it so that the value returned is valid during the lifetime of transaction.
- We no longer pass in a function to Value().
- During a long-running iteration with prefetch, the callback to unlock value file log is called as soon as the value is copied over, thus freeing the value log files to do GC etc.
- Without prefetch, or if called item.Value directly, the transaction would store the file unlock callback, and would run it on txn.Discard().
- txn.Discard automatically gets called from txn.Commit, for convenience.
- All read only txns must discard the transaction.
- Provide a View function to make read-only txns easier to be called.
- Error out on txn reuse.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/254)

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -45,7 +45,7 @@ var (
 	ErrReadOnlyTxn = errors.New("No sets or deletes are allowed in a read-only transaction.")
 
 	// ErrDiscardedTxn is returned if a previously discarded transaction is re-used.
-	ErrDiscardedTxn = errors.New("This transaction has been discarded. Recreate a new one.")
+	ErrDiscardedTxn = errors.New("This transaction has been discarded. Create a new one.")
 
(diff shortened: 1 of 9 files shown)


janardhan1993:
:lgtm:

---

Reviewed 9 of 9 files at r1.
Review status: 7 of 9 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/254#-:-KvaE4Tq0WDH7usGl4Ua:bnfp4nl)*

manishrjain:
Merged 83aa09d2bd8af49ed2261fa1217b879a6f20c1c2.

PR #261, opened 2017-10-10
Make DB.Open() take a value instead of pointer.

deepakjois:
* Txn.NewIterator() also takes a value, so this makes the API consistent
in terms of options handling.

* Using values instead of pointers might be a little bit safer, because
we don’t have to worry about the options from changing underneath.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/261)

### DIFF
--- a/db.go
+++ b/db.go
@@ -164,9 +164,7 @@ func replayFunction(out *DB) func(entry, valuePointer) error {
 }
 
 // Open returns a new DB object.
-func Open(optParam *Options) (db *DB, err error) {
-	// Make a copy early and fill in maxBatchSize
-	opt := *optParam
+func Open(opt Options) (db *DB, err error) {
 	opt.maxBatchSize = (15 * opt.MaxTableSize) / 100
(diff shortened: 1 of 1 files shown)


manishrjain:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---

*[db.go, line 168 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/261#-Kw4rlYoKVc-vOEUL2vw:-Kw4rlYpLVFGN2Z4ZpFj:bxfky8l) ([raw file](https://github.com/dgraph-io/badger/blob/319712eb6ea909719e0ff8248adf1dc82ef81058/db.go#L168)):*
> ```Go
> // Open returns a new DB object.
> func Open(opt Options) (db *DB, err error) {
> 	// Make a copy early and fill in maxBatchSize
> ```

Remove the comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/261#-:-Kw4ros8lX-dWuANTaOx:bnfp4nl)*

PR #281, opened 2017-10-20
create table with smaller initial buffer size

wuxingyi:
In my case,  I created 100 badger instances, write some data to them, and then close them. Every thing looks good when data writing, but memory usage exploded when DB closing, triggered an oom, I set Options.MaxTableSize to 4MB, but the problem still exists. After profiling, I find the the memory usage comes from calling NewTableBuilder, and the buffer is hard coded to 64MB.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/281)

### DIFF
--- a/table/builder.go
+++ b/table/builder.go
@@ -85,8 +85,8 @@ type Builder struct {
 // NewTableBuilder makes a new TableBuilder.
 func NewTableBuilder() *Builder {
 	return &Builder{
-		keyBuf:     newBuffer(32 << 20),
-		buf:        newBuffer(64 << 20),
+		keyBuf:     newBuffer(1 << 20),
+		buf:        newBuffer(1 << 20),
 		prevOffset: math.MaxUint32, // Used for the first element!
 	}
 }


deepakjois:
:lgtm:

Waiting for @manishrjain to take a look as well.

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/281#-:-Kx69SdxDEmljEh31ZkC:bcmjx2d)*

manishrjain:
The code change is not required. We can just set the new buffer size to 1MB each.

---

Reviewed 5 of 5 files at r1.

---

*[table/builder.go, line 88 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/281#-Kx6H-i1SyyqMsFG27Ct:-Kx6H-i1SyyqMsFG27Cu:b-7od1eh) ([raw file](https://github.com/dgraph-io/badger/blob/1b56c599018e873b5d4411d0e258743a898f206f/table/builder.go#L88)):*
> ```Go
> func NewTableBuilder(tableSize int64) *Builder {
> 	return &Builder{
> 		keyBuf:     newBuffer(int(tableSize) / 2),
> ```

No need for this change. newBuffer just takes a suggestion for how much to allocate in advance. Even if we take a low ball approach here, it would work just as well. So, setting this by default to 1MB each is sufficient. It would avoid the API changes elsewhere in this code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/281#-:-Kx6HCT_P1VtbzPvjaX8:b3rerf3)*

wuxingyi:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[table/builder.go, line 88 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/281#-Kx6H-i1SyyqMsFG27Ct:-Kx6lLTlMvjBDlz2Wht3:b-seavvm) ([raw file](https://github.com/dgraph-io/badger/blob/1b56c599018e873b5d4411d0e258743a898f206f/table/builder.go#L88)):*

No need for this change. newBuffer just takes a suggestion for how much to allocate in advance. Even if we take a low ball approach here, it would work just as well. So, setting this by default to 1MB each is sufficient. It would avoid the API changes elsewhere in this code.
</blockquote></details>

cool, I think it will save much memory in our case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/281)*

manishrjain:
:lgtm:

---

Reviewed 1 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/281#-:-Kx6niVXccEWf2VJIsv1:bnfp4nl)*

PR #289, opened 2017-10-25
Update magic number from 2 to 3.

### LINKED ISSUE #282: Update magic number of Badger just before v1.0 release
This would probably mean anybody who is using Badger prior to v1.0 will have to re-import their data (include Dgraph users).

deepakjois:
Fixes #282.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/289)

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -211,7 +211,7 @@ func (mf *manifestFile) addChanges(changesParam []*protos.ManifestChange) error 
 var magicText = [4]byte{'B', 'd', 'g', 'r'}
 
 // The magic version number.
-const magicVersion = 2
+const magicVersion = 3
 
 func helpRewrite(dir string, m *Manifest) (*os.File, int, error) {
 	rewritePath := filepath.Join(dir, manifestRewriteFilename)


deepakjois:
We should wait to merge this. While we are waiting, I would like to better understand the implications of this for stuff that is currently using Badger, and the guidance we can offer.

So if I understand correctly, any data that was written using a previous build of Badger cannot be read once this change goes in. One hacky workaround is to manually rewrite the Badger version from 2 -> 3 in the manifest file. Even then, it would only work if the data was written using a build of Badger post v0.8.

manishrjain:
We need to provide a way to backup Badger, to a standard key-value file; possibly using protocol buffers for extensibility. This needs to go into both v0.8 and v1.0 -- see BoltDB for reference. Once we have that, we need an import tool which can rebuild Badger.

magik6k:
Just a note - we are using badger with this patch applied in IPFS 0.4.12-rc1, see https://github.com/gxed/badger/commits/master.

deepakjois:
@magik6k This is now in master, but there are some other changes as well (see #305 ). We decided to get those in because we would have needed to change the manifest version for those change.

The changes in #305 also change the `Txn.Set()` API slightly.

PR #291, opened 2017-10-25
Add DB.Dump() and DB.Load() for backup purposes.

### LINKED ISSUE #135: Have a way to create backup of Badger

deepakjois:
We write a backup of the latest snapshot in the DB (including all
previous versions) to a protobuf-encoded file.

Fixes #135

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/291)


manishrjain:
Review status: 0 of 5 files reviewed at latest revision, 13 unresolved discussions.

---

*[backup.go, line 14 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI5elnAGx_M1MmgS1Y:-KxI5elnAGx_M1MmgS1Z:b-49telt) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L14)):*
> ```Go
> )
> 
> func writeToDisk(entries []*protos.KVPair, w io.Writer) error {
> ```

if len(entries) == 0 return nil.

Also, no need to buffer entries before writing to w. This function can just be `writeTo(entry *protos.KVPair, w io.Writer)`.

---

*[backup.go, line 15 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI67B5qVL5KO4PblqQ:-KxI67B5qVL5KO4PblqR:b-2ztm3d) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L15)):*
> ```Go
> 
> func writeToDisk(entries []*protos.KVPair, w io.Writer) error {
> 	buf := make([]byte, binary.MaxVarintLen64+entries[0].Size())
> ```

Why have this buffer? Why not just write directly to io.Writer?

---

*[backup.go, line 27 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI6Rb4gWfSRm0jh0Hs:-KxI6Rb4gWfSRm0jh0Ht:b8hscyo) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L27)):*
> ```Go
> 		}
> 		b = b[n:]
> 		n, err = entry.MarshalTo(b)
> ```

entry.Marshal would avoid having to pass in a buffer.

---

*[backup.go, line 42 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI4RMOYEuDaVSfq5-G:-KxI4RMOYEuDaVSfq5-H:bqmx2i3) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L42)):*
> ```Go
> //
> // This can be used to backup the data in a database at a given point in time.
> func (db *DB) Dump(w io.Writer) error {
> ```

s/Dump/Backup.

And, do return the `txn.ReadTs()`, so a user can later use it for an incremental backup.

In fact, allow passing a threshold ts to Backup, so it can ignore any versions below it. Would be useful for incremental backup.

---

*[backup.go, line 57 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI76Zm0PtNqnDFT4-K:-KxI76Zm0PtNqnDFT4-L:b-r66gw8) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L57)):*
> ```Go
> 			}
> 
> 			entries = append(entries, &protos.KVPair{
> ```

No need to batch up.

---

*[backup.go, line 67 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI7A9CU4l_7fmkyy8M:-KxI7A9CU4l_7fmkyy8N:b-tmpupt) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L67)):*
> ```Go
> 			if count == 1000 {
> 				// Write entries to disk
> 				if err := writeToDisk(entries, w); err != nil {
> ```

io.Writer could be anything, doesn't have to be disk. Could be `http.ResponseWriter`

---

*[backup.go, line 86 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI9SnNfNykqMc2QfHD:-KxI9SnNfNykqMc2QfHE:b9i98yc) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L86)):*
> ```Go
> // DB.Load() should be called on a database that is not running any other
> // concurrent transactions while it is running.
> func (db *DB) Load(r io.Reader) error {
> ```

Does this need to be an exposed function? Instead, we need a binary which can rebuild Badger from backup. That way, we can use async writes, memory map for LSM etc. -- optimizations for high write throughput.

---

*[backup.go, line 115 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI7qkiflA25iHdWW3T:-KxI7qkiflA25iHdWW3U:b-ghk26m) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L115)):*
> ```Go
> 	for {
> 		buf, err := br.Peek(binary.MaxVarintLen64)
> 		if err == io.EOF {
> ```

Need to handle ErrBufferFull.

---

*[backup.go, line 148 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI8kU_SwiuWRMs18O3:-KxI8kU_SwiuWRMs18O4:b4vga6c) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L148)):*
> ```Go
> 			Value:    e.Value,
> 			UserMeta: e.UserMeta[0],
> 			// TODO do we have to set the Meta field ourselves
> ```

No need. batch set async would write the meta.

---

*[backup.go, line 150 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI92N-1wGJEL_hZi3n:-KxI92N-1wGJEL_hZi3o:brvqgnf) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L150)):*
> ```Go
> 			// TODO do we have to set the Meta field ourselves
> 		})
> 		count++
> ```

Why do you need count? len(entries) should do.

---

*[backup.go, line 153 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI8t1EkVJ8U4TY6V8A:-KxI8t1EkVJ8U4TY6V8B:b-n65lwe) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L153)):*
> ```Go
> 
> 		if count == 1000 {
> 			if err := batchSetAsyncIfNoErr(entries); err != nil {
> ```

batch set here makes sense.

---

*[backup.go, line 157 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI8wW8xfvYcymzC0-E:-KxI8wW8xfvYcymzC0-F:b-sap7b1) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L157)):*
> ```Go
> 			}
> 			count = 0
> 			entries = []*entry{}
> ```

entries = entries[:0]. This would reuse the slice instead of creating a new one.

---

*[errors.go, line 80 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI981rRMIa6rI3TNWw:-KxI981rRMIa6rI3TNWx:b-lsmnen) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/errors.go#L80)):*
> ```Go
> 
> 	ErrInvalidBackupData = errors.New(
> 		"Backup data cannot be read")
> ```

This can fit in one line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/291)*

manishrjain:
Reviewed 5 of 5 files at r1.
Review status: 4 of 5 files reviewed at latest revision, 13 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/291)*

deepakjois:
Review status: 2 of 5 files reviewed at latest revision, 13 unresolved discussions.

---

*[backup.go, line 14 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI5elnAGx_M1MmgS1Y:-KxLgP0LD21mPRDGI4z9:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L14)):*

if len(entries) == 0 return nil.

Also, no need to buffer entries before writing to w. This function can just be `writeTo(entry *protos.KVPair, w io.Writer)`.
</blockquote></details>

Done.

---

*[backup.go, line 15 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI67B5qVL5KO4PblqQ:-KxLa4ZRScoksVBAosEN:b-mb4ltc) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L15)):*

Why have this buffer? Why not just write directly to io.Writer?
</blockquote></details>

I used the same approach used in dgraphloader. I am guessing it saves us multiple allocations. If we write directly to io.Writer, won’t we be allocating on every loop iteration.

I am not sure if we should care though. I have changed it for now.

---

*[backup.go, line 27 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI6Rb4gWfSRm0jh0Hs:-KxLaSs4IngbB69XmaJB:bw9jeyb) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L27)):*

entry.Marshal would avoid having to pass in a buffer.
</blockquote></details>

And would also allocate a new slice on every iteration. It would have been nice to have an API in protobufs that directly writes into a Writer, but we don’t AFAICT.

---

*[backup.go, line 42 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI4RMOYEuDaVSfq5-G:-KxLaffe_7ZLnx36eegR:ba2jj3x) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L42)):*

s/Dump/Backup.

And, do return the `txn.ReadTs()`, so a user can later use it for an incremental backup.

In fact, allow passing a threshold ts to Backup, so it can ignore any versions below it. Would be useful for incremental backup.
</blockquote></details>

 "Backup" implies a more all-encompassing thing, which isn’t what we are doing here. I was thinking of tools like mysqldump, which dump an SQL output to file. Hence the name.

Adjusted the code to read and return a timestamp

---

*[backup.go, line 57 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI76Zm0PtNqnDFT4-K:-KxLll4lGpluj4ylKoOg:b-ihha2l) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L57)):*

No need to batch up.
</blockquote></details>

Done. Changed the code here to directly write into the writer.

---

*[backup.go, line 67 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI7A9CU4l_7fmkyy8M:-KxLbiio7MZJ_EnZWcmr:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L67)):*

io.Writer could be anything, doesn't have to be disk. Could be `http.ResponseWriter`
</blockquote></details>

Done.

---

*[backup.go, line 86 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI9SnNfNykqMc2QfHD:-KxLfbUZCgGI0eVfL_US:b-ms3a5i) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L86)):*

Does this need to be an exposed function? Instead, we need a binary which can rebuild Badger from backup. That way, we can use async writes, memory map for LSM etc. -- optimizations for high write throughput.
</blockquote></details>

A binary CLI tool is definitely needed. I was assuming it would be located in a different folder, under `cmd`. That tool would still need access to some public method like this.

---

*[backup.go, line 115 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI7qkiflA25iHdWW3T:-KxLsH0-Guz3iHlnC1Oq:b91ryso) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L115)):*
I am not sure. The documentation says :

> If Peek returns fewer than n bytes, it also returns an error explaining why the read is short.

So I have added this additional check: `if len(buf) > 0 && len(buf) < binary.MaxVarintLen64 && err != nil`

---

*[backup.go, line 148 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI8kU_SwiuWRMs18O3:-KxLbVaA6U-Et6Lz12nT:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L148)):*

No need. batch set async would write the meta.
</blockquote></details>

Done.

---

*[backup.go, line 150 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI92N-1wGJEL_hZi3n:-KxLm043C2JJpBIQI4HZ:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L150)):*

Why do you need count? len(entries) should do.
</blockquote></details>

Done.

---

*[backup.go, line 153 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI8t1EkVJ8U4TY6V8A:-KxLm0WlHKayVP9KyI9b:b-lc5l7j) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L153)):*

batch set here makes sense.
</blockquote></details>

Are you saying we should use batchSet instead of batchSetAsync here? Arent there benefits to using async writes here?

---

*[backup.go, line 157 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI8wW8xfvYcymzC0-E:-KxLmEFkksDbs9iHhQ6U:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L157)):*

entries = entries[:0]. This would reuse the slice instead of creating a new one.
</blockquote></details>

Done.

---

*[errors.go, line 80 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI981rRMIa6rI3TNWw:-KxLruvyW2n4RemMkc4Z:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/errors.go#L80)):*

This can fit in one line.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/291)*

deepakjois:
Review status: 2 of 5 files reviewed at latest revision, 13 unresolved discussions.

---

*[backup.go, line 115 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI7qkiflA25iHdWW3T:-KxLvgZAiB8tUSTzC9b5:b-vwco5u) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L115)):*
<details><summary><i>Previously, deepakjois (Deepak Jois) wrote…</i></summary><blockquote>

I am not sure. The documentation says :

> If Peek returns fewer than n bytes, it also returns an error explaining why the read is short.

So I have added this additional check: `if len(buf) > 0 && len(buf) < binary.MaxVarintLen64 && err != nil`
</blockquote></details>

Also, this code was adapted from dgraphloader, so maybe it needs to be changed there as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/291)*

manishrjain:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[backup.go, line 15 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI67B5qVL5KO4PblqQ:-KxM9iM2zOwI4vl5M_wC:b-1vl62f) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L15)):*
<details><summary><i>Previously, deepakjois (Deepak Jois) wrote…</i></summary><blockquote>

I used the same approach used in dgraphloader. I am guessing it saves us multiple allocations. If we write directly to io.Writer, won’t we be allocating on every loop iteration.

I am not sure if we should care though. I have changed it for now.
</blockquote></details>

var [8]buf would be allocated on stack and be cheaper instead of a heap allocation.

---

*[backup.go, line 27 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI6Rb4gWfSRm0jh0Hs:-KxMA07a9P-_c4AilMTd:b-hnn21t) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L27)):*
<details><summary><i>Previously, deepakjois (Deepak Jois) wrote…</i></summary><blockquote>

And would also allocate a new slice on every iteration. It would have been nice to have an API in protobufs that directly writes into a Writer, but we don’t AFAICT.
</blockquote></details>

The allocation was happening anyway -- because you were manually doing it above. So, don't worry about it.

---

*[backup.go, line 42 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI4RMOYEuDaVSfq5-G:-KxMAX7rgYHUoSu0wgm1:b-cpt0je) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L42)):*
<details><summary><i>Previously, deepakjois (Deepak Jois) wrote…</i></summary><blockquote>

 "Backup" implies a more all-encompassing thing, which isn’t what we are doing here. I was thinking of tools like mysqldump, which dump an SQL output to file. Hence the name.

Adjusted the code to read and return a timestamp


</blockquote></details>

I think it should be called Backup. We are provisioning for incremental backup as well. And have a tool to restore to a new Badger. Unless there're specific things that we are not doing, which we must as part of the "backup" terminology.

Also, have writer as the first option. So `Backup(w, 0)`. Rename `ts` to `since`.

---

*[backup.go, line 115 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI7qkiflA25iHdWW3T:-KxNGJOmNDeroxtL_SuA:bm27aj9) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L115)):*
<details><summary><i>Previously, deepakjois (Deepak Jois) wrote…</i></summary><blockquote>

Also, this code was adapted from dgraphloader, so maybe it needs to be changed there as well.
</blockquote></details>

Remove all the var int stuff. Better to use Uint64 directly.

---

*[backup.go, line 153 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI8t1EkVJ8U4TY6V8A:-KxNGogxsiTkvZwKuLGm:b13jq1h) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L153)):*
<details><summary><i>Previously, deepakjois (Deepak Jois) wrote…</i></summary><blockquote>

Are you saying we should use batchSet instead of batchSetAsync here? Arent there benefits to using async writes here?
</blockquote></details>

I think I just meant this is good. Ignore.

---

*[backup.go, line 25 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxMAAtJq_RR8hU2UbtR:-KxMAAtJq_RR8hU2UbtS:bix7725) ([raw file](https://github.com/dgraph-io/badger/blob/8e9ddfdc59681f4033fd1f39a818a4683cb0014d/backup.go#L25)):*
> ```Go
> 		return err
> 	}
> 	if _, err = w.Write(buf); err != nil {
> ```

_, err = w.Write(buf)
return err

---

*[backup.go, line 41 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxNFwgtSJIIBCwgNe4b:-KxNFwgtSJIIBCwgNe4c:bogeafk) ([raw file](https://github.com/dgraph-io/badger/blob/8e9ddfdc59681f4033fd1f39a818a4683cb0014d/backup.go#L41)):*
> ```Go
> 	var tsNew uint64
> 	err := db.View(func(txn *Txn) error {
> 		opts := DefaultIteratorOptions
> ```

Actually, take in a parameter for whether we should backup all versions or just one.

---

*[backup.go, line 56 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxNFJ5hkiwRuXNy8Lmq:-KxNFJ5hkiwRuXNy8Lmr:b-mjk5qy) ([raw file](https://github.com/dgraph-io/badger/blob/8e9ddfdc59681f4033fd1f39a818a4683cb0014d/backup.go#L56)):*
> ```Go
> 
> 			entry := &protos.KVPair{
> 				Key:      y.Safecopy([]byte{}, item.Key()),
> ```

How about a `y.Copy()` function, which would only take one arg?

---

*[backup.go, line 63 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxNFXVKLnlLAoF4nHNK:-KxNFXVKLnlLAoF4nHNL:b-gmpuno) ([raw file](https://github.com/dgraph-io/badger/blob/8e9ddfdc59681f4033fd1f39a818a4683cb0014d/backup.go#L63)):*
> ```Go
> 
> 			// Write entries to disk
> 			if err := writeTo(entry, w); err != nil {
> ```

Just a nitpick. `writeTo(w, entry)` looks a slightly nicer api call.

---

*[backup.go, line 156 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxNHL0dXz1KgIs4XbXz:-KxNHL0dXz1KgIs4XbY-:bv241fz) ([raw file](https://github.com/dgraph-io/badger/blob/8e9ddfdc59681f4033fd1f39a818a4683cb0014d/backup.go#L156)):*
> ```Go
> 			return err
> 		}
> 	}
> ```

You need to wait for wg here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/291)*

deepakjois:
Review status: 3 of 8 files reviewed at latest revision, 11 unresolved discussions.

---

*[backup.go, line 15 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI67B5qVL5KO4PblqQ:-KxNNsQh6FwMIhqrElEx:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L15)):*

var [8]buf would be allocated on stack and be cheaper instead of a heap allocation.
</blockquote></details>

Done.

---

*[backup.go, line 27 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI6Rb4gWfSRm0jh0Hs:-KxNNtdTtKuHwdMx13fS:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L27)):*

The allocation was happening anyway -- because you were manually doing it above. So, don't worry about it.
</blockquote></details>

Done.

---

*[backup.go, line 42 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI4RMOYEuDaVSfq5-G:-KxNO5-ZUrn-21n2GEA_:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L42)):*

I think it should be called Backup. We are provisioning for incremental backup as well. And have a tool to restore to a new Badger. Unless there're specific things that we are not doing, which we must as part of the "backup" terminology.

Also, have writer as the first option. So `Backup(w, 0)`. Rename `ts` to `since`.
</blockquote></details>

Done.

---

*[backup.go, line 86 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI9SnNfNykqMc2QfHD:-KxNQFvnCH06lFT5HaIZ:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L86)):*
<details><summary><i>Previously, deepakjois (Deepak Jois) wrote…</i></summary><blockquote>

A binary CLI tool is definitely needed. I was assuming it would be located in a different folder, under `cmd`. That tool would still need access to some public method like this.
</blockquote></details>

Done.

---

*[backup.go, line 115 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI7qkiflA25iHdWW3T:-KxNQHEINfxnR354WGAU:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L115)):*

Remove all the var int stuff. Better to use Uint64 directly.
</blockquote></details>

Done.

---

*[backup.go, line 153 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI8t1EkVJ8U4TY6V8A:-KxNQI9E3VPCLPd1U7PM:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L153)):*

I think I just meant this is good. Ignore.
</blockquote></details>

Done.

---

*[backup.go, line 25 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxMAAtJq_RR8hU2UbtR:-KxNO417AyeGLtYUUbHO:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/8e9ddfdc59681f4033fd1f39a818a4683cb0014d/backup.go#L25)):*

_, err = w.Write(buf)
return err
</blockquote></details>

Done.

---

*[backup.go, line 41 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxNFwgtSJIIBCwgNe4b:-KxNOSqax7RHcJxwsRB8:byskjul) ([raw file](https://github.com/dgraph-io/badger/blob/8e9ddfdc59681f4033fd1f39a818a4683cb0014d/backup.go#L41)):*

Actually, take in a parameter for whether we should backup all versions or just one.
</blockquote></details>

It is a bit awkward to take in a parameter to backup all versions, as well as have a timestamp. What does it mean if both are set to something?

I am thinking it is possible to break it up. We could have a tool to delete all previous versions (`badger purge`). Using `badger backup` in combination with `badger purge` (after or before backup) should give us what we want. Anything more sophisticated we can build it in later.

---

*[backup.go, line 56 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxNFJ5hkiwRuXNy8Lmq:-KxNPJ1ItUStycv6DH0T:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/8e9ddfdc59681f4033fd1f39a818a4683cb0014d/backup.go#L56)):*

How about a `y.Copy()` function, which would only take one arg?
</blockquote></details>

Done.

---

*[backup.go, line 63 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxNFXVKLnlLAoF4nHNK:-KxNQEVpFy8kjSBU1aUI:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/8e9ddfdc59681f4033fd1f39a818a4683cb0014d/backup.go#L63)):*

Just a nitpick. `writeTo(w, entry)` looks a slightly nicer api call.
</blockquote></details>

Done.

---

*[backup.go, line 156 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxNHL0dXz1KgIs4XbXz:-KxNQr2xGOYVpl2_5P_z:b-gys9wn) ([raw file](https://github.com/dgraph-io/badger/blob/8e9ddfdc59681f4033fd1f39a818a4683cb0014d/backup.go#L156)):*

You need to wait for wg here.
</blockquote></details>

Done.

Sorry, bad copypasta.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/291)*

peterstace:
Review status: 3 of 6 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[backup.go, line 15 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI67B5qVL5KO4PblqQ:-KxPqgTet2q9IQuz0q0v:b-xblp9s) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L15)):*
<details><summary><i>Previously, deepakjois (Deepak Jois) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

FYI, the reason we use a buffer in the bulk loader is because we know we're writing directly to an `os.File` (better to do one big write rather than small writes). Since here we're writing via an `os.Writer`, it's up to whoever passes it in to provide an implementation that does the appropriate buffering (if they want it).

---

*[backup.go, line 66 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxPsipsuwZyd5B2WrsZ:-KxPsipsuwZyd5B2Wrs_:b5j9eey) ([raw file](https://github.com/dgraph-io/badger/blob/a12c86e747eacfd6da9069363d3429cc7cabfc12/backup.go#L66)):*
> ```Go
> 		return nil
> 	})
> 	return tsNew, err
> ```

`tsNew` is never set, so will always be zero. Did you mean to return `ts` instead?

---

*[backup.go, line 110 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxPuiflaP_EdpiHaRSC:-KxPuiflaP_EdpiHaRSD:bdqvwiy) ([raw file](https://github.com/dgraph-io/badger/blob/a12c86e747eacfd6da9069363d3429cc7cabfc12/backup.go#L110)):*
> ```Go
> 		}
> 
> 		for cap(unmarshalBuf) < int(sz) {
> ```

Could use an `if` instead of `for`, since the body will either execute 0 or 1 times.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/291)*

deepakjois:
Review status: 3 of 6 files reviewed at latest revision, 13 unresolved discussions.

---

*[backup.go, line 15 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI67B5qVL5KO4PblqQ:-Kxf29X81mwMZAT0jtJF:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L15)):*
<details><summary><i>Previously, peterstace (Peter Stace) wrote…</i></summary><blockquote>

FYI, the reason we use a buffer in the bulk loader is because we know we're writing directly to an `os.File` (better to do one big write rather than small writes). Since here we're writing via an `os.Writer`, it's up to whoever passes it in to provide an implementation that does the appropriate buffering (if they want it).
</blockquote></details>

Done.

---

*[backup.go, line 66 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxPsipsuwZyd5B2WrsZ:-Kxf2P7O0lswreA-DhR3:brh3klk) ([raw file](https://github.com/dgraph-io/badger/blob/a12c86e747eacfd6da9069363d3429cc7cabfc12/backup.go#L66)):*
<details><summary><i>Previously, peterstace (Peter Stace) wrote…</i></summary><blockquote>

`tsNew` is never set, so will always be zero. Did you mean to return `ts` instead?
</blockquote></details>

There is a typo two lines above. I meant to set `tsNew` instead of `ts` there.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/291)*

deepakjois:
Review status: 3 of 6 files reviewed at latest revision, 13 unresolved discussions.

---

*[backup.go, line 110 at r3](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxPuiflaP_EdpiHaRSC:-Kxf2rOl8C4bQXu-g5uJ:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/a12c86e747eacfd6da9069363d3429cc7cabfc12/backup.go#L110)):*
<details><summary><i>Previously, peterstace (Peter Stace) wrote…</i></summary><blockquote>

Could use an `if` instead of `for`, since the body will either execute 0 or 1 times.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/291)*

manishrjain:
:lgtm: Some comments.

---

Reviewed 2 of 3 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[backup.go, line 15 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI67B5qVL5KO4PblqQ:-KxfMBxuNZAlj25sDY0-:bruwfsw) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L15)):*
<details><summary><i>Previously, deepakjois (Deepak Jois) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

if err := func(); err != nil { return err }. Have seen it a few times, so would make sense to try to make it a habit by some setting in your editor.

---

*[backup.go, line 41 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxNFwgtSJIIBCwgNe4b:-KxfMRICLuR5geTrMZkq:b2nhktp) ([raw file](https://github.com/dgraph-io/badger/blob/8e9ddfdc59681f4033fd1f39a818a4683cb0014d/backup.go#L41)):*
<details><summary><i>Previously, deepakjois (Deepak Jois) wrote…</i></summary><blockquote>

It is a bit awkward to take in a parameter to backup all versions, as well as have a timestamp. What does it mean if both are set to something?

I am thinking it is possible to break it up. We could have a tool to delete all previous versions (`badger purge`). Using `badger backup` in combination with `badger purge` (after or before backup) should give us what we want. Anything more sophisticated we can build it in later.
</blockquote></details>

OK.

---

*[backup.go, line 34 at r4](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxfMNT9HPvKFKXYp0vJ:-KxfMNT9HPvKFKXYp0vK:b-dmdzho) ([raw file](https://github.com/dgraph-io/badger/blob/29dba49f46343df9eaa61c6399e071c9a9e42ca9/backup.go#L34)):*
> ```Go
> //
> // This can be used to backup the data in a database at a given point in time.
> func (db *DB) Backup(w io.Writer, ts uint64) (uint64, error) {
> ```

s/ts/since?

---

*[backup.go, line 115 at r4](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxfMyuBfVSA5vPq1SC2:-KxfMyuBfVSA5vPq1SC3:bij2b77) ([raw file](https://github.com/dgraph-io/badger/blob/29dba49f46343df9eaa61c6399e071c9a9e42ca9/backup.go#L115)):*
> ```Go
> 
> 		e := &protos.KVPair{}
> 		_, err = io.ReadFull(br, unmarshalBuf[:sz])
> ```

if _, err = ...; err != nil { .. } here and elsewhere.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/291#-:-KxfNH-xIEeDuw0jQpAy:b-xkzuoz)*

deepakjois:
Review status: 5 of 6 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[backup.go, line 15 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxI67B5qVL5KO4PblqQ:-KxfUmqxuEflBDoMlSVw:b-gj3ob8) ([raw file](https://github.com/dgraph-io/badger/blob/e6b06709542ff143fc2f743000e8f0d96e532c22/backup.go#L15)):*

if err := func(); err != nil { return err }. Have seen it a few times, so would make sense to try to make it a habit by some setting in your editor.
</blockquote></details>

Done. Don’t think an editor setting would help. Ideally the linting tools would help flag this as a potential simplification. I will look.

---

*[backup.go, line 34 at r4](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxfMNT9HPvKFKXYp0vJ:-KxfTiOFC-D1cDzR_LgS:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/29dba49f46343df9eaa61c6399e071c9a9e42ca9/backup.go#L34)):*

s/ts/since?
</blockquote></details>

Done.

---

*[backup.go, line 115 at r4](https://reviewable.io:443/reviews/dgraph-io/badger/291#-KxfMyuBfVSA5vPq1SC2:-KxfUZY0HHPoNH0ln6o6:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/29dba49f46343df9eaa61c6399e071c9a9e42ca9/backup.go#L115)):*

if _, err = ...; err != nil { .. } here and elsewhere.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/291)*

PR #329, opened 2017-11-28
Use a helper method to setup Badger before running test.

### LINKED ISSUE #304: Refactor tests to avoid code repetition
Our test files have gotten very big (`db_test.go` is over 1000 lines). There seems to be a fair bit of repetition in the code. For example, this bit of code is repeated in every test:

```Go
…<snip>…
dir, err := ioutil.TempDir("", "badger")
	require.NoError(t, err)
	defer os.RemoveAll(dir)
	db, err := Open(getTestOptions(dir))
	require.NoError(t, err)
…<snip>…
```

We could refactor this out into a helper func and reuse it.

deepakjois:
Most of the tests just setup a temporary directory, open Badger
and then run the test. This commit refactors the repeated code
in each test into a helper method.

Some tests have been left alone because they do something extra
like setting different options, closing/re-opening the DB etc.

Fixes #304.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/329)

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -71,513 +71,435 @@ func txnDelete(t *testing.T, kv *DB, key []byte) {
 	require.NoError(t, txn.Commit(nil))
 }
 
-func TestWrite(t *testing.T) {
+// Opens a badger db and runs a a test on it.
+func runBadgerTest(t *testing.T, opts *Options, test func(t *testing.T, db *DB)) {
 	dir, err := ioutil.TempDir("", "badger")
 	require.NoError(t, err)
 	defer os.RemoveAll(dir)
-	kv, err := Open(getTestOptions(dir))
-	require.NoError(t, err)
-	defer kv.Close()
-
-	for i := 0; i < 100; i++ {
-		txnSet(t, kv, []byte(fmt.Sprintf("key%d", i)), []byte(fmt.Sprintf("val%d", i)), 0x00)
+	if opts == nil {
+		opts = new(Options)
+		*opts = getTestOptions(dir)
 	}
+	db, err := Open(*opts)
+	require.NoError(t, err)
+	defer db.Close()
+	test(t, db)
+}
+
+func TestWrite(t *testing.T) {
+	runBadgerTest(t, nil, func(t *testing.T, db *DB) {
+		for i := 0; i < 100; i++ {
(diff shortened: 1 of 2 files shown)


manishrjain:
Reviewed 2 of 2 files at r1.

---

*[db_test.go, line 75 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/329#-L-4rhjuD5IDL3-akIi3:-L-4rhjvrz8TJS7_UtlR:box325l) ([raw file](https://github.com/dgraph-io/badger/blob/87a7b9506459d85e4359add1165bb4784d5d4bf9/db_test.go#L75)):*
> ```Go
> 
> // Opens a badger db and runs a a test on it.
> func runBadgerTest(t *testing.T, opts *Options, test func(t *testing.T, db *DB)) {
> ```

This could have been a `setup` function, which does the setup and just gives back a DB object. I wonder what's the benefit of taking in a function, compared to returning a DB object. Also this approach adds an extra level of indentation for every test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/329)*

deepakjois:

---

*[db_test.go, line 75 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/329#-L-4rhjuD5IDL3-akIi3:-L-4uL6ZgiItqLqu_VNW:bp5czkj) ([raw file](https://github.com/dgraph-io/badger/blob/87a7b9506459d85e4359add1165bb4784d5d4bf9/db_test.go#L75)):*

This could have been a `setup` function, which does the setup and just gives back a DB object. I wonder what's the benefit of taking in a function, compared to returning a DB object. Also this approach adds an extra level of indentation for every test.
</blockquote></details>

We will still need to do the cleanup by calling `os.RemoveAll(…)` and `db.Close()`, and we can only do that once we have run the test. So it is best to take in a func. It is an approach similar to when sub-tests are run using `t.Run(…)`

Not sure I understand why the extra level of indentation is so significant?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/329)*

manishrjain:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/329#-:-L-5SvU9K_mlR8UilD4L:bnfp4nl)*

PR #337, opened 2017-12-04
Change TTL resolution to nanoseconds and allow subsecond TTLs

brunotm:
Currently if we set a key with a subsecond TTL the entry is automatically expired because the TTL resolution is in seconds.

This change allows setting subsecond TTLS.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/337)

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -334,7 +334,7 @@ func isDeletedOrExpired(vs y.ValueStruct) bool {
 	if vs.ExpiresAt == 0 {
 		return false
 	}
-	return vs.ExpiresAt <= uint64(time.Now().Unix())
+	return vs.ExpiresAt <= uint64(time.Now().UnixNano())
 }
 
 // parseItem is a complex function because it needs to handle both forward and reverse iteration
--- a/structs.go
+++ b/structs.go
@@ -81,7 +81,7 @@ type Entry struct {
 	Key       []byte
 	Value     []byte
 	UserMeta  byte
-	ExpiresAt uint64 // time.Unix
+	ExpiresAt uint64 // time.UnixNano
 	meta      byte
 
 	// Fields maintained internally.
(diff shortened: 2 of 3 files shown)


manishrjain:
:lgtm: Thanks for your contribution.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/337#-:-L-ZFtPkUsthHQUAQ3y0:bc9pyo)*

peterstace:
I think we might have a backwards compatibility issue here.

E.g. someone is using the current version of badger, they have some keys with TTL 24 hours into the future. They upgrade to badger containing this change. When they start up badger, they data could get garbage collected immediately because the persisted TTLs (in seconds) are now interpreted to be in nanos and (and so will be in the past).

manishrjain:
That's a good point. I missed that.

Any resolutions here? Otherwise, this change can't be part of v1.0.

peterstace:
I can't think of any clean solutions. There's lots of hacky things you could do though. 64 bits gives you a lot of space to play with. E.g. MSB set to 0 could indicate timestamp (remaining 63 bits) is in seconds. MSB set to 1 could indicate timestamp (remaining 63 bits) is in nanos.

deepakjois:
I am going to revert this PR for now, and create a new issue, where we can get feedback on how many people really want this feature immediately.

We would like to get this in, but it will definitely involve a major version bump. So its just a matter of timing.

deepakjois:
I created #339 to track this. Please vote on the issue or add further comments there.

brunotm:
@peterstace @manishrjain 
>E.g. someone is using the current version of badger, they have some keys with TTL 24 hours into the future. They upgrade to badger containing this change. When they start up badger, they data could get garbage collected immediately because the persisted TTLs (in seconds) are now interpreted to be in nanos and (and so will be in the past).

I hadn't anticipated that, sorry.

@deepakjois 
> We would like to get this in, but it will definitely involve a major version bump. So its just a matter of timing.

Thanks.

Thanks.

Issue #339, opened 2017-12-05
Change TTL resolution to nanoseconds and allow subsecond TTLs


deepakjois:
Currently, Badger allows setting TTL values on key-value pairs. We store these TTL values as unix timestamps with a per-second granularity. This was probably an oversight, but it prevents setting sub-second TTL values. Changing the way timestamps are stored is a backwards incompatible change.

I am creating this issue to track feedback and comments on how urgently this change is required, and whether it is useful. We would also need to put some thought into how we backup/restore a Badger database that stores the timestamps currently in seconds, so that it works with a version of Badger that stores the timestamps in nanoseconds.

brunotm:
+1

magik6k:
What I see could be done to avoid breakage would be allowing the user to interact with those values more directly - Add getTime field to options use it's return as 'time' when possible, fall back to seconds when nil. 

This would allow subsecond resolution with the side effect of enabling custom time definitions - like 'n app launches', 'n inserts' , etc

deepakjois:
> Add getTime field to options use it's return as 'time' when possible, fall back to seconds when nil.

Interesting idea. Let me try it out.

deepakjois:
Pls see #342 and let me know what you think.

manishrjain:
Let's postpone this until we're ready for v2.0 launch. For now, second level TTL should suffice. We've tied into it for this version anyway.

deepakjois:
We thought more about this, and there are some inconvenient tradeoffs to think about:

* Like we have already discussed, to implement this we need to change the way we store timestamps on disk. It will break any existing Badger DBs that store TTL values. The only way out for those using TTLs now would be to backup and restore using a special version of our backup tool that would rewrite timestamps.
* Strictly speaking, a change in the on-disk format means we should be increment the Badger manifest version as well. This would be very disruptive for _all_ existing Badger users. Given 1.0 just came out, this is not something we want to do in a hurry.

For now we think it is better to leave things as they are. We would like to hear more about real-world use cases for a sub-second TTL. If there is a convincing case for us to make such a disruptive change in the future, we could do it. Otherwise, we could just wait and slip these changes in, in the future when we are compelled to upgrade the Badger manifest version.

jiminoc:
[...]

PR #342, opened 2017-12-07
[RFC] Add option to use sub-second TTL

deepakjois:
This is a first draft of a change based on suggestion by @magik6k in #339.

It is a bit hacky, but not as hacky as some other suggestions that were made earlier. This change also allows use to gracefully drop this hack when we do a major version upgrade (we will still need to provide a viable way to backup the existing data such that it can be imported seamlessly when we do switch to subsecond TTLs by default).

If this approach is acceptable, I will refactor the code a bit more and add tests as well.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/342)

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -327,14 +327,20 @@ func (it *Iterator) Next() {
 	}
 }
 
-func isDeletedOrExpired(vs y.ValueStruct) bool {
+func isDeletedOrExpired(vs y.ValueStruct, useSubsecondTTL bool) bool {
 	if vs.Meta&bitDelete > 0 {
 		return true
 	}
 	if vs.ExpiresAt == 0 {
 		return false
 	}
-	return vs.ExpiresAt <= uint64(time.Now().Unix())
+	var now int64
+	if useSubsecondTTL {
+		now = time.Now().UnixNano()
+	} else {
+		now = time.Now().Unix()
+	}
+	return vs.ExpiresAt <= uint64(now)
 }
 
 // parseItem is a complex function because it needs to handle both forward and reverse iteration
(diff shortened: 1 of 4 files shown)


magik6k:
Well I guess that may work for sub-second TTLs, these changes shouldn't break anything, though I'd still much prefer being able to pass custom TTL provider function in options as time isn't really the only TTL metric that is useful.

I'm not sure how that would work with the backup tool.

gwik:
Why not have configurable the unit instead of a bool ?

The options parameter would be a time.Duration.
Default would be 1*time.Second.
You can configure 10*time.Millisecond or 1*time.Nanosecond if you like.

```
t.UnixNano() / timeUnit
```
Gives you the key.

It might not be suitable if the idea is to migrate to nanos for everyone.

deepakjois:
@magik6k Is there an example that you have in mind of any other KV store that provides a TTL provider function. Could you provide a link that I could refer to.

brunotm:
I like the idea of having the resolution configurable by a time.Duration as suggested by @gwik. Beyond flexible is simple and clean.

manishrjain:
@magik6k : The idea to have not just TTL, but generic functions is interesting. But, functions can't be stored on disk, so we'll have to create a new language to support those cases; which might not be generally useful at this point.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/342#-:-L02WxwxSWVm99ifIo8x:b-pth47v)*

deepakjois:
We had a bit of discussion here regarding the change, and decided that it is better to just do a major version upgrade, and record the timestamps in the database as nanosecond values.

@gwik @brunotm I believe that will address your concerns. We already have an API that can set TTL values using `time.Duration`, so it is flexible in that it can accept values at any granularity. The problem happened because we converted those timestamps into seconds rather than storing them directly.

Closing this for now. Will make a breaking change in another branch and release it as v2.0. The v2.0 release should happen in the next couple of weeks, and incorporate some other changes that are in flight as well.

Issue #350, opened 2017-12-11
syscall.O_DSYNC compile time error


gedw99:
Compiled for ios 

go version go1.10beta1 darwin/amd64

````
matcha build gomatcha.io/matcha/examples
go build -pkgdir=/Users/apple/workspace/go/pkg/matcha/pkg_darwin_arm -tags matcha ios -buildmode=c-archive -o /var/folders/wp/ff6sz9qs6g71jnm12nj2kbyw0000gp/T/gomobile-work-656669643/matcha-arm.a /var/folders/wp/ff6sz9qs6g71jnm12nj2kbyw0000gp/T/gomobile-work-656669643/src/iosbin/main.go failed: exit status 2

# github.com/dgraph-io/badger/y
../../../../../../dgraph-io/badger/y/file_dsync.go:24:21: undefined: syscall.O_DSYNC

````

gedw99:
fixed by adjusting build flags so it works on ios.
Will test on android once i get the test harness setup.

file_dsync.go:
// +build !dragonfly,!freebsd,!windows,!darwin

file_nodsync.go:
// +build dragonfly freebsd windows darwin

Would you mind making this change and see if its all OK through your CI ?

deepakjois:
I pushed a change to #351. Will merge once the tests have passed.

gedw99:
great - thanks for super quick followup.

PR #378, opened 2017-12-27
correct fix for #350

Stebalien:
Re-enables DSYNC on MacOS and iOS. As far as I can tell, these platforms *do*
support DSYNC, it's just that the constant is missing from
syscall (semi-deprecated and frozen).

I have neither an iPhone nor a Mac so someone should test this. However, it
appears to build (tested by manually building with `GOOS` and `GOARCH`).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/378)

### DIFF
--- a/y/file_dsync.go
+++ b/y/file_dsync.go
@@ -1,4 +1,4 @@
-// +build !dragonfly,!freebsd,!windows,!darwin
+// +build !dragonfly,!freebsd,!windows
 
 /*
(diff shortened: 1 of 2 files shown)


Stebalien:
(not copyrightable, feel free to re-do this PR locally if it's easier).

deepakjois:
This change is committed as 39bfb9cbbe922c50a53aa839476d052b19eaa519

PR #400, opened 2018-01-16
defrag tool and related changes.

deepakjois:
* Code for defrag command

* Method to do offline GC on all files w/ discard ratio
  set to 0.0.

* Refactoring to keep code DRY.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/400)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.vscode
+badger.test
--- a/README.md
+++ b/README.md
@@ -598,6 +598,11 @@ observe the full IOPS throughput provided by modern SSDs. In Dgraph, we have set
 it to 128. For more details, [see this
 thread](https://groups.google.com/d/topic/golang-nuts/jPb_h3TvlKE/discussion).
 
+-- **Are there any linux specific settings that I should use?**
+
+We *highly* recommend setting max file descriptors to a high number depending upon the expected size of
+you data.
+
 ## Contact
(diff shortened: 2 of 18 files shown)


deepakjois:
Still working on the LSM tree compaction part, but would appreciate a review for the offline GC which runs on all value log files w/ discard ratio set to 0.0.

deepakjois:
Now added code to do offline compaction as well.

janardhan1993:
For offline we need to remove tombstones, i don't see any code for that. 

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/400#-:-L31m16Xaho-HOC_zrN6:b-8eeckf)*

deepakjois:
Hmm…yes. I forgot about that case. I haven’t really dug too deep into the code for the LSM stuff. The compaction part was easy to infer, but actually removing entries from the tables sounds a bit more involved. Any pointers for that?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/400#-:-L32RMFcmJbSk51DF1vn:b-kyu7wo)*

janardhan1993:
While compacting if the final table belongs to last level, we can skip adding all the entries to table builder for which bitDelete or expired(there should be a function for checking that) is set in the meta

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/400#-:-L35CuhWsxnBGZhXbj6a:b-w0kcrf)*

Issue #401, opened 2018-01-18
Bug? Missing keys after backup then restore of a large db


osiloke:
I'm having trouble grasping the results of backing up and restoring a badgered database.
I have a database which has a couple thousand entries. I run badger_info which gives me the result below:
```
> badger_info --dir original_db
[Manifest]
MANIFEST           1562  2018-01-18T10:37:32+01:00
[Level 0]
000085.sst        15587  2018-01-18T10:22:43+01:00 0
[Level 1]
000084.sst       362235  2018-01-15T19:39:06+01:00 1
[Value Log]
000000.vlog     4286777  2018-01-18T10:37:32+01:00
[Summary]
Level 0 size: 15587
Level 1 size: 362235
Total index size: 377822
Value log size: 4286777
Abnormalities: None.
0 extra files.
0 missing files.
0 empty files.
0 truncated manifests.
```
I run badger backup to backup the the database. Then i restore the database to a different folder and run badger_info which then gives me the result below.

```
> badger_info --dir new_db
[Manifest]
MANIFEST             28  2018-01-18T11:03:15+01:00
[Level 0]
000001.sst       111648  2018-01-18T11:03:15+01:00 0
[Value Log]
000000.vlog     3414691  2018-01-18T11:03:15+01:00
[Summary]
Level 0 size: 111648
Total index size: 111648
Value log size: 3414691
Abnormalities: None.
0 extra files.
0 missing files.
0 empty files.
0 truncated manifests.
```
I assume it compacted the database to just level 0 based on this result. if i tail the value log i can see that it has at least most of the entries.

This all looks fine, based on my assumptions, until i iterate all keys using the snippet from the docs. It returns only 3 keys.
This does not seem to correlate with the metric data returned by badger_info

Draal:
I hit the same issue with backup / restore, this is the gist which reproduces it
https://gist.github.com/Draal/f7feb3f5ca58d5902ff134e446de31b1

I've tried different options/closing opening, etc with no luck.

MoZhonghua:
I found the db.orc.curRead is not updated after item loaded,  so later txn can only see item with ts == 0. Two ways to fix:
  1. set db.orc.curRead to max(db.orc.curRead, max(ts of entries)) after load
  2. ignore item's old timestamp and re-insert them using normal txn

janardhan1993:
Fixed in master.

Draal:
@janardhan1993  unfortunately the fix in master didn't fix the case below:
[...]

PR #404, opened 2018-01-20
Fix bug for #401

MoZhonghua:
Update db.orc.curRead to maximum version of loaded items, else later transaction will get old readTs and can't see these items.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/404)

### DIFF
--- a/backup.go
+++ b/backup.go
@@ -98,6 +98,7 @@ func (db *DB) Load(r io.Reader) error {
 		}
 	}
 
+	maxVersion := uint64(0)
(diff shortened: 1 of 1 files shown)


janardhan1993:
Thanks for the fix, really appreciate it. We need to update nextCommit before writing to new badger instance. And ofcourse update readTs at end as you have done in this PR.
I have raised a PR based on your change with unit tests. #406. So closing this.

Issue #410, opened 2018-01-26
Deadlock on single-threaded, large writes, including reproducer


hillu:
Hi,

I am trying to implement an inverted index based on bitmaps stored in BadgerDB and have run into deadlock issues that seem similar to those described in #384. I have put together a [reproducer](https://github.com/hillu/badger-lockup-reproducer).

There are bulk writes that can become rather large and often need to be split across multiple transactions. I had also experimented with reducing the transaction size before hitting the `ErrTxnTooBig` error, but that has not helped in my case.

manishrjain:
I ran the reproducer, added some logs as well. But, it kept on going and all it did was freeze my laptop. I haven't been able to replicate the deadlock.

hillu:
@manishrjain Did the program ever reach entry 134?

hillu:
@manishrjain Would you like to share the log statements that you added?

manishrjain:
Sorry, haven't had a chance to investigate further. Let me get back.

hillu:
Meanwhile, I have been able to reproduce the issue with a few older versions of badger. It also happens with 1.2.0, 1.1.0, 1.0.0. (The test could not be run as-is using previous versions due to changes in 5ad4981c9a2940efe8bd052a2bda2d7a508d90c8.)

The program will use well over 5 GB virtual memory.

hillu:
I just discovered the `golang.org/x/trace` support and added a local http server to my reproducer. The last entries of the event trace lead me to think that something is going wrong as `(*DB).writeRequests` calls `(*DB).vlog.write`:
```
12:43:11.580872	 .  2705	... writeRequests called. Writing to value log
12:43:11.590391	 .  9519	... Writing to memtable
12:43:11.590401	 .    10	... 2 entries written
12:43:11.611276	 . 20875	... writeRequests called. Writing to value log
12:43:11.623989	 . 12713	... Writing to memtable
12:43:11.630259	 .  6269	... 1631 entries written
12:43:11.641416	 . 11157	... writeRequests called. Writing to value log
12:43:11.645462	 .  4046	... Writing to memtable
12:43:11.645478	 .    16	... 3 entries written
12:43:12.845980	1.200502	... writeRequests called. Writing to value log
```
These log output lines should correspond roughly to the event trace.
```
2018/01/30 12:43:11 entry 133, cardinality: 1630
2018/01/30 12:43:11 todo: 1630, done: 0
[...]

PR #413, opened 2018-01-31
Commit: early callback

hillu:
The change to transaction.go moves processing of the callbacks to a point at which no writes that might 
cause a new value log file to be opened. It solves #410 for me.

This is based on the assumptions that
1. the callback mechanism is only used for releasing read locks. (Is this correct? Do you have any other use planned?)
2. releasing the read locks before flushing write ops to disk is okay. API-wise, this should be the case, as far as I can see (from looking at `(*Item).Value` documentation).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/413)

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -1225,6 +1225,7 @@ func TestGetSetDeadlock(t *testing.T) {
 	opt.ValueLogFileSize = 1 << 20
 	db, err := Open(opt)
 	require.NoError(t, err)
+	defer db.Close()
 
 	val := make([]byte, 1<<19)
 	key := []byte("key1")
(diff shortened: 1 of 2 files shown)


hillu:
Meh. Of course, this by itself does not help if another transaction (perhaps from another goroutine) is holding read locks when a new value log is being started.

janardhan1993:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/413#-:-L6Gm-Q0Y_uJZHJPnUrH:bnfp4nl)*

PR #423, opened 2018-02-24
add Iterator.Close to each examples

h12w:
Iterator must be closed, otherwise, serious memory leak might happen

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/423)

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,7 @@ err := db.View(func(txn *badger.Txn) error {
     }
     fmt.Printf("key=%s, value=%s\n", k, v)
   }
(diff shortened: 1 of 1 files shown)


pawanrawal:
Thanks for the change @h12w. Could you change `Close` calls to be a defer statement after the iterator is initialized? Its more idiomatic in Go to defer the execution of such statements which do cleanup.

```
it := txn.NewIterator(opts)
defer it.Close()
```

h12w:
Sure. Done.

Issue #372, opened 2017-12-20
Can badger open database in readonly mode without locking?


vrealzhou:
Is it possible to add option to open database in readonly mode which no need to lock the database? It could share the database between different read only processes?

deepakjois:
I don’t believe this is possible without a major design change, that will affect multiple layers in the code. So I am not sure if this is something we will look into actively.

ivanjaros:
I just did a blocking test and seems badger does not block readers and writers so you could utilize multiple readers and to avoid writes, add api layer on top of badger.

manishrjain:
This can be done -- would essentially need to avoid doing the lock on DB, and modify the scope of APIs a bit. I don't know how much demand is there for this feature. If we get 5-10 thumbs-up on the top post, we can build this.

manishrjain:
The PR which adds this functionality is merged.

PR #432, opened 2018-03-07
Non-locking, read-only opens

allenluce:
This introduces a read-only option on the DB struct. The main effect is to force all transactions to be read-only. It also skips anything that might write to disk.

It will put a shared lock on the directory when it's opened read-only. This will allow other read-only opens but will deny read-write opens until all readers have closed. If the database is already open read-write and an attempt is made to open it again (read-only or read-write), an error occurs.

It also:

 - Fails if the manifest doesn't exist on read-only open
 - Does not attempt to truncate the manifest
 - Skips compactors and memtable
 - All vlogs are opened read-only

This change assumes that there will be no need to write anything during read-only operations, no need to do garbage collection or compaction or revising vlogs or anything. I'm not sure that this is a 100% valid assumption.

This addresses #372


allenluce:
The checks fail because internal references to `dgraph-io/badger` pull up the code on master, not the code to be tested. This is a problem with Go packaging in general.

kataras:
Truly, nice job @allenluce, is that tested on windows operating system as well?

manishrjain:
Really nice change! Got a few comments. Once you address them, I'll pull this change in.

@allenluce : You can use the `-t` flag. The -t flag instructs get to also download the packages required to build
the tests for the specified packages.

---

Reviewed 11 of 11 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[db.go, line 261 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71L_7WPm4z9uZxl3Rf:-L71L_7WPm4z9uZxl3Rg:b-c3yoln) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/db.go#L261)):*
> ```Go
> 	}
> 
> 	if !opt.ReadOnly {
> ```

If multiple processes are using the same Badger directory, then none of them should write back anything. So, also ensure that value log isn't replayed on start.

This would mean that if the DB had crashed before, then a read-only open would appear to lose some data. We must warn users about that in the documentation.

Alternatively: If ReadOnly is set, then the number of memtables above L0 should be set to a very high value, to allow value log to be replayed and stored in memtables, but without the process getting blocked on them getting compacted. This might be a better option to avoid an apparent loss of data.

In any case, all these only take effect if the last run of read-write instance Badger had crashed, and not properly closed.

---

*[db.go, line 333 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71N9X8shtE3ff1KwPD:-L71N9X94SXjUSkVIaw1:bbk638m) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/db.go#L333)):*
> ```Go
> 
> 	// Now close the value log.
> 	if vlogErr := db.vlog.Close(db.opt.ReadOnly); err == nil {
> ```

Value log takes a copy of options on Open. So, it should already know about readonly flag and doesn't need to be passed explicitly.

https://github.com/dgraph-io/badger/blob/e8ce3e9d89507ec60f9b1f2f4c7539ae3efe61cd/db.go#L265

---

*[db_test.go, line 1573 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71OX6WxqIb4dp9uTJM:-L71OX6WxqIb4dp9uTJN:bsnb96o) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/db_test.go#L1573)):*
> ```Go
> 	require.Contains(t, err.Error(), "No sets or deletes are allowed in a read-only transaction")
> 	err = txn.Commit(nil)
> 	require.NoError(t, err)
> ```

LGTM Nice test.

---

*[manifest.go, line 263 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71Pj64J_ljoSQ2AKlI:-L71Pj64J_ljoSQ2AKlJ:b-s8law3) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/manifest.go#L263)):*
> ```Go
> 		return nil, 0, err
> 	}
> 	fp, err = y.OpenExistingSyncedFile(manifestPath, false, false)
> ```

Instead of 2 boolean params, how about using an int based flag. And do something like:

`y.OpenExistingFile(manifestPath, y.Sync | y.ReadOnly)`

---

*[options.go, line 90 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71Q6xgbTazkyDjqtg4:-L71Q6xgbTazkyDjqtg5:b-zdc6s8) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/options.go#L90)):*
> ```Go
> 	maxBatchSize  int64 // max batch size in bytes
> 
> 	// Whether the DB allows writes. With ReadOnly set, multiple Opens
> ```

// Open the DB as read-only. With this set, multiple processes can open the same Badger DB.

---

*[transaction.go, line 514 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71QHJnwpNMJabBbYAk:-L71QHJnwpNMJabBbYAl:blkzwwk) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/transaction.go#L514)):*
> ```Go
> //  // Call various APIs.
> func (db *DB) NewTransaction(update bool) *Txn {
> 	// Is the DB read-only?
> ```

Convert this question into a statement.

---

*[value.go, line 558 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71QXRtvC-7qkjzBIyy:-L71QXRtvC-7qkjzBIyz:b-6qjhy0) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/value.go#L558)):*
> ```Go
> 		if fid == maxFid {
> 			if lf.fd, err = y.OpenExistingSyncedFile(vlog.fpath(fid),
> 				vlog.opt.SyncWrites, readOnly); err != nil {
> ```

Same as above. This should be `y.OpenExistingFile(path, y.Sync|y.ReadOnly)`.

Note the function name has been changed to reflect that this function is no longer only about sync.

---

*[value.go, line 615 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71Qncg_bO9mnrbsX8J:-L71Qncg_bO9mnrbsX8K:b-vqz8cw) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/value.go#L615)):*
> ```Go
> }
> 
> func (vlog *valueLog) Close(readOnly bool) error {
> ```

This argument passing shouldn't be required because value log holds an opt as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/432#-:-L71QwgleJQ7R2knSFea:b-s946qc)*

allenluce:
I don't have easy access to a Windows build environment. I'd expect this to fail on Windows: the signature for `acquireDirectoryLock` in `dir_windows.go` is out of sync and the locking is exclusive only.

---

Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/432#-:-L71TVGQDbLUsd9xcyRC:b-bifj5l)*

allenluce:
What's the `-t` comment for?

manishrjain:
The -t comment is for:

> The checks fail because internal references to dgraph-io/badger pull up the code on master, not the code to be tested. This is a problem with Go packaging in general.

allenluce:
I was explaining the failed check for AppVeyor there.

allenluce:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[db.go, line 261 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71L_7WPm4z9uZxl3Rf:-L71UXPnPrUDyyydihCT:b-3jucyz) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/db.go#L261)):*

If multiple processes are using the same Badger directory, then none of them should write back anything. So, also ensure that value log isn't replayed on start.

This would mean that if the DB had crashed before, then a read-only open would appear to lose some data. We must warn users about that in the documentation.

Alternatively: If ReadOnly is set, then the number of memtables above L0 should be set to a very high value, to allow value log to be replayed and stored in memtables, but without the process getting blocked on them getting compacted. This might be a better option to avoid an apparent loss of data.

In any case, all these only take effect if the last run of read-write instance Badger had crashed, and not properly closed.
</blockquote></details>

I've taken the first approach somewhat modified: if the DB is opened read-only and there's data to play back, `Open()` returns an error saying the database had not been properly closed.

---

*[db.go, line 333 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71N9X8shtE3ff1KwPD:-L71d_Tq5vgwc7EU1E-b:bqzq25y) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/db.go#L333)):*

Value log takes a copy of options on Open. So, it should already know about readonly flag and doesn't need to be passed explicitly.

https://github.com/dgraph-io/badger/blob/e8ce3e9d89507ec60f9b1f2f4c7539ae3efe61cd/db.go#L265
</blockquote></details>

Cool, I've corrected.

---

*[db_test.go, line 1573 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71OX6WxqIb4dp9uTJM:-L71i4GaGHJ46JbCu-gs:b-ucfknh) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/db_test.go#L1573)):*

LGTM Nice test.
</blockquote></details>

Thanks!

---

*[manifest.go, line 263 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71Pj64J_ljoSQ2AKlI:-L71gA8Woa5ghXP2LEQw:blbeh2x) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/manifest.go#L263)):*

Instead of 2 boolean params, how about using an int based flag. And do something like:

`y.OpenExistingFile(manifestPath, y.Sync | y.ReadOnly)`
</blockquote></details>

OK, I switched to that. Not sure the way the function is currently used makes the flags any less clunky.

---

*[options.go, line 90 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71Q6xgbTazkyDjqtg4:-L71gurwDa14iVDTmL1r:b-a2ujfl) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/options.go#L90)):*

// Open the DB as read-only. With this set, multiple processes can open the same Badger DB.
</blockquote></details>

Changed.

---

*[transaction.go, line 514 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71QHJnwpNMJabBbYAk:-L71i5qQkTEbvKgzo9U6:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/transaction.go#L514)):*

Convert this question into a statement.
</blockquote></details>

Done.

---

*[value.go, line 558 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71QXRtvC-7qkjzBIyy:-L71i770Gz3n_g90HL8f:b-wz4n92) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/value.go#L558)):*

Same as above. This should be `y.OpenExistingFile(path, y.Sync|y.ReadOnly)`.

Note the function name has been changed to reflect that this function is no longer only about sync.
</blockquote></details>

Gottit. I've changed the function names.

---

*[value.go, line 615 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71Qncg_bO9mnrbsX8J:-L71iA7tNHnMS42xU5dy:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/value.go#L615)):*

This argument passing shouldn't be required because value log holds an opt as well.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/432)*

manishrjain:
Thanks for addressing the comments. One last round of follow up comments, and then we should be good to go.

---

Reviewed 10 of 10 files at r2.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[db.go, line 295 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71xZxAtZf2C4xY_cd0:-L71xZxAtZf2C4xY_cd1:be9ih68) ([raw file](https://github.com/dgraph-io/badger/blob/691e0c42d8a507da00957c80b1c5a7670ee2ea6d/db.go#L295)):*
> ```Go
> 	go db.doWrites(replayCloser)
> 
> 	if err = db.vlog.Replay(vptr, opt.ReadOnly, replayFunction(db)); err != nil {
> ```

You don't need to pass the opt here. vlog already has that opt.

---

*[db_test.go, line 1526 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71xktLZlLVffRkbRu6:-L71xktLZlLVffRkbRu7:b-mlrc6g) ([raw file](https://github.com/dgraph-io/badger/blob/691e0c42d8a507da00957c80b1c5a7670ee2ea6d/db_test.go#L1526)):*
> ```Go
> 	_, err = Open(opts)
> 	require.Error(t, err)
> 	if err.Error() == "read-only mode is not supported on Windows" {
> ```

Instead of checking the wordings, create an error variable (errors.go), and check equality of that variable.

---

*[dir_windows.go, line 62 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L72-XKQ9y-tQ7PHXIFr:-L72-XKQ9y-tQ7PHXIFs:b25h00y) ([raw file](https://github.com/dgraph-io/badger/blob/691e0c42d8a507da00957c80b1c5a7670ee2ea6d/dir_windows.go#L62)):*
> ```Go
> func acquireDirectoryLock(dirPath string, pidFileName string, readOnly bool) (*directoryLockGuard, error) {
> 	if readOnly {
> 		return nil, fmt.Errorf("read-only mode is not supported on Windows")
> ```

Use an error variable. You can define it in errors.go.

---

*[manifest.go, line 263 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L71Pj64J_ljoSQ2AKlI:-L72-miYosGZDL3vAJhJ:b-6mwgl1) ([raw file](https://github.com/dgraph-io/badger/blob/f5370179b1995d7e1df2f4092b6b5b10921983dc/manifest.go#L263)):*
<details><summary><i>Previously, allenluce (Allen Luce) wrote…</i></summary><blockquote>

OK, I switched to that. Not sure the way the function is currently used makes the flags any less clunky.
</blockquote></details>

Thanks. This is more in line with how Linux flags are passed.

---

*[options.go, line 93 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L72-xDLxw5YYMRYqL6L:-L72-xDLxw5YYMRYqL6M:b-m6j451) ([raw file](https://github.com/dgraph-io/badger/blob/691e0c42d8a507da00957c80b1c5a7670ee2ea6d/options.go#L93)):*
> ```Go
> 	// open the same Badger DB. Note: if the DB being opened had crashed
> 	// before and has vlog data to be replayed, ReadOnly will cause Open
> 	// to fail (with an appropriate message).
> ```

Remove the round brackets.

---

*[value.go, line 303 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L7204MXMpdbhQwWcykJ:-L7204MXMpdbhQwWcykK:b-loiz8f) ([raw file](https://github.com/dgraph-io/badger/blob/691e0c42d8a507da00957c80b1c5a7670ee2ea6d/value.go#L303)):*
> ```Go
> 
> 		if readOnly {
> 			return errors.Errorf("database was not properly closed, cannot open read-only")
> ```

Create an error variable in errors.go, and use that error.

---

*[value.go, line 672 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L720H7gCD78fjy6PiCF:-L720H7gCD78fjy6PiCG:boxdkga) ([raw file](https://github.com/dgraph-io/badger/blob/691e0c42d8a507da00957c80b1c5a7670ee2ea6d/value.go#L672)):*
> ```Go
> 
> // Replay replays the value log. The kv provided is only valid for the lifetime of function call.
> func (vlog *valueLog) Replay(ptr valuePointer, readOnly bool, fn logEntry) error {
> ```

readOnly doesn't need to be passed. vlog should already know this via opt.

---

*[value_test.go, line 533 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/432#-L720OSp5Q70sHdWp4fc:-L720OSp5Q70sHdWp4fd:b-klut8s) ([raw file](https://github.com/dgraph-io/badger/blob/691e0c42d8a507da00957c80b1c5a7670ee2ea6d/value_test.go#L533)):*
> ```Go
> 	kv, err = Open(opts)
> 	require.Error(t, err)
> 	require.Regexp(t, "database was not properly closed, cannot open read-only|read-only mode is not supported on Windows", err.Error())
> ```

You can check error variable equality instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/432#-:-L720WDBfuUfipigMqoe:b-xvxntz)*

Issue #374, opened 2017-12-22
Calling DB.Close() multiple times is not safe


luma:
Calling `Close()` multiple times panics during an attempt to close the `closed` channel in `Closer` multiple times.

It seems like either calling `Close` after the DB is closed should suceed, or the docs should mention that it's not safe so developers can guard against it.

Gory details:
```
     close of closed channel
      /usr/local/go/src/runtime/panic.go:491

      Full Stack Trace
      	/usr/local/go/src/runtime/panic.go:491 +0x283
      github.com/luma/profuse/vendor/github.com/dgraph-io/badger/y.(*Closer).Signal(...)
      	/home/rolly/projects/src/github.com/luma/profuse/vendor/github.com/dgraph-io/badger/y/y.go:173
      github.com/luma/profuse/vendor/github.com/dgraph-io/badger/y.(*Closer).SignalAndWait(0xc4204aa400)
      	/home/rolly/projects/src/github.com/luma/profuse/vendor/github.com/dgraph-io/badger/y/y.go:194 +0x2e
      github.com/luma/profuse/vendor/github.com/dgraph-io/badger.(*DB).Close(0xc4203f4900, 0xbea260, 0xce6c58)
      	/home/rolly/projects/src/github.com/luma/profuse/vendor/github.com/dgraph-io/badger/db.go:320 +0x7a
```

manishrjain:
I'm not convinced that we need to add more logic to catch this case. So, let's mention this in our documentation.

PR #437, opened 2018-03-10
Do not panic if database is closed

### LINKED ISSUE #420: Badger Writes Should Return an Error if DB is Closed
Currently, writes to a closed database result in a panic: 
```
panic: send on closed channel

goroutine 412 [running]:
github.com/capsule8/reactive8/vendor/github.com/dgraph-io/badger.(*DB).sendToWriteCh(0xc42036a600, 0xc420bed2b0, 0x2, 0x2, 0xc420293bb8, 0x10, 0x89

        /go/src/github.com/capsule8/reactive8/vendor/github.com/dgraph-io/badger/db.go:589 +0x168
github.com/capsule8/reactive8/vendor/github.com/dgraph-io/badger.(*DB).batchSet(0xc42036a600, 0xc420bed2b0, 0x2, 0x2, 0x4, 0x8)
        /go/src/github.com/capsule8/reactive8/vendor/github.com/dgraph-io/badger/db.go:659 +0x4d
github.com/capsule8/reactive8/vendor/github.com/dgraph-io/badger.(*Txn).Commit(0xc420be3440, 0x0, 0x0, 0x0)
        /go/src/github.com/capsule8/reactive8/vendor/github.com/dgraph-io/badger/transaction.go:500 +0x847
github.com/capsule8/reactive8/vendor/github.com/dgraph-io/badger.(*DB).Update(0xc42036a600, 0xc420293d70, 0x0, 0x0)
        /go/src/github.com/capsule8/reactive8/vendor/github.com/dgraph-io/badger/transaction.go:568 +0xd3
```

Ideally, writes to a closed DB should probably return an exported error named `ErrBadgerClosed` or something similar.

adelowo:
Fixes #420 and #374

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/437)

### DIFF
--- a/db.go
+++ b/db.go
@@ -322,7 +322,8 @@ func Open(opt Options) (db *DB, err error) {
 }
 
 // Close closes a DB. It's crucial to call it to ensure all the pending updates
-// make their way to disk.
+// make their way to disk. Calling DB.Close() multiple times is not safe and would
+// cause panic.
 func (db *DB) Close() (err error) {
 	if db.isClosed() {
 		return ErrBadgerClosed
@@ -382,6 +383,23 @@ func (db *DB) Close() (err error) {
 	if db.closers.compactors != nil {
 		db.closers.compactors.SignalAndWait()
 		db.elog.Printf("Compaction finished")
+	}
+
+	// Force Compact L0
+	// We don't need to care about cstatus since no parallel compaction is running.
+	cd := compactDef{
+		elog:      trace.New("Badger", "Compact"),
+		thisLevel: db.lc.levels[0],
+		nextLevel: db.lc.levels[1],
+	}
+	cd.elog.SetMaxEvents(100)
+	defer cd.elog.Finish()
+	if db.lc.fillTablesL0(&cd) {
+		if err := db.lc.runCompactDef(0, cd); err != nil {
+			cd.elog.LazyPrintf("\tLOG Compact FAILED with error: %+v: %+v", err, cd)
+		}
+	} else {
+		cd.elog.LazyPrintf("fillTables failed for level zero. No compaction required")
 	}
 
 	if lcErr := db.lc.close(); err == nil {
--- a/db.go
+++ b/db.go
@@ -32,6 +32,8 @@ import (
 
 	"golang.org/x/net/trace"
 
+	"sync/atomic"
+
 	"github.com/dgraph-io/badger/skl"
 	"github.com/dgraph-io/badger/table"
 	"github.com/dgraph-io/badger/y"
@@ -75,7 +77,7 @@ type DB struct {
 
 	orc *oracle
 
-	closed chan struct{}
+	closed uint32
 }
 
 const (
@@ -247,7 +249,7 @@ func Open(opt Options) (db *DB, err error) {
 		dirLockGuard:  dirLockGuard,
 		valueDirGuard: valueDirLockGuard,
 		orc:           orc,
-		closed:        make(chan struct{}, 1),
+		closed:        0,
 	}
 
 	// Calculate initial size.
@@ -336,7 +338,7 @@ func (db *DB) Close() (err error) {
 	// Stop writes next.
 	db.closers.writes.SignalAndWait()
 
-	db.closed <- struct{}{}
+	atomic.AddUint32(&db.closed, 1)
 
 	// Now close the value log.
 	if vlogErr := db.vlog.Close(); err == nil {
--- a/db.go
+++ b/db.go
@@ -74,6 +74,8 @@ type DB struct {
 	flushChan chan flushTask // For flushing memtables.
 
 	orc *oracle
+
+	closed chan struct{}
 }
 
 const (
@@ -245,6 +247,7 @@ func Open(opt Options) (db *DB, err error) {
 		dirLockGuard:  dirLockGuard,
 		valueDirGuard: valueDirLockGuard,
 		orc:           orc,
+		closed:        make(chan struct{}, 1),
 	}
 
 	// Calculate initial size.
@@ -322,12 +325,18 @@ func Open(opt Options) (db *DB, err error) {
 // Close closes a DB. It's crucial to call it to ensure all the pending updates
 // make their way to disk.
 func (db *DB) Close() (err error) {
+	if db.isClosed() {
+		return ErrBadgerClosed
+	}
+
 	db.elog.Printf("Closing database")
 	// Stop value GC first.
 	db.closers.valueGC.SignalAndWait()
 
 	// Stop writes next.
 	db.closers.writes.SignalAndWait()
+
+	db.closed <- struct{}{}
 
 	// Now close the value log.
 	if vlogErr := db.vlog.Close(); err == nil {
@@ -905,6 +914,10 @@ func (db *DB) updateSize(lc *y.Closer) {
 
 // PurgeVersionsBelow will delete all versions of a key below the specified version
 func (db *DB) PurgeVersionsBelow(key []byte, ts uint64) error {
+	if db.isClosed() {
+		return ErrBadgerClosed
+	}
+
 	txn := db.NewTransaction(false)
 	defer txn.Discard()
 	return db.purgeVersionsBelow(txn, key, ts)
--- a/db.go
+++ b/db.go
@@ -26,18 +26,15 @@ import (
 	"path/filepath"
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/dgraph-io/badger/options"
-
-	"golang.org/x/net/trace"
-
-	"sync/atomic"
-
 	"github.com/dgraph-io/badger/skl"
 	"github.com/dgraph-io/badger/table"
 	"github.com/dgraph-io/badger/y"
 	"github.com/pkg/errors"
+	"golang.org/x/net/trace"
 )
 
 var (
@@ -331,14 +328,14 @@ func (db *DB) Close() (err error) {
 		return ErrBadgerClosed
 	}
 
+	atomic.StoreUint32(&db.closed, 1)
+
 	db.elog.Printf("Closing database")
 	// Stop value GC first.
 	db.closers.valueGC.SignalAndWait()
 
 	// Stop writes next.
 	db.closers.writes.SignalAndWait()
-
-	atomic.AddUint32(&db.closed, 1)
 
 	// Now close the value log.
 	if vlogErr := db.vlog.Close(); err == nil {
--- a/db_test.go
+++ b/db_test.go
@@ -103,6 +103,48 @@ func TestWrite(t *testing.T) {
 	})
 }
 
+func TestDB_Close_MultipleTimes(t *testing.T) {
+	runBadgerTest(t, nil, func(t *testing.T, db *DB) {
+		db.Close()
+
+		err := db.Close()
+		require.Equal(t, ErrBadgerClosed, err)
+	})
+}
+
+func TestDB_Close(t *testing.T) {
+	dir, err := ioutil.TempDir("", "badger")
+	require.NoError(t, err)
+
+	db, err := Open(getTestOptions(dir))
+	require.NoError(t, err)
+
+	writeFn := func(txn *Txn) error {
+		return txn.Set([]byte("name"), []byte("Badger"))
+	}
+
+	err = db.Update(writeFn)
+	require.NoError(t, err)
+
+	err = db.View(func(txn *Txn) error {
+		_, err := txn.Get([]byte("name"))
+		return err
+	})
+	require.NoError(t, err)
+
+	db.Close()
+
+	err = db.View(func(txn *Txn) error {
+		_, err := txn.Get([]byte("name"))
+		return err
+	})
+	require.Error(t, err)
+
+	err = db.Update(writeFn)
+	require.Error(t, err)
+	require.Equal(t, ErrBadgerClosed, err)
+}
+
 func TestUpdateAndView(t *testing.T) {
 	runBadgerTest(t, nil, func(t *testing.T, db *DB) {
 		err := db.Update(func(txn *Txn) error {
--- a/db_test.go
+++ b/db_test.go
@@ -395,14 +395,14 @@ func TestGetMore(t *testing.T) {
 
 				vs, err := db.get(y.KeyWithTs(k, math.MaxUint64))
 				require.NoError(t, err)
-				fmt.Printf("wanted=%q Item: %s\n", k, item.ToString())
+				fmt.Printf("wanted=%q Item: %s\n", k, item)
 				fmt.Printf("on re-run, got version: %+v\n", vs)
 
 				txn := db.NewTransaction(false)
 				itr := txn.NewIterator(DefaultIteratorOptions)
 				for itr.Seek(k); itr.Valid(); itr.Next() {
 					item := itr.Item()
-					fmt.Printf("item=%s\n", item.ToString())
+					fmt.Printf("item=%s\n", item)
 					if !bytes.Equal(item.Key(), k) {
 						break
 					}
@@ -410,7 +410,7 @@ func TestGetMore(t *testing.T) {
 				itr.Close()
 				txn.Discard()
 			}
-			require.EqualValues(t, expectedValue, string(getItemValue(t, item)), "wanted=%q Item: %s\n", k, item.ToString())
+			require.EqualValues(t, expectedValue, string(getItemValue(t, item)), "wanted=%q Item: %s\n", k, item)
 			txn.Discard()
 		}
 
--- a/db_test.go
+++ b/db_test.go
@@ -145,6 +145,10 @@ func TestDB_Close(t *testing.T) {
 	require.Equal(t, ErrBadgerClosed, err)
 }
 
+func TestTxn_ClosedDB(t *testing.T) {
+
+}
+
 func TestUpdateAndView(t *testing.T) {
 	runBadgerTest(t, nil, func(t *testing.T, db *DB) {
 		err := db.Update(func(txn *Txn) error {
--- a/db_test.go
+++ b/db_test.go
@@ -145,10 +145,6 @@ func TestDB_Close(t *testing.T) {
 	require.Equal(t, ErrBadgerClosed, err)
 }
 
-func TestTxn_ClosedDB(t *testing.T) {
-
-}
-
 func TestUpdateAndView(t *testing.T) {
 	runBadgerTest(t, nil, func(t *testing.T, db *DB) {
 		err := db.Update(func(txn *Txn) error {
--- a/db_test.go
+++ b/db_test.go
@@ -324,6 +324,38 @@ func TestTxnTooBig(t *testing.T) {
 	})
 }
 
+func TestForceCompactL0(t *testing.T) {
+	dir, err := ioutil.TempDir("", "badger")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	opts := getTestOptions(dir)
+	opts.ValueLogFileSize = 15 << 20
+	db, err := OpenManaged(opts)
+	require.NoError(t, err)
+
+	data := func(i int) []byte {
+		return []byte(fmt.Sprintf("%b", i))
+	}
+	n := 80
+	m := 45 // Increasing would cause ErrTxnTooBig
+	sz := 32 << 10
+	v := make([]byte, sz)
+	for i := 0; i < n; i += 2 {
+		version := uint64(i)
+		txn := db.NewTransactionAt(version, true)
+		for j := 0; j < m; j++ {
+			require.NoError(t, txn.Set(data(j), v))
+		}
+		require.NoError(t, txn.CommitAt(version+1, nil))
+	}
+	db.Close()
+
+	db, err = OpenManaged(opts)
+	defer db.Close()
+	require.Equal(t, len(db.lc.levels[0].tables), 0)
+}
+
 func TestGetAfterPurge(t *testing.T) {
 	dir, err := ioutil.TempDir("", "badger")
 	require.NoError(t, err)
--- a/errors.go
+++ b/errors.go
@@ -24,7 +24,7 @@ import (
 
 var (
 
-	// ErrBadgerClosed is returned when trying to write to a closed database
+	// ErrBadgerClosed is returned when trying to perform an operation to a closed database.
 	ErrBadgerClosed = errors.New("Database is already closed")
 
 	// ErrValueLogSize is returned when opt.ValueLogFileSize option is not within the valid
--- a/errors.go
+++ b/errors.go
@@ -23,6 +23,10 @@ import (
 )
 
 var (
+
+	// ErrBadgerClosed is returned when trying to write to a closed database
+	ErrBadgerClosed = errors.New("Database is already closed")
+
 	// ErrValueLogSize is returned when opt.ValueLogFileSize option is not within the valid
 	// range.
 	ErrValueLogSize = errors.New("Invalid ValueLogFileSize, must be between 1MB and 2GB")
--- a/iterator.go
+++ b/iterator.go
@@ -67,9 +67,16 @@ func (item *Item) ToString() string {
 // Key returns the key.
 //
 // Key is only valid as long as item is valid, or transaction is valid.  If you need to use it
-// outside its validity, please copy it.
+// outside its validity, please use KeyCopy
 func (item *Item) Key() []byte {
 	return item.key
+}
+
+// KeyCopy returns a copy of the key of the item, writing it to dst slice.
+// If nil is passed, or capacity of dst isn't sufficient, a new slice would be allocated and
+// returned.
+func (item *Item) KeyCopy(dst []byte) []byte {
+	return y.SafeCopy(dst, item.key)
 }
 
 // Version returns the commit timestamp of the item.
@@ -121,6 +128,11 @@ func (item *Item) hasValue() bool {
 		return false
 	}
 	return true
+}
+
+// IsDeletedOrExpired returns true if item contains deleted or expired value.
+func (item *Item) IsDeletedOrExpired() bool {
+	return isDeletedOrExpired(item.meta, item.expiresAt)
 }
 
 func (item *Item) yieldItemValue() ([]byte, func(), error) {
--- a/iterator.go
+++ b/iterator.go
@@ -53,10 +53,15 @@ type Item struct {
 	txn       *Txn
 }
 
+// String returns a string representation of Item
+func (item *Item) String() string {
+	return fmt.Sprintf("key=%q, version=%d, meta=%x", item.Key(), item.Version(), item.meta)
+}
+
+// Deprecated
 // ToString returns a string representation of Item
 func (item *Item) ToString() string {
-	return fmt.Sprintf("key=%q, version=%d, meta=%x", item.Key(), item.Version(), item.meta)
-
+	return item.String()
 }
 
 // Key returns the key.
--- a/transaction.go
+++ b/transaction.go
@@ -284,6 +284,8 @@ func (txn *Txn) SetWithTTL(key, val []byte, dur time.Duration) error {
 // with other metadata to the database.
 func (txn *Txn) SetEntry(e *Entry) error {
 	switch {
+	case txn.db.isClosed():
+		return ErrBadgerClosed
 	case !txn.update:
 		return ErrReadOnlyTxn
 	case txn.discarded:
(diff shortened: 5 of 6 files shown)


manishrjain:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[db.go, line 78 at r1](https://reviewable.io:443/reviews/dgraph-io/badger/437#-L7SS7Nv7-3gp8REeejM:-L7SS7NwsaPKVFAXRhEP:bntenvs) ([raw file](https://github.com/dgraph-io/badger/blob/a6d9c884d6cc6ee7b61f898341cdd7f9fef6ebf6/db.go#L78)):*
> ```Go
> 	orc *oracle
> 
> 	closed chan struct{}
> ```

Instead of using a chan as a flag, use atomics.

https://golang.org/pkg/sync/atomic/#StoreInt32

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/437)*

kataras:
Yes... @adelowo why you choosed a channel instead of atomic uint32 for example, is a theory behind it that we should care about?

adelowo:
I would update it to make use of atomics

adelowo:
Can you take a look at this now @manishrjain

manishrjain:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[db.go, line 341 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/437#-L7SbP_LsHd_vWiRvoA9:-L7SbP_LsHd_vWiRvoAA:b-9qi2yd) ([raw file](https://github.com/dgraph-io/badger/blob/77a0dfb05eda76a82231d5fdada2c5c8156fd543/db.go#L341)):*
> ```Go
> 	db.closers.writes.SignalAndWait()
> 
> 	atomic.AddUint32(&db.closed, 1)
> ```

Use Store instead of add. Also, move this right after checking `isClosed`. So, all the waiting on signals, etc. happens after. In particular, the writes.SignalAndWait is the place where most people get that writeCh panic.

https://github.com/dgraph-io/badger/blob/8e23f14e8d0d95756ae04380b3201ded0ea61f0b/db.go#L669

---

*[errors.go, line 27 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/437#-L7ScGTqYPBxNZAsMYXx:-L7ScGTqYPBxNZAsMYXy:b-ulafwy) ([raw file](https://github.com/dgraph-io/badger/blob/77a0dfb05eda76a82231d5fdada2c5c8156fd543/errors.go#L27)):*
> ```Go
> var (
> 
> 	// ErrBadgerClosed is returned when trying to write to a closed database
> ```

full stop at end.

---

*[transaction.go, line 551 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/437#-L7ScO0jwy0qoIGpyAWC:-L7ScO0jwy0qoIGpyAWD:bw3fjcb) ([raw file](https://github.com/dgraph-io/badger/blob/77a0dfb05eda76a82231d5fdada2c5c8156fd543/transaction.go#L551)):*
> ```Go
> // Update executes a function, creating and managing a read-write transaction
> // for the user. Error returned by the function is relayed by the Update method.
> func (db *DB) Update(fn func(txn *Txn) error) error {
> ```

Instead of doing this over here and in View, do this within NewTransaction.

We'd still have to deal with the case where DB is closed while a pending transaction is still live. That txn would try to read and write, and still cause panics.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/437)*

adelowo:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[transaction.go, line 551 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/437#-L7ScO0jwy0qoIGpyAWC:-L7TbdRegJuiQSIjJCwP:b-x52rsx) ([raw file](https://github.com/dgraph-io/badger/blob/77a0dfb05eda76a82231d5fdada2c5c8156fd543/transaction.go#L551)):*

Instead of doing this over here and in View, do this within NewTransaction.

We'd still have to deal with the case where DB is closed while a pending transaction is still live. That txn would try to read and write, and still cause panics.
</blockquote></details>

`NewTransaction` does not return an error.. If i have to do add the `isClosed` check to it, the public api would change as the error must obviously be returned

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/437)*

adelowo:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[errors.go, line 27 at r2](https://reviewable.io:443/reviews/dgraph-io/badger/437#-L7ScGTqYPBxNZAsMYXx:-L7TgIhac3wwNxsLPygr:bvmsmrm) ([raw file](https://github.com/dgraph-io/badger/blob/77a0dfb05eda76a82231d5fdada2c5c8156fd543/errors.go#L27)):*

full stop at end.
</blockquote></details>

Done... updated the comment anyways

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/437)*

adelowo:
---

<details><summary><i>The tests passes on my macbook funny enough</i></summary><blockquote>

backup1 length: 17384
backup2 length: 34954
Deleting i=0
Testing i=0
Testing i=0
Testing i=1000
Testing i=2000
Testing i=3000
Testing i=4000
Testing i=5000
Testing i=6000
Testing i=7000
Testing i=8000
Testing i=9000
Deleting i=0
Deleting i=9000
Testing i=0
Done and closing
Writing to dir /var/folders/_s/2s83cpt1051679hznx1943bw0000gn/T/badger352731551
Putting i=0
Testing i=0
FileIDs: [12 13 26 27 28 29 30 31 32 33 34]
/var/folders/_s/2s83cpt1051679hznx1943bw0000gn/T/badger947432848
/var/folders/_s/2s83cpt1051679hznx1943bw0000gn/T/badger583476143
01500
Writes done. Iteration and updates starting...
Putting i=0
PASS
ok  	github.com/dgraph-io/badger	83.765s
</blockquote></details>
---

manishrjain:
:lgtm_strong: This all looks good. My only concern that's left is the impact on read/write performance, as every Get now has to load an atomic var before starting. Any thoughts on that?

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/437#-:-L7WWJNtOIeGDmCA2mwF:b-cinwzy)*

manishrjain:
In general, some sort of before and after read benchmarking would be reassuring.

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/437#-:-L7WWeDBr0EOPboprBqI:bgxc5sn)*

adelowo:
I would get those in place

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/badger/437#-:-L7Xx86_w_M_v5rT5PB2:bu7fs34)*

adelowo:
I ran some benchmark tests and ___there isn't a noticeable difference___ 

[benchmark-patch.txt](https://github.com/dgraph-io/badger/files/1810768/benchmark.txt)

> Running against master ( `go test -run=YYY -bench=Oops`)

```sh
goos: darwin
goarch: amd64
pkg: github.com/dgraph-io/badger
BenchmarkDB_Oops-8   	   10000	    178647 ns/op	    9044 B/op	      59 allocs/op
PASS
ok  	github.com/dgraph-io/badger	10.982s

```

> Against this PR

```sh

goos: darwin
goarch: amd64
pkg: github.com/dgraph-io/badger
BenchmarkDB_Oops-8   	   10000	    205644 ns/op	    9042 B/op	      59 allocs/op
PASS
ok  	github.com/dgraph-io/badger	12.341s

```

I have also used pprof to measure (`go test -run=YYY -bench=Oops -cpuprofile=oops.out`.. I ran list `Update`, here is what I got 

```sh
(pprof) list Update
Total: 560ms
ROUTINE ======================== github.com/dgraph-io/badger.(*DB).Update in /Users/lanreadelowo/go/src/github.com/dgraph-io/badger/transaction.go
         0       70ms (flat, cum) 12.50% of Total
         .          .    564:
         .          .    565:	if err := fn(txn); err != nil {
         .          .    566:		return err
         .          .    567:	}
         .          .    568:
         .       70ms    569:	return txn.Commit(nil)
         .          .    570:}
         .          .    571:
         .          .    572:func (db *DB) isClosed() bool {
         .          .    573:	return atomic.LoadUint32(&db.closed) > 0
         .          .    574:}
(pprof) list Commit
Total: 560ms
ROUTINE ======================== github.com/dgraph-io/badger.(*Txn).Commit in /Users/lanreadelowo/go/src/github.com/dgraph-io/badger/transaction.go
         0       70ms (flat, cum) 12.50% of Total
         .          .    464:		Value: []byte(strconv.FormatUint(commitTs, 10)),
         .          .    465:		meta:  bitFinTxn,
         .          .    466:	}
         .          .    467:	entries = append(entries, e)
         .          .    468:
         .       70ms    469:	req, err := txn.db.sendToWriteCh(entries)
         .          .    470:	state.writeLock.Unlock()
         .          .    471:	if err != nil {
         .          .    472:		return err
         .          .    473:	}
         .          .    474:
(pprof)

```

manishrjain:
That looks like a 15% degradation, which is significant.

Can you try with reads as well? I think there should be more -ve impact on the reads than the writes.

Overall, what I'm trying to determine is, whether this functionality is really worth the performance degradation that it brings. Or, is this something we should just handle by adding documentation?

manishrjain:
Can you add the profile for reads as well? Also include topN.

adelowo:
I would do just that in a bit and update this comment soon

adelowo:
Hi @manishrjain , nothing points to atomic. It would be nice if someone helps verify this.

pprof for `Read`s


```sh
(pprof) top20 -cum
Showing nodes accounting for 1.21s, 34.28% of 3.53s total
Dropped 78 nodes (cum <= 0.02s)
Showing top 20 nodes out of 112
      flat  flat%   sum%        cum   cum%
         0     0%     0%      2.82s 79.89%  testing.(*B).launch
         0     0%     0%      2.82s 79.89%  testing.(*B).runN
         0     0%     0%      2.81s 79.60%  github.com/dgraph-io/badger.BenchmarkDB_Damn
     0.01s  0.28%  0.28%      2.75s 77.90%  github.com/dgraph-io/badger.(*DB).Update
     0.03s  0.85%  1.13%      1.85s 52.41%  github.com/dgraph-io/badger.BenchmarkDB_Damn.func3
     0.02s  0.57%  1.70%      1.70s 48.16%  github.com/dgraph-io/badger.(*Txn).Get
     0.30s  8.50% 10.20%      1.62s 45.89%  runtime.mallocgc
     0.06s  1.70% 11.90%      1.09s 30.88%  github.com/dgraph-io/badger.(*DB).get
     0.02s  0.57% 12.46%      1.05s 29.75%  runtime.newobject
         0     0% 12.46%      1.02s 28.90%  runtime.systemstack
         0     0% 12.46%      0.74s 20.96%  runtime.gcStart
         0     0% 12.46%      0.73s 20.68%  runtime.gcStart.func2
     0.73s 20.68% 33.14%      0.73s 20.68%  runtime.kevent
         0     0% 33.14%      0.73s 20.68%  runtime.netpoll
         0     0% 33.14%      0.73s 20.68%  runtime.startTheWorldWithSema
         0     0% 33.14%      0.62s 17.56%  runtime.schedule
         0     0% 33.14%      0.57s 16.15%  runtime.mcall
         0     0% 33.14%      0.57s 16.15%  runtime.park_m
         0     0% 33.14%      0.56s 15.86%  github.com/dgraph-io/badger.(*DB).NewTransaction (inline)
     0.04s  1.13% 34.28%      0.54s 15.30%  github.com/dgraph-io/badger/skl.(*Skiplist).Get
(pprof) top20
Showing nodes accounting for 2.74s, 77.62% of 3.53s total
Dropped 78 nodes (cum <= 0.02s)
Showing top 20 nodes out of 112
      flat  flat%   sum%        cum   cum%
     0.73s 20.68% 20.68%      0.73s 20.68%  runtime.kevent
     0.43s 12.18% 32.86%      0.43s 12.18%  runtime.mach_semaphore_wait
     0.30s  8.50% 41.36%      1.62s 45.89%  runtime.mallocgc
     0.18s  5.10% 46.46%      0.20s  5.67%  runtime.heapBitsSetType
     0.15s  4.25% 50.71%      0.15s  4.25%  runtime.memclrNoHeapPointers
     0.14s  3.97% 54.67%      0.14s  3.97%  runtime.mach_semaphore_signal
     0.12s  3.40% 58.07%      0.21s  5.95%  runtime.deferreturn
     0.08s  2.27% 60.34%      0.38s 10.76%  github.com/dgraph-io/badger/y.ParseKey
     0.08s  2.27% 62.61%      0.08s  2.27%  runtime.nextFreeFast (inline)
     0.06s  1.70% 64.31%      1.09s 30.88%  github.com/dgraph-io/badger.(*DB).get
     0.06s  1.70% 66.01%      0.36s 10.20%  github.com/dgraph-io/badger.(*DB).getMemTables
     0.06s  1.70% 67.71%      0.06s  1.70%  runtime.acquirem (inline)
     0.06s  1.70% 69.41%      0.12s  3.40%  runtime.deferproc
     0.05s  1.42% 70.82%      0.07s  1.98%  runtime.freedefer
     0.05s  1.42% 72.24%      0.05s  1.42%  syscall.Syscall
     0.04s  1.13% 73.37%      0.19s  5.38%  github.com/dgraph-io/badger.(*Txn).Commit
     0.04s  1.13% 74.50%      0.54s 15.30%  github.com/dgraph-io/badger/skl.(*Skiplist).Get
     0.04s  1.13% 75.64%      0.05s  1.42%  runtime.newdefer
     0.04s  1.13% 76.77%      0.05s  1.42%  runtime.scanobject
     0.03s  0.85% 77.62%      0.07s  1.98%  github.com/dgraph-io/badger.(*oracle).decrRef
(pprof) top50
Showing nodes accounting for 3.35s, 94.90% of 3.53s total
Dropped 78 nodes (cum <= 0.02s)
Showing top 50 nodes out of 112
      flat  flat%   sum%        cum   cum%
     0.73s 20.68% 20.68%      0.73s 20.68%  runtime.kevent
     0.43s 12.18% 32.86%      0.43s 12.18%  runtime.mach_semaphore_wait
     0.30s  8.50% 41.36%      1.62s 45.89%  runtime.mallocgc
     0.18s  5.10% 46.46%      0.20s  5.67%  runtime.heapBitsSetType
     0.15s  4.25% 50.71%      0.15s  4.25%  runtime.memclrNoHeapPointers
     0.14s  3.97% 54.67%      0.14s  3.97%  runtime.mach_semaphore_signal
     0.12s  3.40% 58.07%      0.21s  5.95%  runtime.deferreturn
     0.08s  2.27% 60.34%      0.38s 10.76%  github.com/dgraph-io/badger/y.ParseKey
     0.08s  2.27% 62.61%      0.08s  2.27%  runtime.nextFreeFast (inline)
     0.06s  1.70% 64.31%      1.09s 30.88%  github.com/dgraph-io/badger.(*DB).get
     0.06s  1.70% 66.01%      0.36s 10.20%  github.com/dgraph-io/badger.(*DB).getMemTables
     0.06s  1.70% 67.71%      0.06s  1.70%  runtime.acquirem (inline)
     0.06s  1.70% 69.41%      0.12s  3.40%  runtime.deferproc
     0.05s  1.42% 70.82%      0.07s  1.98%  runtime.freedefer
     0.05s  1.42% 72.24%      0.05s  1.42%  syscall.Syscall
     0.04s  1.13% 73.37%      0.19s  5.38%  github.com/dgraph-io/badger.(*Txn).Commit
     0.04s  1.13% 74.50%      0.54s 15.30%  github.com/dgraph-io/badger/skl.(*Skiplist).Get
     0.04s  1.13% 75.64%      0.05s  1.42%  runtime.newdefer
     0.04s  1.13% 76.77%      0.05s  1.42%  runtime.scanobject
     0.03s  0.85% 77.62%      0.07s  1.98%  github.com/dgraph-io/badger.(*oracle).decrRef
     0.03s  0.85% 78.47%      1.85s 52.41%  github.com/dgraph-io/badger.BenchmarkDB_Damn.func3
     0.03s  0.85% 79.32%      0.03s  0.85%  github.com/dgraph-io/badger/skl.(*Skiplist).DecrRef (inline)
     0.03s  0.85% 80.17%      0.03s  0.85%  github.com/dgraph-io/badger/y.(*ValueStruct).Decode
     0.03s  0.85% 81.02%      0.05s  1.42%  github.com/dgraph-io/badger/y.CompareKeys
     0.03s  0.85% 81.87%      0.30s  8.50%  runtime.convT2Eslice
     0.03s  0.85% 82.72%      0.03s  0.85%  runtime.jmpdefer
     0.03s  0.85% 83.57%      0.03s  0.85%  runtime.mach_semaphore_timedwait
     0.03s  0.85% 84.42%      0.03s  0.85%  runtime.usleep
     0.03s  0.85% 85.27%      0.04s  1.13%  runtime.wbBufFlush1
     0.02s  0.57% 85.84%      1.70s 48.16%  github.com/dgraph-io/badger.(*Txn).Get
     0.02s  0.57% 86.40%      0.02s  0.57%  github.com/dgraph-io/badger/y.ParseTs (inline)
     0.02s  0.57% 86.97%      0.40s 11.33%  github.com/dgraph-io/badger/y.SameKey
     0.02s  0.57% 87.54%      0.03s  0.85%  github.com/dgryski/go-farm.Fingerprint64
     0.02s  0.57% 88.10%      0.02s  0.57%  runtime.cmpbody
     0.02s  0.57% 88.67%      0.08s  2.27%  runtime.growslice
     0.02s  0.57% 89.24%      0.02s  0.57%  runtime.heapBitsForAddr (inline)
     0.02s  0.57% 89.80%      0.02s  0.57%  runtime.heapBitsForObject
     0.02s  0.57% 90.37%      1.05s 29.75%  runtime.newobject
     0.02s  0.57% 90.93%      0.11s  3.12%  runtime.rawbyteslice
     0.02s  0.57% 91.50%      0.02s  0.57%  runtime.releasem (inline)
     0.02s  0.57% 92.07%      0.02s  0.57%  sync.(*Mutex).Lock
     0.02s  0.57% 92.63%      0.02s  0.57%  sync.(*Mutex).Unlock
     0.02s  0.57% 93.20%      0.02s  0.57%  sync.(*RWMutex).RLock
     0.01s  0.28% 93.48%      2.75s 77.90%  github.com/dgraph-io/badger.(*DB).Update
     0.01s  0.28% 93.77%      0.08s  2.27%  github.com/dgraph-io/badger.(*Txn).Discard
     0.01s  0.28% 94.05%      0.10s  2.83%  github.com/dgraph-io/badger/y.KeyWithTs
     0.01s  0.28% 94.33%      0.05s  1.42%  runtime.gcDrain
     0.01s  0.28% 94.62%      0.02s  0.57%  runtime.lock
     0.01s  0.28% 94.90%      0.20s  5.67%  runtime.makemap_small
         0     0% 94.90%      0.02s  0.57%  github.com/dgraph-io/badger.(*DB).Close
(pprof) list Get
Total: 3.53s
ROUTINE ======================== github.com/dgraph-io/badger.(*Txn).Get in /Users/lanreadelowo/go/src/github.com/dgraph-io/badger/transaction.go
      20ms      1.70s (flat, cum) 48.16% of Total
         .          .    331:           return nil, ErrEmptyKey
         .          .    332:   } else if txn.discarded {
         .          .    333:           return nil, ErrDiscardedTxn
         .          .    334:   }
         .          .    335:
      10ms      380ms    336:   item = new(Item)
         .          .    337:   if txn.update {
         .       10ms    338:           if e, has := txn.pendingWrites[string(key)]; has && bytes.Equal(key, e.Key) {
         .          .    339:                   if isDeletedOrExpired(e.meta, e.ExpiresAt) {
         .          .    340:                           return nil, ErrKeyNotFound
         .          .    341:                   }
         .          .    342:                   // Fulfill from cache.
         .          .    343:                   item.meta = e.meta
         .          .    344:                   item.val = e.Value
         .          .    345:                   item.userMeta = e.UserMeta
         .          .    346:                   item.key = key
         .          .    347:                   item.status = prefetched
         .          .    348:                   item.version = txn.readTs
         .          .    349:                   // We probably don't need to set db on item here.
         .          .    350:                   return item, nil
         .          .    351:           }
         .          .    352:           // Only track reads if this is update txn. No need to track read if txn serviced it
         .          .    353:           // internally.
         .       30ms    354:           fp := farm.Fingerprint64(key)
      10ms       90ms    355:           txn.reads = append(txn.reads, fp)
         .          .    356:   }
         .          .    357:
         .      100ms    358:   seek := y.KeyWithTs(key, txn.readTs)
         .      1.09s    359:   vs, err := txn.db.get(seek)
         .          .    360:   if err != nil {
         .          .    361:           return nil, errors.Wrapf(err, "DB::Get key: %q", key)
         .          .    362:   }
         .          .    363:   if vs.Value == nil && vs.Meta == 0 {
         .          .    364:           return nil, ErrKeyNotFound
ROUTINE ======================== github.com/dgraph-io/badger/skl.(*Skiplist).Get in /Users/lanreadelowo/go/src/github.com/dgraph-io/badger/skl/skl.go
      40ms      540ms (flat, cum) 15.30% of Total
         .          .    379:}
         .          .    380:
         .          .    381:// Get gets the value associated with the key. It returns a valid value if it finds equal or earlier
         .          .    382:// version of the same key.
         .          .    383:func (s *Skiplist) Get(key []byte) y.ValueStruct {
      10ms       70ms    384:   n, _ := s.findNear(key, false, true) // findGreaterOrEqual.
         .          .    385:   if n == nil {
         .          .    386:           return y.ValueStruct{}
         .          .    387:   }
         .          .    388:
         .          .    389:   nextKey := s.arena.getKey(n.keyOffset, n.keySize)
      20ms      420ms    390:   if !y.SameKey(key, nextKey) {
         .          .    391:           return y.ValueStruct{}
         .          .    392:   }
         .          .    393:
         .          .    394:   valOffset, valSize := n.getValueOffset()
      10ms       40ms    395:   vs := s.arena.getVal(valOffset, valSize)
         .       10ms    396:   vs.Version = y.ParseTs(nextKey)
         .          .    397:   return vs
         .          .    398:}
         .          .    399:
         .          .    400:// NewIterator returns a skiplist iterator.  You have to Close() the iterator.
         .          .    401:func (s *Skiplist) NewIterator() *Iterator {
(pprof) 

```

manishrjain:
Thanks for this. But, the code which is calling the new `isClosed` function isn't being shown in the list. Can you paste the full `Get` function?

```
(pprof) list Get
Total: 3.53s
ROUTINE ======================== github.com/dgraph-io/badger.(*Txn).Get in /Users/lanreadelowo/go/src/github.com/dgraph-io/badger/transaction.go
      20ms      1.70s (flat, cum) 48.16% of Total
         .          .    331:           return nil, ErrEmptyKey
         .          .    332:   } else if txn.discarded {
         .          .    333:           return nil, ErrDiscardedTxn
         .          .    334:   }
         .          .    335:
      10ms      380ms    336:   item = new(Item)
         .          .    337:   if txn.update {
```

adelowo:
That is all that is being shown... Not too familiar with pprof, can you help verify that ? Anyways, I generated an image of the entire profile 
profile001

whyrusleeping:
Hey guys, anything the ipfs team can do to help move this one along?

manishrjain:
I want to get a sense of how much slowness this change brings to concurrent reads. Maybe if your team (@whyrusleeping) can help benchmark that, that'd be good. The potential performance hit of this approach is the main concern here.

whyrusleeping:
@schomatis you game for some benchmarking?

schomatis:
Sure.

schomatis:
@manishrjain Before going through with the benchmark tests let me make some remarks that I would appreciate your feedback on:

1. These checks seem more a responsibility of the user's code than Badger's API. Avoiding `panic` is not a bad thing but it will hide more serious underlying problems, multiple closes or closing an ongoing transaction are likely symptoms of a logic error from the consumer of the API than a problem with Badger itself. (In my opinion documentation can be an alternative solution here.)

2. As you already [commented](https://reviewable.io/reviews/dgraph-io/badger/437#-L7ScO0jwy0qoIGpyAWC), the biggest issue here is what happens when closing the database with an ongoing operation on it (which is the problem in the IPFS issue referenced here), is that being handled in this PR?

3. Atomic operations like load or store used here do not provide synchronization per se, they just guarantee no data races over the address itself. Calling `Close` from different goroutines would still trigger issues like #374 (similarly if calling `Close` and other transaction operation simultaneously). I think channels or locks will be necessary if synchronization is the goal here, which will probably imply some performance degradation and I'm more than willing to benchmark that if you want.

manishrjain:
Hey @schomatis ,

1. Totally agree about this. This issue falls in the user domain and can be fixed with documentation.
2. The PR now does check for that, by checking for closed in every read or update call within a transaction.
3. Issue #374 can be fixed with this PR because it checks if DB is already closed before closing again.

schomatis:
> The PR now does check for that, by checking for closed in every read or update call within a transaction.

I was referring (and maybe you weren't and I misunderstood the review comment) to the case were an operation like `get` is already on its way (it already passed the initial `isClosed` check) when another goroutine starts the `Close` operation. This is the case of the IPFS issue referenced here and from what I'm testing locally this PR doesn't handle that case (if it should or not is another discussion, see below).

> Issue #374 can be fixed with this PR because it checks if DB is already closed before closing again.

I've meant this is not thread-safe. Maybe it's not meant to be, but if it is, the atomic operations do not guarantee that. Two different goroutines can call `Close`, pass the initial `isClosed` together, and then later both write atomically to `closed` without errors, but only one of them will be able to close the channel and the other will panic. The atomic load/stores only guarantee that the value won't be messed up (data race) if two goroutines read/write to it simultaneously and this goes toward thread-safety but it's not enough, what should be atomic is the read-then-write operation over the `closed` flag, e.g., with a RW mutex (the `Close` would try to hold the write lock and any other operation would contend only for the read lock). As mentioned before any kind of synchronization like a RW mutex *will* have a performance impact and that is something worth benchmarking.

In case thread-safety wasn't the goal the atomic operations can be discarded, they have a performance penalty without achieving its goal.

---------------------------------------

From my perspective, the central issue is that clearly certain operations cannot run in parallel in the DB (e.g., `Get` and `Close`, or `Close` and `Close`) because they manipulate the same control structures (e.g., the channel being closed that causes #374 or the `mt` pointer  that causes ipfs/go-ipfs#4820), they need to be synchronized so that one waits on the other to finish (`Close` in part does that with its `closers` channels), *should that synchronization be enforced in the API for all those operations?*

That's what (in my opinion) is worth evaluating. As you correctly worry, going down this road will have (not negligible) performance implications. Some users would appreciate it, as they won't have to check for that since it will be handled by Badger (avoiding some unpleasant panics), other users won't, as the performance degradation may be undesirable (and maybe they already have an underlying synchronization mechanism in the code that consumes the API and Badger's own synchronization mechanism would seem redundant).

I think this PR goes in the right direction but falls short in defining what is Badger willing to do and what is left to the user (and that should be documented in the API.)

manishrjain:
I agree with your assessment -- if we try to solve this problem with more code, we'd need to consider all the various edge cases that could happen to do it right. All of these would then add more complexity and come with some performance impact.

I think this issue is purely in the domain of having clear documentation instead. We clearly tell users to not access the DB after closing it and point out in FAQ the exact error they might receive if that happens. That'd be the right division between what Badger can do and what the user should do at their end.

If there are no objections, my take is to close this PR and fix documentation instead.

whyrusleeping:
One thing that might help, and have minimal impact of performance, would be to have an atomic reference counter for operations that may panic if the db is closed. Then we could at least check if things may panic before calling close (and use that to wait for cleanup/finishup on the user side)

whyrusleeping:
Then again, that will probably have the same perf impact as this PR

schomatis:
@manishrjain I agree that this PR could be closed and in its place another issue could be opened to fully discuss how to handle this.

Merging this PR *as is* will solve some of the problems mentioned here and its cost will probably be a very low performance impact but it has the danger of giving the impression that Badger's API is thread-safe and that might confuse the design of use cases for the consumers of the API.

manishrjain:
Agreed. Thanks for the detailed write-up and thorough analysis of the issue, @schomatis.

Thanks, @adelowo for the PR and for following through with all the reviews requested. Unfortunately, can't accept this PR, but there are many other issues in Badger which I could use your help with. Thanks again!

adelowo:
Thanks for your input(s) @schomatis . I agree with your statement on thread safety..

PR #445, opened 2018-03-26
Check specifically for `errNoRoom` when calling `ensureRoomForWrite`

schomatis:
Other errors should be handled outside the `for` loop (in the next [`if`](https://github.com/dgraph-io/badger/compare/master...schomatis:fix/write/errnoroom-check?expand=1#diff-7a117cbfd5485ed35efcbb2995c37aa0L606) which now appears to be dead code).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/445)

### DIFF
--- a/db.go
+++ b/db.go
@@ -593,7 +593,7 @@ func (db *DB) writeRequests(reqs []*request) error {
 		}
 		count += len(b.Entries)
 		var i uint64
-		for err := db.ensureRoomForWrite(); err != nil; err = db.ensureRoomForWrite() {
+		for err := db.ensureRoomForWrite(); err == errNoRoom; err = db.ensureRoomForWrite() {
 			i++
 			if i%100 == 0 {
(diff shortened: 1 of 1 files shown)


schomatis:
There is a similar situation in [`replayFunction`](https://github.com/dgraph-io/badger/blob/a39f9e06de643cd1edbc85492b18fec552fc7c7e/db.go#L93) but in that case the closure (`toLSM`) that calls `ensureRoomForWrite` doesn't return an error so I'm not sure what would be the most appropriate way to handle that.

manishrjain:
:lgtm: Thanks for this improvement.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/445#-:-LC1gD954arN6Y7K9dYg:b-7qjvwl)*

Issue #446, opened 2018-03-30
[enhancement] Block search performance: export the internal `restartInterval` configuration to the `Options` structure


schomatis:
# Summary

The rationale behind this request is that in the IPFS project (where Badger is intended to be the default data store in the near future) the keys are hashes (of the content being stored) and hence are (ideally) uniformly distributed with maximum distance in its key space, so the objective of this issue would be to be able to decrement the `restartInterval` value (by including it in Badger's [`Options`](https://github.com/dgraph-io/badger/blob/a39f9e06de643cd1edbc85492b18fec552fc7c7e/options.go#L52)) in production systems to minimize block searches (expecting a performance improvement).

# Detailed analysis

From my understanding of the code (please correct me if necessary), the keys in the SSTables are distributed in a two layer system. In the higher layer keys are grouped in blocks with each block exposing only one key that represents the group. In the lower layer, inside each block, it can be seen that the exposed key is the initial (lower) key of the group (`baseKey`) serving as a reference key (pivot in the searches) from where the rest of the keys of the blocks are derived, encoded as differences (`keyDiff`) from the `baseKey`. The objective behind this encoding system  is to reduce storage space by compressing the keys at the cost of CPU power (see later). The number of keys in a block is set by the [`restartInterval`](https://github.com/dgraph-io/badger/blob/a39f9e06de643cd1edbc85492b18fec552fc7c7e/table/builder.go#L30) variable which is hard-coded in the logic and can't be modified by the user (is not part of the API).

[...]

PR #447, opened 2018-04-01
[WIP] Export block size option

schomatis:
This is a work in progress, do not merge. It addresses issue #446.

It exports the number of entries per block (block size) configuration to Badger's options. There was no direct way for the table builder to reach the database general options so a temporal solution was used (passing the block size as a parameter and setting it with `SetBlockSize`).

The added test is a PoC benchmark to measure the performance impact of different block sizes.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/447)

### DIFF
--- a/db.go
+++ b/db.go
@@ -770,10 +770,11 @@ func arenaSize(opt Options) int64 {
 }
 
 // WriteLevel0Table flushes memtable. It drops deleteValues.
-func writeLevel0Table(s *skl.Skiplist, f *os.File) error {
+func writeLevel0Table(s *skl.Skiplist, f *os.File, blockSize int) error {
 	iter := s.NewIterator()
 	defer iter.Close()
 	b := table.NewTableBuilder()
+	b.SetBlockSize(blockSize)
 	defer b.Close()
 	for iter.SeekToFirst(); iter.Valid(); iter.Next() {
 		if err := b.Add(iter.Key(), iter.Value()); err != nil {
@@ -821,7 +822,7 @@ func (db *DB) flushMemtable(lc *y.Closer) error {
 		dirSyncCh := make(chan error)
 		go func() { dirSyncCh <- syncDir(db.opt.Dir) }()
 
-		err = writeLevel0Table(ft.mt, fd)
+		err = writeLevel0Table(ft.mt, fd, db.opt.BlockSize)
 		dirSyncErr := <-dirSyncCh
 
 		if err != nil {
--- a/db_test.go
+++ b/db_test.go
@@ -235,6 +235,68 @@ func TestGet(t *testing.T) {
 	})
 }
 
+// TODO: Not a real test.
+// This is a PoC benchmark written in the form of a test
+// to make the AppVeyor run it. It runs a serious of
+// get operations (with small values to avoid reading
+// from the log value file) with different block
+// sizes and different database sizes, as discussed in #446.
+func TestBenchmarkGetWithBlockSize(t *testing.T) {
+	for _, entriesNum  := range []float64{.1 , .5, 2} { // In M
+		for _, blockSize := range []int{1 , 10, 100, 1000} {
+			fmt.Printf("Entries: %.1fM. Block size: %d.\n", entriesNum, blockSize)
+			testBenchmarkGetWithBlockSize(t, blockSize, int(entriesNum * 1e6))
+		}
+	}
+}
+func testBenchmarkGetWithBlockSize(t *testing.T, blockSize int, entriesNum int) {
+	dir, err := ioutil.TempDir("", "badger")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+	opts := DefaultOptions
+	opts.Dir = dir
+	opts.ValueDir = dir
+
+	opts.BlockSize = blockSize
+
+	db, err := Open(opts)
+	require.NoError(t, err)
+
+	started := time.Now()
+	txn := db.NewTransaction(true)
+	for i := 0; i < entriesNum; i++ {
+		err = txn.Set([]byte(fmt.Sprintf("%d", i)), []byte{0})
+		require.NoError(t, err)
+
+		// Commit transaction to avoid "Txn is too big" error.
+		if i > 0 && i % 1000 == 0 {
+			err = txn.Commit(nil)
+			require.NoError(t, err)
+			txn = db.NewTransaction(true)
+		}
+	}
+	err = txn.Commit(nil)
+	require.NoError(t, err)
+
+	db.Close() // Flush SST to disk
+	fmt.Printf("Put time: %s\n", time.Since(started))
+
+	// Reopen to load SSTables with blocks.
+	db, err = Open(opts)
+	require.NoError(t, err)
+
+	started = time.Now()
+	txn = db.NewTransaction(false)
+	for i := 0; i < entriesNum; i++ {
+		_, err = txn.Get([]byte(fmt.Sprintf("%d", i)))
+		require.NoError(t, err)
+	}
+	txn.Discard()
+	fmt.Printf("Get time: %s\n", time.Since(started))
+
+	db.Close()
+}
+
 func TestGetAfterDelete(t *testing.T) {
 	runBadgerTest(t, nil, func(t *testing.T, db *DB) {
 		// populate with one entry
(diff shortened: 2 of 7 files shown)


manishrjain:
I think this PR is putting all the offsets into a long slice. The issue I see with that is it would consume a lot more memory than it currently does. Typically, a single SSTable in Badger can have ~2M key-value pairs. So, that'd be 2M uint32 offsets, which would be 2MB per SSTable, which gets very expensive very quickly. Currently, that consumes 20KB (because we only store 1 offset per 100 keys). And these offsets are always kept in memory, irrespective of whether the table is mmapped or on disk, etc.

So, there's a memory consumption angle to this PR, which needs to be explored further.

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/447#-:-LC1OMDO2jJkTRt0TndJ:b60iz2n)*

schomatis:
@manishrjain This PR ended up being a holding place for a couple of points discussed in #446 which deals with the number of entries stored per block.

The first two commits exported that option to perform benchmarks on it which show preliminary results that the block size may impact on search time. The reason for that (from what I've studied in the code) is because the search is linear inside the blocks (contrasted with logarithmic at the SST level).

The third commit tried to investigate further into this linear search and its impact on performance. It added an array of offsets to support a logarithmic search at the block level. Note that the added memory space could be recovered from the `prev` field in the entry headers which *already* exists and is taking the exact (`uint32`) size (please review https://github.com/dgraph-io/badger/issues/446#issuecomment-378036562).

Note that the removal of this `prev` field was not undertaken here (this would also mean modyfing the reverse iteration logic to rely on the added array of offsets) as the objective of the PR was to supply a PoC to support the arguments mentioned in #446, more work would be needed to get this into master.

schomatis:
@manishrjain I understand your concern and it's a good point.

Please bear with me as I'm lacking some Badger concepts and terminology.

> However, if we remove the prev field and have all the offsets in the index,

What would be the "index" here? If you're referring to the [`entryIndex`](https://github.com/dgraph-io/badger/pull/447/commits/47da3d6cb4924e0fbdfa3ac24ca420d6c4b11f49#diff-e41d31c5d7fbb3d20bd9ec01678bb258R101) added in this PR it's just a slice to the main data of the block obtained from the SSTable.

> The prev field is part of the SSTable. SSTable can be put on disk, mmaped, or loaded in RAM.

Yes, the headers of the key-value pairs (containing `prev`) were [loaded](https://github.com/dgraph-io/badger/blob/fa3538847194ebe5ab662141c3faf05359a29112/table/table.go#L175) from the SSTable, but as I understand so is everything else.

I'm not following how some part of the data is obtained from the SSTable and hence linked to its loading mechanism while some other (like the added array of offsets of key-value pairs) would not.

manishrjain:
See `readIndex` function here:

https://github.com/dgraph-io/badger/blob/f450ea41a32afd0b95b0f1dc4b7f1b146b3dc99f/table/table.go#L196

To run the binary search, to determine which block (of 100 keys) to read, we keep the `offset: key` in memory. This helps us determine which part of the SSTable file to start from.  This part if kept in memory, and it's relatively cheap. If we store 1M keys, this would be 10K `key+offset` pairs.

Look at blockIndex: https://github.com/dgraph-io/badger/blob/f450ea41a32afd0b95b0f1dc4b7f1b146b3dc99f/table/table.go#L51

This blockIndex is kept in memory always, to facilitate the binary search. To make an approach work where a block is practically of size one, we'll have to get rid of the block index; which *I think* would slow down the reads.

If you want to try, then maybe get rid of block index, and see what kind of performance you get by doing binary searches (with block size of 1) with sstable mmap-ed, instead of in `LoadToRAM`.

schomatis:
Thanks for that clear explanation.

> To make an approach work where a block is practically of size one, we'll have to get rid of the block index; which I think would slow down the reads.

Yes, you're right, but you're stating the *extreme* case a block of size one (i.e., removing the block architecture), this PR doesn't have that as an objective.

The first objective here is to expose the block size to the user for it to adapt it to its needs (as shown in some very simple benchmarks the block size does have an impact on search times). Sure, the user can set a block size of one, also it could set a block size as big as the size of the SST itself (which also wouldn't make much sense), but that is left for the user to experiment.

The second objective is to change the linear search of the block to a logarithmic one (which also has an impact on performance) without bloating the memory (reusing the `prev` field as mentioned above).

manishrjain:
> The first objective here is to expose the block size

Ok. I got that. That makes sense.

> The second objective is to change the linear search of the block to a logarithmic one

How would that be achieved, without somehow storing the list of offsets in memory? My main concern is here really.

schomatis:
> How would that be achieved, without somehow storing the list of offsets in memory? My main concern is here really.

I'm not sure I follow. This list of offsest, `entryIndex`, *will* be stored in memory, at the block level, as the rest of the key-value pairs in that block. It will live as long as the block lives, when the table iterator goes outside of the block the memory will be released (or at least it won't be referenced anymore). This is not a global list at the SST level, it doesn't replace `blockIndex`, it replicates its logic at the block level.

> The issue I see with that is it would consume a lot more memory than it currently does. Typically, a single SSTable in Badger can have ~2M key-value pairs. So, that'd be 2M uint32 offsets,

Yes, but they won't be loaded simultaneously all at the same time, they won't even coexist all together, they are stored (and loaded) 100 at a time (or whatever the block size is) in each block, that would be just 100 * 4 bytes, and even that extra space could be reclaimed by eliminating the `prev` field in the key-value pair and using the list of offsets for the reverse iteration logic.

Does that address your concern?

manishrjain:
> I'm not sure I follow. This list of offsets, entryIndex, will be stored in memory, at the block level, as the rest of the key-value pairs in that block.

> It will live as long as the block lives, when the table iterator goes outside of the block the memory will be released

So, every time that block needs to be read, you'd need to load up the offsets (releasing them when the search is done), then make the binary search. I think that'd be slow too.

schomatis:
> you'd need to load up the offsets (releasing them when the search is done),

The offsets are loaded with the rest of the block data (they *are* part of the block data) and release the same way as the rest of the data when the block iterator is deleted.

manishrjain:
Creation of the block iterator is really cheap right now. The extra part of parsing and loading offsets would add cost, which I doubt would be offset easily.

https://github.com/dgraph-io/badger/blob/fa3538847194ebe5ab662141c3faf05359a29112/table/table.go#L300

But, don't take my word for it, you could try your approach. Should be easy to benchmark and check if there are any performance gains.

schomatis:
> The extra part of parsing and loading offsets would add cost, 

That is done on demand, the `Search` method only reads (loads and parses) the offsets it needs for the binary search.

manishrjain:
I see. Re-read your code, it makes a lot more sense now. Could use more comments.

And did you run the benchmarks? How are they performing?

schomatis:
I performed some simple benchmarks (in one of the [commits](https://github.com/dgraph-io/badger/pull/447/commits/f682ed52dee11170c798dbb01ba0995507dcee88) of this PR) which indicated an improvement with respect to the linear search and also with various block sizes, see https://github.com/dgraph-io/badger/issues/446#issuecomment-378036562 for a detailed discussion. As mentioned there the official [benchmarks](https://github.com/dgraph-io/badger-bench/blob/master/BENCH-lmdb-bolt.md) you recommended were too involved for me to reproduce and also I lack the hardware to do a proper benchmark. Could someone from your team give it a try?

schomatis:
> Could use more comments.

Could you give me some feedback on what part of the [commit](https://github.com/dgraph-io/badger/pull/447/commits/47da3d6cb4924e0fbdfa3ac24ca420d6c4b11f49) should be explained further?

schomatis:
@manishrjain I've extracted the binary search commit in a separate [PR](https://github.com/dgraph-io/badger/pull/481) for you to test as you see fit and I'm closing this one.

My *main* objective here, as discussed in the issue, was to export the block size as a Badger option, all the rest (the binary search feature and all the following discussions) was submitted to support the argument that the block size has an impact on performance (especially if its being walked linearly).

Regrettably I don't have more time to allocate to this particular issue, as this is not a priority for the Badger transition in IPFS, and instead I'll be focusing in other aspects of Badger, like the ALICE tests that we've discussed.

manishrjain:
Thanks for putting the effort, @schomatis. This is definitely an area of exploration. If I get some cycles, I might test out your binary search change. If that shows a real win, then the block size change wouldn't be needed anyway.

PR #470, opened 2018-04-28
Fix locking under Windows

### LINKED ISSUE #424: windows LOCK exists, program  can't restart
In the windows environment, if the program is killed and the LOCK file is not deleted, the restart program will be wrongly reported 
```
Cannot create pid lock file "C:\\Users\\User\\Desktop\\chain.exe\\datadir\\LOCK".  Another process is using this Badger database: open C:\Users\User\Desktop\chain.exe\datadir\LOCK: The file exists.
```

CapacitorSet:
Fixes #424 by using an external lock library. I would have gladly ported it to dir_unix.go too, but I wasn't sure of how to deal with the readOnly flag (since I'm not practical with Badger).

I tested this PR under Windows, and it seems to work fine. I used this code by \@zjb0807:

```go
package main

import (
	"log"
	"time"

	"github.com/dgraph-io/badger"
)

func main() {
	opts := badger.DefaultOptions
	opts.Dir = "./badger"
	opts.ValueDir = "./badger"
	db, err := badger.Open(opts)
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()
	for i := 1; i < 1000; i++ {
		time.Sleep(1 * time.Second)
		log.Println(i)
	}
}
```

I launched it in a terminal, then launched it again, and it correctly exits with "Cannot get lock. Another process is using this Badger database".

If I quit the first process (Ctrl+C) and launch it again, it quits with "Unable to replay data value log: "./badger\\000000.vlog": Data corruption detected. Value log truncate required to run DB. This would result in data loss.", I'm not sure it is correct but given that it is a crash it might be normal.

### DIFF
--- a/dir_windows.go
+++ b/dir_windows.go
@@ -26,6 +26,7 @@ import (
 	"syscall"
 
 	"github.com/pkg/errors"
+	"github.com/theckman/go-flock"
 )
 
 func openDir(path string) (*os.File, error) {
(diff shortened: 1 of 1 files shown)


schomatis:
@djdv Could you please give this PR a try in Windows to test if it solves https://github.com/ipfs/go-ipfs/issues/4301?

djdv:
@schomatis 
I tested the example above first. Killing that process does not remove the lockfile, however, even with its existence I can obtain the lock as long as no other process holds it.
This seems mostly correct, but in my opinion, the lockfile should only exist while a process actually holds the lock.
I recently implemented that on Windows here:
https://github.com/go4org/go4/tree/master/lock
The exclusive lock is created, held, and stored in memory by the OS, so even if the OS suddenly halts, the lockfile will not exist unless a process holds it.

I'd suggest using that instead ;^)

schomatis:
@djdv Thanks for testing and for your input on the solution.

> Killing that process does not remove the lockfile, however, even with its existence I can obtain the lock as long as no other process holds it.

Great!



> the lockfile should only exist while a process actually holds the lock.

Agreed.

> I recently implemented that on Windows here:
> https://github.com/go4org/go4/tree/master/lock
> The exclusive lock is created, held, and stored in memory by the OS,

The flag `FILE_FLAG_DELETE_ON_CLOSE` seems very useful. So, @CapacitorSet, you have the package at your disposal if you're interested on adapting the solution.

Issue #464, opened 2018-04-24
Discard invalid versions of keys during compaction


jiminoc:
I'm hoping this is a configuration related issue but I've played around with the settings and I keep getting the same behavior. Tested on an i3.4XL in AWS, raid0 on the two SSD drives. 

Expected behavior of the code below:
 * keys/data are stored for 1hr, after a few hours the badger directory should stay fairly constant as you write/expire keys
 * I would expect to see sst files written and multiple size levels each level a larger file size
* memory should stay fairly consistent

Actual behavior seen:
 * OOM's after 12 hours
 * all sst files at 67MB (thousands of them)
 * disk fills up on a 4TB drive, no data appears to ttl out
 * file counts steadily increase until oom (there's no leveling off)
 * every hour the process stalls (assuming the stall count is being hit according to profiler)

Please advise of what is wrong in the code below, thanks!

3HRs of runtime you can see just linear growth
https://imgur.com/a/2UUfIrG

UPDATE:
I've also tried with these settings and memory doesn't grow as fast but it linearly climbs until OOM as well and the same behavior as above
```
dir := "/raid0/badgertest"
opts := badger.DefaultOptions
opts.Dir = dir
opts.ValueDir = dir
opts.TableLoadingMode = options.FileIO
opts.SyncWrites = false
db, err := badger.Open(opts)
```

```go
package usecases

import (
	"github.com/dgraph-io/badger"
	"github.com/dgraph-io/badger/options"
	"time"
	"fmt"
	"encoding/binary"
	"github.com/spaolacci/murmur3"
	"path/filepath"
	"os"
	"github.com/Sirupsen/logrus"
)

type writable struct {
	key   []byte
	value []byte
}


type BadgerTest struct {
	db *badger.DB
}

func NewBadgerTest() *BadgerTest {

	dir := "/raid0/badgertest"
	opts := badger.DefaultOptions
	opts.Dir = dir
	opts.ValueDir = dir
	opts.TableLoadingMode = options.MemoryMap
	opts.NumCompactors = 1
	opts.NumLevelZeroTables = 20
	opts.NumLevelZeroTablesStall = 50
	opts.SyncWrites = false
	db, err := badger.Open(opts)
	if err != nil {
		panic(fmt.Sprintf("unable to open badger db; %s", err))
	}
	bt := &BadgerTest{
		db: db,
	}

	go bt.filecounts(dir)
	return bt

}

func (b *BadgerTest) Start() {

	workers := 4
[...]

PR #471, opened 2018-05-01
Discard key versions during compaction

manishrjain:
- In response to #464 .
- If a key version has deletion marker or is expired, we can safely drop all the older versions of that key.
- If there's no overlap from lower levels, we can even drop the first such version.
- To avoid an edge case bug, we need to ensure that all versions of the same key are contained by the same SSTable. This is reflected by not closing a table as long as more versions are found. And by picking key ranges to include all versions of the keys. Both these measures ensure that all versions at the same level get compacted together.
- To avoid another edge case bug, we need to ensure that we don't drop versions above the current `readTs`. Note that managed DB would therefore not result in any version being discarded. Handling that is an open problem.

Badger Info:
- Print out the key ranges for each SSTable via `badger info`.
- Open an available port when running `badger` tool, so one can view the `/debug/request` and `/debug/events` links to understand what's going on behind the scenes.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/471)

### DIFF
--- a/cmd/badger/cmd/info.go
+++ b/cmd/badger/cmd/info.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/dgraph-io/badger"
 	"github.com/dgraph-io/badger/table"
+	"github.com/dgraph-io/badger/y"
 	humanize "github.com/dustin/go-humanize"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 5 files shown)


manishrjain:
Merged as e597fb7.

PR #472, opened 2018-05-02
Enable skiplist values that are up to 32 bits in length

andy-kimball:
Currently, inline skiplist values are limited to a length of 16 bits.
This commit increases that limit to 32 bits. It does this by making
node.value a uint64, and then packing the value offset and size into
it, as two uint32 values.

Because node.value is accessed with atomic.LoadUint64, it must be
aligned on a 64-bit boundary. This is guaranteed by Arena.putNode,
which is changed to always align on a 64-bit boundary rather than
on a pointer boundary (which would incorrectly align on 32-bit
boundary on a 32-bit machine).

### DIFF
--- a/skl/arena.go
+++ b/skl/arena.go
@@ -25,7 +25,12 @@ import (
 
 const (
 	offsetSize = int(unsafe.Sizeof(uint32(0)))
-	ptrAlign   = int(unsafe.Sizeof(uintptr(0))) - 1
+
+	// Always align nodes on 64-bit boundaries, even on 32-bit architectures,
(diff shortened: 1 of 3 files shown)


manishrjain:
:lgtm: Thanks a ton, @andy-kimball for sending out this PR quickly. Much appreciated!

Can you please sign the CLA, so I can merge this PR?

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/472#-:-LBTUaYu6adAztUf4OgD:b-7751d9)*

PR #473, opened 2018-05-02
[Do not merge] Mrjn/gc

manishrjain:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/473)


manishrjain:
Assimilated most of the useful changes here into my other PRs. So, closing this PR, and deleting this branch now.

Made these obsolete:
- Purging keys. Remove all of key purge logic. All that is now taken care of during compactions.
- Hole punching. It only works on Linux, and nothing else. So, removed all of that. Instead, I decreased the size of value log files, limiting them to 1M key-value pairs, so we can reclaim them in a limited duration (in my tests on external HDD, it was 30s).
- Periodic GC discard stats update: Again taken care of during compactions when it's discarding keys.

---

Review status: 0 of 16 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/473#-:-LC1Q2s4AFCGnTFKaCy7:b6c0mly)*

PR #479, opened 2018-05-08
Many GC related changes

manishrjain:
Max entries per value log file:
To ensure that we can GC a file in bounded time, we should restrict the max number of key-value pairs that can be contained by one log file. A new option now enforces that, taking the minimum of value log file size, and the number of entries contained to determine when a file should be rotated.

New move keyspace:
We should never write an entry with an older timestamp for the same key. We need to maintain this invariant to search for the latest value of a key, or else we need to search in all tables and find the max version among them.

- Value log GC moves key-value pairs to new files. This used to cause older versions of keys to be written to the top of LSM tree. This change puts those moves into a separate keyspace, prefixed by badgerMove. When doing a value read, we can look into this keyspace if the previous value pointer is no longer valid (pointed to a deleted value log file).
- Add an integration test, which runs for 3 mins and ensures that GC moves work as expected.
- Switch db.get back to stopping as soon as we encounter the first key (instead of looking into all levels).
- Collect value log GC stats when doing compactions.
- Trace log the entire value log GC, so we know how the file was picked and the result of sampling.
- Pick two value log files for GC. First from discard stats, and second randomly. Then try to GC both, and return on whichever is successful.
- Add a function to delete move keys corresponding to a value log file deletion.

Other option changes:
Increase the value threshold to 32 by default.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/479)


skilxnTL:
Will badgermove keys increase read iops on disk?

skilxnTL:
BTW，if user put key "foo" in badger, and do gc job, which inserts "$badgermove-foo" be stored in lsm tree, when will this '$bagermove-foo' key be deleted when user delete the key 'foo'?

PR #481, opened 2018-05-11
block iterator: seek with binary search

schomatis:
Add entry offsets to the block iterator to use a binary search algorithm
in `Seek` instead of the linear search (with repetitive calls to `Next`). The
entire object can benefit from this index, but in this first iteration the
code is modified as less as possible, just the `Seek` function uses it. Because
of this the binary search still uses the `Next` logic to parse the tested
element to keep the consistency in the `pos` and `err` internal attributes.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/481)

### DIFF
--- a/table/builder.go
+++ b/table/builder.go
@@ -75,6 +75,10 @@ type Builder struct {
 
 	restarts []uint32 // Base offsets of every block.
 
+	// Offsets of every entry within the current block being built.
+	// The offsets are relative to the start of the block.
+	entryOffsets []uint32
+
 	// Tracks offset for the previous key-value pair. Offset is relative to block base offset.
 	prevOffset uint32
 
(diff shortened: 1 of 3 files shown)


manishrjain:
Thanks for putting this together. Please include the before and after comparison benchmarks in the comments, before I begin the review.

justinfx:
@schomatis I'm curious to see the benchmarks for this feature so that it could be reviewed. I have a lot of seeking in databases that over 8 millions keys and wondering if this would really help when implementing "bucket"-like keyspaces.

schomatis:
Hey @justinfx, I did some very basic benchmarks in https://github.com/dgraph-io/badger/pull/447/commits/f682ed52dee11170c798dbb01ba0995507dcee88 to show the binary search improved performance compared to a linear one (see https://github.com/dgraph-io/badger/issues/446#issuecomment-378036562). At this moment I'm not planning on doing any more benchmarks (the previous ones were enough for what I wanted to prove) but feel free to ping me if you would like to expand on those or test this PR on your own DB.

PR #486, opened 2018-05-15
BeforeCommit

dc0d:
A `BeforeCommit func(txn *Txn, key, val []byte) error` function added to database options. This allows interesting scenarios for performing extra stuff inside the same transaction.

I have implemented this because I wanted to implement an embedded document database based on `badger`.

Using `BeforeCommit(...)` function, it is very easy to implement secondary indexes and even arbitrary indexes like views in CouchDB (I am taking this path).

And if `BeforeCommit` is not provided (being `nil`) or if `len(txn.pendingWrites) == 0`, then it will be ignored.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/486)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.vscode/bookmarks.json
--- a/db.go
+++ b/db.go
@@ -74,6 +74,8 @@ type DB struct {
 	flushChan chan flushTask // For flushing memtables.
 
 	orc *oracle
+
+	beforeCommit func(txn *Txn, key, val []byte) error
 }
 
 const (
@@ -166,6 +168,13 @@ func replayFunction(out *DB) func(Entry, valuePointer) error {
 func Open(opt Options) (db *DB, err error) {
 	opt.maxBatchSize = (15 * opt.MaxTableSize) / 100
 	opt.maxBatchCount = opt.maxBatchSize / int64(skl.MaxNodeSize)
+
+	defer func() {
+		if db == nil || err != nil {
+			return
+		}
+		db.beforeCommit = opt.BeforeCommit
+	}()
 
 	if opt.ValueThreshold > math.MaxUint16-16 {
 		return nil, ErrValueThreshold
--- a/options.go
+++ b/options.go
@@ -102,6 +102,9 @@ type Options struct {
 
 	// Truncate value log to delete corrupt data, if any. Would not truncate if ReadOnly is set.
 	Truncate bool
+
+	// If provided, will be called inside (*DB) Update(...) right before txn.Commit(nil).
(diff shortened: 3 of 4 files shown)


manishrjain:
Reviewed 4 of 4 files at r1.

---

*[db.go, line 176 at r1](https://reviewable.io/reviews/dgraph-io/badger/486#-LCa-N0yBT4lnQ98HXFF:-LCa-N0yBT4lnQ98HXFG:b1oxu6t) ([raw file](https://github.com/dgraph-io/badger/blob/c77d31a67b2d6b4c8d000d495f7100e93ce79c28/db.go#L176)):*
> ```Go
> 			return
> 		}
> 		db.beforeCommit = opt.BeforeCommit
> ```

Move this down, close to the end of the function. I don't think this assignment needs a defer func.

---

*[transaction.go, line 547 at r1](https://reviewable.io/reviews/dgraph-io/badger/486#-LCa0AZ3Dd-uSbYmA-ca:-LCa0AZ4ARpM2-jQBmVL:b-6vzykd) ([raw file](https://github.com/dgraph-io/badger/blob/c77d31a67b2d6b4c8d000d495f7100e93ce79c28/transaction.go#L547)):*
> ```Go
> 	}
> 
> 	if err := db.callBeforeCommit(txn); err != nil {
> ```

I don't see what's the need for this particular solution. You could just as well call it from when you're doing an update in your application. There's nothing new that this function is adding, that can't be achieved from within the application. Previously, I thought this needs to know if the commit has been successful or not. But, the actual Commit call is done after this function call -- so this can still fail. So.. what's the real advantage to pushing this inside Badger, instead of doing it outside?

---

*[transaction.go, line 561 at r1](https://reviewable.io/reviews/dgraph-io/badger/486#-LCa086KDaDh2C7seNTe:-LCa086KDaDh2C7seNTf:b-e37vhr) ([raw file](https://github.com/dgraph-io/badger/blob/c77d31a67b2d6b4c8d000d495f7100e93ce79c28/transaction.go#L561)):*
> ```Go
> 		return nil
> 	}
> 	it := txn.newPendingWritesIterator(false)
> ```

You don't need pending writes iterator.

---


*Comments from [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/486)*

dc0d:
(Issue [487](https://github.com/dgraph-io/badger/issues/487))

A key-value store is rarely used without some sort of secondary indexes. And calling this manually is an error prone task and adds lots of clutter to the consumer code-base.

This event has no overhead over the current performance of badger (if it's nil) and passes all the tests.

It seems a convenient utility (IMHO) since one has to create some sort of secondary index in majority of cases.

PR #488, opened 2018-05-15
Fixes #482 (be careful not to reuse keys and vals arguments of transaction methods)

### LINKED ISSUE #482: Mutability of []byte in transaction methods
Hi,

no really an issue with Badger, but rather a point of attention.

The Txn methods take *[]byte* arguments for keys and values. Byte slices are *mutable*.

For example, when performing a Delete(key), the Delete operation will not be completed right at the method call, but later on, when the transaction ends. In the meantime, the key is stored in an Entry object and lives inside the txn. 

So what happens I modify the 'key' byte slice after the Delete() call, but before the transaction ends ? Chaos :-)

Got bitten by that when introducing sync.Pool to reuse keys and values. The byte slices were reused, and the writes  started to be non sense. Pretty hard to debug too. Maybe that should be documented?

Thank you

stephane-martin:
PR to improve documentation as requested.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/488)

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -243,6 +243,9 @@ func (txn *Txn) checkSize(e *Entry) error {
 //
 // It will return ErrReadOnlyTxn if update flag was set to false when creating the
 // transaction.
+//
+// The current transaction keeps a reference to the key and val byte slice
+// arguments. Users must not modify key and val until the end of the transaction.
 func (txn *Txn) Set(key, val []byte) error {
 	e := &Entry{
 		Key:   key,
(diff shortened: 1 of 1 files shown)


stephane-martin:
added missing word, thx

manishrjain:
:lgtm: Thanks for your contribution!

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/488#-:-LCa20Kv2PBaFZmcp0hu:bc9pyb)*

PR #497, opened 2018-05-22
eliminate recursive call in yieldItemValue

stephane-martin:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/497)

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -140,42 +140,43 @@ func (item *Item) DiscardEarlierVersions() bool {
 }
 
 func (item *Item) yieldItemValue() ([]byte, func(), error) {
-	if !item.hasValue() {
-		return nil, nil, nil
-	}
+	for {
+		if !item.hasValue() {
+			return nil, nil, nil
+		}
 
-	if item.slice == nil {
-		item.slice = new(y.Slice)
-	}
+		if item.slice == nil {
+			item.slice = new(y.Slice)
+		}
 
-	if (item.meta & bitValuePointer) == 0 {
-		val := item.slice.Resize(len(item.vptr))
-		copy(val, item.vptr)
-		return val, nil, nil
-	}
+		if (item.meta & bitValuePointer) == 0 {
+			val := item.slice.Resize(len(item.vptr))
+			copy(val, item.vptr)
+			return val, nil, nil
+		}
 
-	var vp valuePointer
-	vp.Decode(item.vptr)
-	result, cb, err := item.db.vlog.Read(vp, item.slice)
-	if err != ErrRetry {
-		return result, cb, err
-	}
+		var vp valuePointer
+		vp.Decode(item.vptr)
+		result, cb, err := item.db.vlog.Read(vp, item.slice)
+		if err != ErrRetry {
+			return result, cb, err
+		}
 
-	// The value pointer is pointing to a deleted value log. Look for the move key and read that
-	// instead.
-	runCallback(cb)
-	key := y.KeyWithTs(item.Key(), item.Version())
-	moveKey := append(badgerMove, key...)
-	vs, err := item.db.get(moveKey)
-	if err != nil {
-		return nil, nil, err
+		// The value pointer is pointing to a deleted value log. Look for the move key and read that
+		// instead.
+		runCallback(cb)
+		key := y.KeyWithTs(item.Key(), item.Version())
+		moveKey := append(badgerMove, key...)
+		vs, err := item.db.get(moveKey)
+		if err != nil {
+			return nil, nil, err
+		}
+		if vs.Version != item.Version() {
+			return nil, nil, nil
+		}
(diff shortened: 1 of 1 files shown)


manishrjain:
:lgtm: Looks alright to me. Just needs to be tested. Can you run all the badger tests and ack that they run fine?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/497#-:-LD8v-lO078MPbPWQjsx:b-kbnlge)*

stephane-martin:
@manishrjain : well, I think your appveyor chain took care of running the tests! haven't found any problem myself.

manishrjain:
Yeah, remembered that after I asked you to run the tests.

So, my main question here would be, is this looping happening because we're looking for the "move" key repeatedly infinitely? And to answer that, I'd like to see if you can reproduce the issue by writing a unit test which brings Badger into this state of infinite recursion. So, can you add that unit test?

stephane-martin:
Not "infinitely", just a huge number of times.

I could not get a perfectly reproducible test case. What i would do was: ingest 200k key values (keys were 16 bytes, values 1 kbyte), delete the key values. Restart the program. Rince and repeat until crash happens. Usually I needed 3 or 4 restarts to get the crash. After the PR i can not reproduce anything anymore.

So i guess it has to do with the way yieldItemValues looks for the move key when massive deletes happened. But I'm not familiar enough with badger code to fully understand.

manishrjain:
Still would be good to put together a unit test to achieve this crash, with the previous code. You can close and reopen multiple times to do the "restarts".

romansemko:
In my case no restarts were necessary after the deletes. But I did restart after saving the millions batch of values: 
1. Written 40M values
2. Restarted
3. Deleted 68K values, reading some key values in between (not necessarily the ones to be deleted).
4. Recursive stack overflow on specific Get keys (always the same keys, even after restarting and trying again from step 2) freezing the program for some time before crashing.

I can try writing a simple example if I get free time these days. Does this PR prevent 'freezing" as when going into the infinite loop? Did you experience that?

romansemko:
Also, I am running database cleanup on start. Not sure if that has anything to do with it.

stephane-martin:
@romansemko : yes the PR prevents the panic from happening. You will still experience some pretty long processing times though.

romansemko:
Then we have two separate issues here:
1. Long lookup leading to huge recursive/for loops.
2. Stack overflow due to the huge recursive calls.

The PR fixes the second issue. I wonder what is causing the underlying one.

stephane-martin:
@romansemko AFAIU, the first issue is not an "issue", but the normal behaviour of badger after many KV have been deleted. It happens because the key space is not immediately compacted after the bulk delete.

manishrjain:
I want to validate the reason for the recursion. If we're just recursively looking for the same move key, then we can just exit after the first try. If there's some other reason for recursion, then we need to dig deeper. Switching the recursion with the for loop is only fixing the symptom, not the disease.

manishrjain:
It's not a chain really. If a key-value is GCed and moved multiple times, the move key gets overwritten. In the yieldItemValue, if you look here, after looking for the key and version, if we don't find it, we'll look for the move key with recursion, and then the same move key over and over again.

Actually, would be good to print out the key and the version that Item holds when yieldItemValue is called, and ascertain that. That's my hunch anyway.

https://github.com/dgraph-io/badger/blob/744717d82b6d465963bf8cb6690fdeb8e176dfc9/iterator.go#L168

stephane-martin:
So my PR to get rid of the recursive call does not solve anything. @manishrjain, you can cancel the PR. Thanks.

romansemko:
Will try the solution as described here:

https://reviewable.io/reviews/dgraph-io/badger/499#-LDO3M6E7GctC7Kmx1Rw

romansemko:
@manishrjain the solution described in https://reviewable.io/reviews/dgraph-io/badger/499#-LDO3M6E7GctC7Kmx1Rw fromt he closed/deleted PR https://github.com/dgraph-io/badger/pull/499 seems to have fixed the issue for me. At least there is no crash any more at the place where it was consistently failing before.

manishrjain:
@romansemko : Can you share your code which reproduces this issue? I can generate a fix, but I want to figure out why is this happening in the first place. Are we messing up somewhere else? Those sort of questions can only be answered if I'm able to reproduce this bug.

romansemko:
@manishrjain , the source is not yet open source. I would need to share the whole program, since it uses badger all over the place. 

Unfortunately, I couldn't isolate the the issue, yet! If I manage to create a simple example, I will post it. Hopefully, I can get some free time this week.

manishrjain:
Use #503 , not #499 . Also, are you comfortable sharing your set up with me in private?

PR #561, opened 2018-09-11
Share max batch count and size

valinurovam:
Cause badger can exec batch transactions, it should be easy to know current max batch size and count. Otherwise user (for me for example) take an error like
ErrTxnTooBig = errors.New("Txn is too big to fit into one request")

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/561)


manishrjain:
Thanks for the PR.

valinurovam:
Thank you

PR #577, opened 2018-09-21
Add a Jepsen style bank test for Badger

manishrjain:
**Breaking Change**

This change breaks Value API. Instead of Value returning a byte slice, we now expose the byte slice from within a function. This is because when we read a value, we acquire a read lock on the value log. With the previous API it was possible for a transaction to deadlock, if multiple reads are done on the same log file, and those read locks get interleaved with a write lock caused by log file rotation. The new API fixes that by releasing the lock before another value is read; and also makes it obvious to the end user that they must copy the value to be able to use it outside the function.

This change also adds a Jepsen style bank test for Badger, which is currently showing some transaction violations, needing further investigation.

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/577)

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -95,16 +95,23 @@ func (item *Item) Version() uint64 {
 // If you need to use a value outside a transaction, please use Item.ValueCopy
 // instead, or copy it yourself. Value might change once discard or commit is called.
 // Use ValueCopy if you want to do a Set after Get.
-func (item *Item) Value() ([]byte, error) {
+func (item *Item) Value(fn func(val []byte)) error {
 	item.wg.Wait()
 	if item.status == prefetched {
-		return item.val, item.err
+		if item.err == nil {
+			fn(item.val)
+		}
+		return item.err
 	}
 	buf, cb, err := item.yieldItemValue()
-	if cb != nil {
-		item.txn.callbacks = append(item.txn.callbacks, cb)
+	defer runCallback(cb)
+	if err == nil {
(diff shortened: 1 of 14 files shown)


iterator.go
```diff
@@ -95,16 +95,20 @@ func (item *Item) Version() uint64 {
 // If you need to use a value outside a transaction, please use Item.ValueCopy
 // instead, or copy it yourself. Value might change once discard or commit is called.
 // Use ValueCopy if you want to do a Set after Get.
-func (item *Item) Value() ([]byte, error) {
+func (item *Item) Value(fn func(val []byte)) error {
```
shazow: Hey, sorry for the unsolicited feedback. Not sure if you're mimicking another reference API, but in this case I wonder if this function signature would be more convenient:

```diff
- func (item *Item) Value(fn func(val []byte)) error {
+ func (item *Item) Value(fn func(val []byte) error) error {
```

Where the returned error would be forwarded into Value's error. This way it could be used as:

```go
	return item.Value(func(val []byte) error {
		return NewDecoder(bytes.NewReader(val)).Decode(into)
	})
```

Instead of

```go
    var err error
	err = item.Value(func(val []byte) {
		if decodeErr := NewDecoder(bytes.NewReader(val)).Decode(into); decodeErr != nil {
			err = decodeErr
		}
	})
    return err
```

Feel free to disregard this feedback if it's not relevant. :)

PR #580, opened 2018-09-23
Seek all levels of LSM tree for move keyspace

### LINKED ISSUE #578: Badger unable to return data after value log GC
Here's the reproducible example for #570 
Run it using the following bash script.

```
#!/bin/bash
while true; do
  go run main.go
  if [ $? -ne 0 ]; then
     break
  fi
done
```

```
package main

import (
	"io/ioutil"
	"log"
	"os"
	"fmt"

	"github.com/dgraph-io/badger"
)


var db *badger.DB

func main() {
	var err error
	dir, err := ioutil.TempDir("", "badger")
	logError(err, "create tmp")
	defer os.RemoveAll(dir)

	opts := badger.DefaultOptions
	opts.Dir = dir
	opts.ValueDir = dir
    opts.ValueLogMaxEntries = 64
    opts.MaxTableSize = 1 << 13

	db, err = badger.Open(opts)
	logError(err, "open badger")

	data := fmt.Sprintf("%100d",1)
    key := func(i int) string {
    	return fmt.Sprintf("%d%100d",i, i)
    }

    for i := 0; i < 32; i++ {
    	err := saveByKey(key(i), []byte(data))
    	logError(err, "saving value")
    }

    for i := 0; i < 10; i++ {
    	err := saveByKey(key(i), []byte(data))
    	logError(err, "saving value")
    }

    for i := 50; i < 72; i++ {
    	err := saveByKey(key(i), []byte(data))
    	logError(err, "saving value")
    }
  
    for i := 40; i < 72; i++ {
    	err := saveByKey(key(i), []byte(data))
    	logError(err, "saving value")
    }
     
    for i := 40; i < 72; i++ {
    	err := saveByKey(key(i), []byte(data))
    	logError(err, "saving value")
    }

    err = db.RunValueLogGC(0.5)

    if err == nil {
    	err = db.RunValueLogGC(0.5)
    }

	for i := 0; i < 10; i++ {
    	err = getByKey(key(i))
    	logError(err, "getting value")
    }
}


func saveByKey(key string, value []byte) error {
	return db.Update(func(txn *badger.Txn) error {
		return txn.Set([]byte(key), value)
	})
}

func getByKey(key string) error {
	return db.View(func(txn *badger.Txn) error {
		item, err := txn.Get([]byte(key))
		if err != nil {
			return err
		}
		val, err := item.Value()
		if err != nil {
			return err
		}
		if len(val) == 0 {
			log.Fatalf("key not found %q", key)
		}
		return err
	})
}

func logError(err error, context string) {
	if err != nil {
		log.Fatal(context, ":", err)
	}
}
```

manishrjain:
In Badger move keyspace, an older version of the key can be written to the LSM tree after a newer version of the key. After some compactions, this can mean that a higher level in the tree can have an older version, while a lower level can have a newer version. To fix this, we must seek through all the levels to find the right version -- we only need to do this for the move keyspace.

Note that we used to do this exact logic on the entire keyspace before. Move keyspace was introduced to avoid that need to seek through all levels. Move keyspace is still a great thing because we only need to seek all levels for move keys -- seeks for normal keyspace can be stopped once any version of the key is found.

Fixes #578 .

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/580)

### DIFF
--- a/db.go
+++ b/db.go
@@ -499,11 +499,11 @@ func (db *DB) getMemTables() ([]*skl.Skiplist, func()) {
 //
 // Update (Sep 22, 2018): To maintain the above invariant, and to allow keys to be moved from one
 // value log to another (while reclaiming space during value log GC), we have logically moved this
-// need to write older versions in levels above newer versions to the badgerMove keyspace. Thus, for
-// normal gets, we can stop going down the LSM tree once we find any version of the key (note
-// however that we will ALWAYS skip versions with ts greater than the key version). However, if that
-// key has been moved, then for the corresponding "movekey", we'll look through all the levels of
-// the tree to ensure that we pick the highest version of the movekey present.
+// need to write "old versions after new versions" to the badgerMove keyspace. Thus, for normal
+// gets, we can stop going down the LSM tree once we find any version of the key (note however that
+// we will ALWAYS skip versions with ts greater than the key version).  However, if that key has
+// been moved, then for the corresponding movekey, we'll look through all the levels of the tree
+// to ensure that we pick the highest version of the movekey present.
 func (db *DB) get(key []byte) (y.ValueStruct, error) {
 	tables, decr := db.getMemTables() // Lock should be released.
 	defer decr()
--- a/db.go
+++ b/db.go
@@ -17,6 +17,7 @@
 package badger
 
 import (
+	"bytes"
 	"encoding/binary"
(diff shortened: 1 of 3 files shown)


manishrjain:
Need to make the code from bug report #578 part of the test suite before submitting this.

PR #590, opened 2018-09-25
Add (*Txn).GetReadTimestamp

alexandrestein:
Those two functions can make things much easier when you need concurrent but consistant iterators.

From an initial transaction `*Txn`, get its read timestamp to open a new transaction with the same timestamp with`*DB.NewTransactionAt`.

Now you have consistant iterators which has the same values.

I did not found any workaround because **Badger** panics when opening multiple iterators in the same transaction.

For example: 
``` golang
iter := txn.NewIterator()

timestamp := txn.GetReadTimestamp()
txn2 := db.NewTransactionAt(timestamp, false)
iter2 := txn2.NewIterator(badger.DefaultIteratorOptions)
```

I need this or something similar to have multiple consistant iterators. 
I'm trying to implement a [bleve index](https://godoc.org/github.com/blevesearch/bleve/index/store) store and it needs to have concurent but consistent iterators.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/590)

### DIFF
--- a/managed_db.go
+++ b/managed_db.go
@@ -42,7 +42,11 @@ func OpenManaged(opts Options) (*DB, error) {
 // This is only useful for databases built on top of Badger (like Dgraph), and
 // can be ignored by most users.
 func (db *DB) NewTransactionAt(readTs uint64, update bool) *Txn {
-	if !db.opt.ManagedTxns {
+	// Can't use this function if:
+	//     - the database is not managed AND update is true
+	// OR
+	//     - the database is not managed AND the readTs is zero.
+	if !db.opt.ManagedTxns && update || !db.opt.ManagedTxns && readTs == 0 {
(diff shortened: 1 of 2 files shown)


alexandrestein:
I used an other approach in this [branch](https://github.com/dgraph-io/badger/compare/master...alexandrestein:txnClone) which can also solve my issue.

I hope one of those will be OK for you.

manishrjain:
Hey @alexandrestein , I think I can make a change which would allow multiple iterators to run simultaneously. That way you don't need to know about read timestamps and how that mechanism works.

manishrjain:
See commit: https://github.com/dgraph-io/badger/commit/41d9656c08175a3566576bde596e26c288bb64e2

Note that there was a breaking change a couple of commits before this commit, which changes how Value is accessed. So, you'd need to modify your code accordingly.

PR #593, opened 2018-09-26
Fix for a transaction violation in Badger

manishrjain:
- It seems like `farm.Fingerprint` is undeterministic causing different hashes for the same byte slice. This causes the txn conflict detection to fail, causing 2 concurrent txns to both succeed.
- Remove fingerprint and store keys as strings for conflict detection.
- Currently testing this change [WIP, OK for over 4h, longest run ever]

Note from observation:
```
Comparing @ts=437427 with @ts=437428
Index: 92881. Account [{Id:92881 Bal:120}] -> [{Id:92881 Bal:115}]
Index: 94432. Account [{Id:94432 Bal:115}] -> [{Id:94432 Bal:120}]

2018/09/26 09:49:12 Balance did NOT match up. Expected: 10000000. Received: 9999995
Index: 57525. Account [{Id:57525 Bal:115}] -> [{Id:57525 Bal:120}]
Index: 94432. Account [{Id:94432 Bal:120}] -> [{Id:94432 Bal:110}]

So, both read $115. @ts=437428, 94332 acct went from 115 -> 120.
@ts=437443, 94332 acct went from the read 115 -> 110 (but the actual was 120).
This should have been a txn conflict. It seems like farm.Fingerprint might have
generated a non-deterministic hash.
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/593)

### DIFF
--- a/badger/cmd/bank.go
+++ b/badger/cmd/bank.go
@@ -59,7 +59,7 @@ failure of the total invariant.
 	RunE: runDisect,
 }
 
-var numGoroutines, numAccounts int
+var numGoroutines, numAccounts, numPrevious int
 var duration string
 var stopAll int32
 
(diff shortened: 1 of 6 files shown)


manishrjain:
farm.Fingerprint was not the issue. The issue was fixed by https://github.com/dgraph-io/badger/commit/3c96d0831347c0383738f7de911673e7a6d354c8.

PR #597, opened 2018-10-02
reduce allocations by avoiding AssertTruef

Stebalien:
ParseKey is called *a lot*. Unfortunately, AssertTruef converts implicitly the arguments to interfaces which implicitly allocates. This was showing up as the primary source of allocations in go-ipfs by a wide margin.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/597)

### DIFF
--- a/y/y.go
+++ b/y/y.go
@@ -134,7 +134,7 @@ func ParseKey(key []byte) []byte {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


manishrjain:
Thanks for the optimization.

PR #603, opened 2018-10-04
Add a ValueSize method to Item

Stebalien:
This is *much* faster than actually loading the item from disk.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/603)

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -19,6 +19,7 @@ package badger
 import (
 	"bytes"
 	"fmt"
+	"hash/crc32"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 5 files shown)


iterator.go
```diff
@@ -244,6 +245,22 @@ func (item *Item) EstimatedSize() int64 {
...
+func (item *Item) ValueSize() int64 {
+	if !item.hasValue() {
+		return 0
+	}
+	if (item.meta & bitValuePointer) == 0 {
+		return int64(len(item.vptr))
+	}
+	var vp valuePointer
+	vp.Decode(item.vptr)
+	return int64(vp.Len) - int64(len(item.key)) - headerBufSize - crc32.Size - 8 // len(uint64) (timestamp)
```
Stebalien: Is that `-8` a valid assumption to make? I'm pretty sure that's the timeout but I'm not sure if this is something I can rely on.

Stebalien:
Note: This change isn't *critical*, it would just be really nice to have.

iterator.go
manishrjain: The 8 is the timestamp associated with the key. When doing `encodeEntry`, we write the key including the timestamp, but when parsing it via Item, the key is stored without the timestamp (and Item.Version tracks the ts). So, this is right. Could use some comments though, I can add that later.

value.go
```diff
@@ -293,7 +293,7 @@ func (vlog *valueLog) iterate(lf *logFile, offset uint32, fn logEntry) (uint32,
 		}
 
 		var vp valuePointer
-		vp.Len = uint32(headerBufSize + len(e.Key) + len(e.Value) + 4) // len(crcBuf)
+		vp.Len = uint32(headerBufSize + len(e.Key) + len(e.Value) + crc32.Size)
```
manishrjain: This len can be kept. So, if someone switches how crcBuf is created, then we this would stay in sync with automatically.

manishrjain:
LGTM. Couple of comments, if you want to tackle those. Otherwise, I can do them after merging.

value.go
Stebalien: The value is *currently* hard-coded (`len(crcBuf)` is just a comment). Is that what you're referring to?

manishrjain: Ah.. Should be len(crcBuf) then.

manishrjain: I'll take care of it later.

manishrjain:
Thanks for your PR.

PR #605, opened 2018-10-04
New build tags to control logging.

### LINKED ISSUE #556: How do you turn of badger logging?
Every time I open a database I see the following lines in the console. I can't find a setting to turn this debug information off.

```
2018/08/30 17:44:33 Replaying from value pointer: {Fid:0 Len:0 Offset:0}
2018/08/30 17:44:33 Iterating file id: 0
2018/08/30 17:44:33 Iteration took: 15.378µs
```

srfrog:
New build tags to control logging.

This PR adds two new build tags for enabling and disabling
info and event logging. By default, no logging is enabled.

The build tag debugcompat enabled the compatible mode before
this change. The tag debugevents enables only event logging.

To enable compatible logging:
```sh
go build -tags debugcompat
```

To enable event logging:
```sh
go build -tags debugevents
```

Closes #556


JeanMertz:
Just a quick note:

* I doesn't look this change is part of an official release yet
* The introduction of the extra log lines (as reported in https://github.com/dgraph-io/badger/issues/556) _is_ part of the latest release `1.5.4`
* But that release isn't in the `CHANGELOG` or on the Releases page (there's a Git tag for it though)
* We noticed it because updating to the latest version introduced a lot of Badger specific log lines in our benchmarks. Thanks for fixing that, hope this will finds its way to a new release soon 👍

leonklingele:
How does this PR make "default logging silent"?

PR #608, opened 2018-10-08
Create a WriteBatch API to allow efficient serial writes

manishrjain:
**Breaking change**

- This change removes the ability to provide a callback in Commit. Instead, it creates another function called CommitWith, which takes a callback.
- Previously, a user specified callback might or might not get executed. This PR fixes that by guaranteeing that the commit callback is always executed.
- Create a WriteBatch API to batch up multiple updates into a single txn. The txn gets Committed via callback, so the user gets a simple efficient way to do a lot of writes.
- Run only a few goroutines to execute these user callbacks. Previously, we were shooting off one goroutine per Commit callback.

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/608)

### DIFF
--- a/README.md
+++ b/README.md
@@ -614,51 +614,24 @@ There are multiple workarounds during iteration:
 
 Are you creating a new transaction for every single key update, and waiting for
 it to `Commit` fully before creating a new one? This will lead to very low
-throughput. To get best write performance, batch up multiple writes inside a
-transaction using single `DB.Update()` call. You could also have multiple such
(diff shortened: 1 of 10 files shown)


campoy:
This [change](https://github.com/dgraph-io/badger/pull/608/files#diff-086d95904066e9938ebbede858e38b20R579) broke the API by changing the signature of `Txn.Commit`.

Added warning to title.

Issue #532, opened 2018-07-21
Use channel to run txn callbacks


manishrjain:
Currently, we just spin up goroutines for every txn.Commit with callback. Instead, push these to a channel, and just use a few goroutines to run them.

manishrjain:
https://github.com/dgraph-io/badger/commit/6c8b15c2c7441e0958bae3d722cc9570ebdc63a5

Things slow down a lot by doing this. Tried and reverted.

PR #622, opened 2018-11-05
Fix: Use channel to run txn callbacks #532

MaruHyl:
Hi, @manishrjain , I see your issue and I am trying to solve it.

I read your commit (6c8b15), and here are a few questions.
1.Buffered chan?
It can only cache a limited number of elements, so If you set it too small, commitWith may be blocked frequently. If you set it too large, it takes up a lot of memory when it is initialized, even if it is not used.
2.Number of consumer?
If the number is set too small, the performance of handling callbacks will be slower. If the number is set too large, then too many goroutines will be created at startup.

My solution to these problems is
1.Buffered chan?
I made a wrapper struct on the basis of chan and slice. Slice can be used to dynamically cache elements, no limit. Chan can be used to notify and block consumer. And the most important thing is that it won't block CommitWith
2.Number of consumer?
Just 1.Only need to create one goroutine when initializing, it is used to take the callback, and create a goroutine to consume it (if the maximum number of allowed is not exceeded).

Can you take some time to revivew my code? All test is passed.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/622)

### DIFF
--- a/db.go
+++ b/db.go
@@ -44,11 +44,12 @@ var (
 )
 
 type closers struct {
-	updateSize *y.Closer
-	compactors *y.Closer
-	memtable   *y.Closer
-	writes     *y.Closer
-	valueGC    *y.Closer
+	updateSize  *y.Closer
+	compactors  *y.Closer
+	memtable    *y.Closer
+	writes      *y.Closer
+	valueGC     *y.Closer
+	txnCallback *y.Closer
 }
 
 // DB provides the various functions required to interact with Badger.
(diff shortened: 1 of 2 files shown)


manishrjain:
This was fixed by a PR already. And turned out that this mechanism was slow. So, I had removed it here: https://github.com/dgraph-io/badger/commit/6c8b15c2c7441e0958bae3d722cc9570ebdc63a5

MaruHyl:
> This was fixed by a PR already. And turned out that this mechanism was slow. So, I had removed it here: [6c8b15c](https://github.com/dgraph-io/badger/commit/6c8b15c2c7441e0958bae3d722cc9570ebdc63a5)

I saw this commit before I push this PR, and I mentioned it above `I read your commit (6c8b15), and here are a few questions...`. 
 I think there is a better way. Can you take some time to read the code?

MaruHyl:
@manishrjain

manishrjain:
There's no right number of goroutines here. I've tried with 16 and so on... The contention as I mentioned is a mixture of the channel and the goroutines.

PR #633, opened 2018-12-04
Supporting both managed and unmanaged write batches

### LINKED ISSUE #629: [enhancement] Using WriteBatch with OpenManaged
I'm trying to do a lot of bulk writes on a managed DB. Currently, WriteBatch uses the unmanaged transactions API, so the only way I can do bulk writes is by managing my own transactions. It would be convenient to be able to use a similar API for WriteBatches with a managed DB, since I need the `commitAt` functionality for other transactions.

gitlw:
fixes #629

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/633)

### DIFF
--- a/backup_test.go
+++ b/backup_test.go
@@ -18,6 +18,7 @@ package badger
 
 import (
 	"bytes"
+	"encoding/binary"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


manishrjain:
Haven't been worked on for a while, so closing this.

PR #645, opened 2018-12-18
New tools, new option

manishrjain:
- Add a new flatten command, which would flatten out all the LSM tree levels into a single level. It ensures that that level does not exceed its max size limits, as specified in the options.
- Add a new fill command, which would random key-values to Badger. Useful for testing.
- Add a new option to skip force compaction of Level 0 on DB.Close.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/645)

### DIFF
--- a/badger/cmd/fill.go
+++ b/badger/cmd/fill.go
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2018 Dgraph Labs, Inc. and Contributors
+ *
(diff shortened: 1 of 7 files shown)


campoy:
This PR dropped the field `DoNotCompact` from `badger.Options` making it a breaking change.

PR #655, opened 2018-12-29
Renaming and Refactoring

manishrjain:
- Flatten function can now be called from live code path, i.e. while Badger is running.
- Refactor stop and start compactions code, so it can be used in Flatten, DropAll, and other places.
- Rename protos package to pb. Consolidate both the proto files into one pb.proto.
- Rename KeyToKVList to just KeyToList in Stream.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/655)


campoy:
In the future, please mark all breaking changes with `[BREAKING CHANGE]` on their title.

campoy:
In this case, it's almost OK since the API you're changing was introduced in the previous commit, but it's still good practice to mark all breaking changes so we don't miss any.

PR #661, opened 2019-01-02
test: Fix TestDropReadOnly failed on windows

Xuanwo:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/661)

### DIFF
--- a/managed_db_test.go
+++ b/managed_db_test.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"math/rand"
 	"os"
+	"runtime"
 	"sync"
 	"sync/atomic"
 	"testing"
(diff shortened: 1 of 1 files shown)


manishrjain:
The bigger question is, why does this fail on Windows? `vlog.deleteLogFile` does call `unmap` before closing the file and deleting it.

@martinmr : Can you figure out the cause of this failure, and fix it?

Xuanwo:
Well, it makes sense. I force pushed a commit to do the change, PLTA @martinmr

Xuanwo:
OK, it's better if no os-specific code needed.

martinmr:
I checked and this looks like a limitation of the file system so it should be fine to have the OS-specific check since there's no way around it.

@manishrjain,  what do you say?

martinmr:
Thanks for the PR. We discussed this issue internally and since there's no easy way to get around this limitation, we'll merge your PR.

PR #646, opened 2018-12-19
Add support for go-module

thedevsaddam:
Add support for go-module

Updates #758 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/646)

PR #673, opened 2019-01-05
Add import comment for pb package

Photonios:
This makes badger work with Go 1.11's module support. Currently, when
trying to use Go 1.11 modules in a project that uses Badger yields
in the following error:

```
unknown import path "github.com/dgraph-io/badger/pb": cannot find module providing package github.com/dgraph-io/badger/pb
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/673)


Photonios:
It seems that it is not possible to instruct protoc / protoc-gen-go to generate the `// import "..."` comment. This means that the only option to make `dgraph-io/badger/pb` installable through `go get` is to patch the generated file somehow or for Badger to actually use Go 1.11's modules. I see some progress is being made in #646.

Thoughts?

manishrjain:
Not sure. If you figure something out, please create another PR. Closing this one due to inactivity.

PR #679, opened 2019-01-11
Update README with rsync backup script.

danielmai:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/679)

### DIFF
--- a/README.md
+++ b/README.md
@@ -559,7 +559,6 @@ rsync -avz --delete db/ dst
 while !! | grep -q "(MANIFEST\|\.sst)$"; do :; done
 ```
 
-
 ### Memory usage
 Badger's memory usage can be managed by tweaking several options available in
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -543,6 +543,21 @@ command above to upgrade your database to work with the latest version.
 badger_backup --dir <path/to/badgerdb> --backup-file badger.bak
 ```
 
+A Badger database is rsync-friendly, so rsync can be used as a rudimentary way
```
manishrjain: We recommend all users to use the `Backup` and `Restore` APIs and tools. However, Badger is also rsync-friendly because all files are immutable, barring the latest value log which is append-only. So, rsync can be used as rudimentary way to perform a backup. In the following script, we repeat rsync to ensure that the LSM tree remains consistent with the MANIFEST file while doing a full backup.

Issue #683, opened 2019-01-13
Quick way to drop all keys with a certain prefix


manishrjain:
We now have a `DropAll` API, which can drop all the tables from LSM tree, and log files from value log. It contains mechanisms by which we can pause writes and compactions, and alter the tree. We use this to provide a `Flatten` API as well.

We can use these mechanisms to rewrite the LSM tree to drop all the keys with a certain prefix, without using delete markers. That'd be great to provide an idea of namespaces to users, and also to clear out certain predicates in Dgraph.

Asarew:
I think that this AP would be beter if you're able to specify a range instead of only a prefix. Which would also broaden the use case.

PR #715, opened 2019-02-15
Introduce DropPrefix API in Badger

manishrjain:
This API would allow Badger to completely get rid of the given prefix, P. It does this by:

1. Flushing memtables to disk (skipping over the keys with P).
2. Doing normal compaction from L0 -> L1 (skipping keys with P).
3. Doing same level compaction on all the rest of the levels, picking tables which might have the prefix and rewriting them to new tables (again, skipping keys with P).

Before starting, all the writes are paused, the usual bits which apply to `DropAll` apply here as well.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/715)

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -65,7 +65,9 @@ func (r keyRange) overlapsWith(dst keyRange) bool {
 }
 
 func getKeyRange(tables []*table.Table) keyRange {
-	y.AssertTrue(len(tables) > 0)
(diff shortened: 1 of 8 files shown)


mangalaman93:
sure, I will work on it!

mangalaman93:
#683

PR #725, opened 2019-02-26
fix valueLog doRunGC to use consistent units

mschoch:
this version introduces sizeWindowM to use when comparing against
other values already in MB

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/725)


mschoch:
Alternate impl of #724 which keeps MB in places it was already used

PR #728, opened 2019-02-28
Update gogo/protobuf to 1.2.1

Kubuxu:
This PR updates generated code to one generated with gogo/protobuf v1.2.1 and also fixes `gen.sh` script.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/728)


Kubuxu:
Appveyor failed because of:
```
Mmap value log file. Path=C:\Users\appveyor\AppData\Local\Temp\1\badger777207792\000000.vlog. Error=truncate: truncate C:\Users\appveyor\AppData\Local\Temp\1\badger777207792\000000.vlog: There is not enough space on the disk
```

manishrjain:
We use KV defined here in Dgraph. Not sure if this change would break Dgraph or not -- Dgraph uses vanilla protobuf.

Kubuxu:
gogo/protobuf and golang/protobuf can coexist without any issues. You can use their `*/proto` packages interchangeably.

gedw99:
@Kubuxu are you sure ? I also would be broken otherwise. Got a link maybe ?

Kubuxu:
I've updated it to use golang import.

PR #731, opened 2019-03-05
Fixing errors found by go vet

### LINKED ISSUE #694: Fix issues with vet on 1.12
Running the `vet` that will come with 1.12 shows the following issues:

```
github.com/dgraph-io/badger/db.go:332:37: tautological condition: nil == nil
github.com/dgraph-io/badger/db.go:606:10: impossible condition: nil != nil
```
The first one is:
```
        // Now close the value log.
        if vlogErr := db.vlog.Close(); err == nil {
                err = errors.Wrap(vlogErr, "DB.Close")
        }
```
Where the `err` that's being checked for nil should probably be `vlogErr`.

The second one is:
```
                for err := db.ensureRoomForWrite(); err == errNoRoom; err = db.ensureRoomForWrite() {
						// trimmed
						...
                }
                if err != nil {
                        done(err)
                        return errors.Wrap(err, "writeRequests")
                }
```

Again the `err` that's being checked for nil is not the `err` that's declared with `:=` in the for loop as that variable is scoped to just the loop.

gitlw:
Fixes #694 
though I'm not able to reproduce the reported errors by running "go vet" using the go in version 1.11.2 or 1.12.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/731)

### DIFF
--- a/db.go
+++ b/db.go
@@ -333,7 +333,7 @@ func (db *DB) Close() (err error) {
 	db.closers.writes.SignalAndWait()
 
 	// Now close the value log.
-	if vlogErr := db.vlog.Close(); vlogErr == nil {
+	if vlogErr := db.vlog.Close(); vlogErr != nil {
 		err = errors.Wrap(vlogErr, "DB.Close")
 	}
 
--- a/db.go
+++ b/db.go
@@ -333,7 +333,7 @@ func (db *DB) Close() (err error) {
 	db.closers.writes.SignalAndWait()
 
 	// Now close the value log.
-	if vlogErr := db.vlog.Close(); err == nil {
+	if vlogErr := db.vlog.Close(); vlogErr == nil {
 		err = errors.Wrap(vlogErr, "DB.Close")
 	}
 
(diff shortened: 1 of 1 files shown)


db.go
```diff
@@ -602,7 +602,7 @@ func (db *DB) writeRequests(reqs []*request) error {
 		}
 		count += len(b.Entries)
 		var i uint64
-		for err := db.ensureRoomForWrite(); err == errNoRoom; err = db.ensureRoomForWrite() {
+		for err = db.ensureRoomForWrite(); err == errNoRoom; err = db.ensureRoomForWrite() {
```
mangalaman93: Is `go vet` okay with this? We are still shadowing the previous value. Wouldn't it be better if we change the lines above to -

```
if err := db.vlog.write(reqs); err != nil {
	done(err)
	return err
}
```

and then, create the variable inside the loop?

mangalaman93: Seems like we are using this variable outside the for loop. In that case, we should declare it before the loop and then use it?

gitlw: err was already defined above on line 591 of this file.

mangalaman93: :+1:

Issue #723, opened 2019-02-26
Replace "Rewind" with "Seek" in MergeOperator.iterateAndMerge


milkystan:
https://github.com/dgraph-io/badger/blob/291295e18f0eca0ea8ca6df9e2725cfe73b4c915/merge.go#L69-L72

The for loop will quit if the first key in db does not valid for prefix.

PR #734, opened 2019-03-07
IterateAndMerge should use it.Seek(..) instead of it.Rewind()

jarifibrahim:
Fixes https://github.com/dgraph-io/badger/issues/723

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/734)


jarifibrahim:
@manishrjain what should happen if the merge operator was created using a key that doesn't exists in badger? Should there be an error?

PR #746, opened 2019-03-22
Force flush memtable when value log file is rotated

### LINKED ISSUE #549: Force flush memtable when log file is rotated
This would ensure that the value log head would be persisted on disk, thus the max replay that would need to be done would be one log file, not more.

ashish-goswami:
Fixes #549

This commit contains changes to force flush memtable if value
log file has been rotated while writing data. This will ensure
persistent of latest value of value log head pointer to disk. Hence
reducing the replay count to max one file at start.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/746)

### DIFF
--- a/db.go
+++ b/db.go
@@ -45,6 +45,8 @@ var (
 	txnKey            = []byte("!badger!txn")     // For indicating end of entries in txn.
 	badgerMove        = []byte("!badger!move")    // For key-value pairs which got moved during GC.
 	lfDiscardStatsKey = []byte("!badger!discard") // For storing lfDiscardStats
+
+	errNilMemtable = errors.New("memtable is nil")
 )
 
 type closers struct {
@@ -596,10 +598,13 @@ func (db *DB) writeRequests(reqs []*request) error {
 	}
 	db.elog.Printf("writeRequests called. Writing to value log")
 
+	var valueLogFileRotated bool
 	err := db.vlog.write(reqs)
-	if err != nil {
+	if err != nil && err != errLogFileRotated {
 		done(err)
 		return err
+	} else if err == errLogFileRotated {
+		valueLogFileRotated = true
 	}
 
 	db.elog.Printf("Writing to memtable")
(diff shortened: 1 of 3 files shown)


db.go
```diff
@@ -45,6 +45,10 @@ var (
 	txnKey            = []byte("!badger!txn")     // For indicating end of entries in txn.
 	badgerMove        = []byte("!badger!move")    // For key-value pairs which got moved during GC.
 	lfDiscardStatsKey = []byte("!badger!discard") // For storing lfDiscardStats
+
+	// errNilMemtable is returned when memtable is nil
+	// while performing some operation on it
```
mangalaman93: don't need this line

value_test.go
```diff
@@ -757,16 +755,64 @@ func TestPenultimateLogCorruption(t *testing.T) {
...
+		_, err := txn.Get(h.key(7)) // Verify that 7 is gone.
 		require.Equal(t, ErrKeyNotFound, err)
 		return nil
 	})
 	require.NoError(t, err)
 	require.NoError(t, db1.Close())
 }
+func TestForceFlushMemtable(t *testing.T) {
+	// This test checks, if latest value of value log head is getting persisted
+	// after value log file rotation. Latestt value is persisted by flushing current
```
mangalaman93: Latest*

value_test.go
```diff
@@ -757,16 +755,64 @@ func TestPenultimateLogCorruption(t *testing.T) {
...
+	// We want to check persisted value of value log head, hence open another instance of
+	// db from same opt. Release directory lock of first instance
+	if db0.dirLockGuard != nil {
+		require.NoError(t, db0.dirLockGuard.release())
+	}
+	if db0.valueDirGuard != nil {
+		require.NoError(t, db0.valueDirGuard.release())
+	}
+
+	// wait for memtable to flush to disk
```
mangalaman93: Is this a deterministic test?

ashish-goswami: It is passing every time on my local system. Lets see how it behaves on teamcity. Otherwise can increase wait time.

mangalaman93: Just mention it here in that case on the test so that in case it ever fails, somebody else can fix it too.

ashish-goswami:
Closing this, as we have another PR for same issue. https://github.com/dgraph-io/badger/pull/766

PR #750, opened 2019-03-28
Expose Sync method in DB

### LINKED ISSUE #749: Expose method to sync all writes to disk
Currently sync writes are controlled by **syncWrites** flag present in **options**. If **syncWrites** is true, all writes to value log are sync. If **syncWrites** is false, all writes to value log are async and only synced while rotating value log file.

To provide more control to the user, badger can provide a method to sync value logs writes whenever required. This can act as intermediate of above approaches. Note that this will only sync the latest value log, all the value logs before would already be synced on file rotate.

ashish-goswami:
Currently syncing of DB data to disk is controlled by
syncWrites flag present in options. If syncWrites is true
value log file will be synced to disk for every write. If
this flag is false, value log file will be synced only at
file rotation. Sync method provides ability to sync data
whenever required.

Fixes #749

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/750)

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -1697,6 +1697,59 @@ func ExampleTxn_NewIterator() {
 	// Counted 1000 elements
 }
 
+func TestSyncForRace(t *testing.T) {
+	dir, err := ioutil.TempDir("", "badger")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	opts := DefaultOptions
+	opts.Dir = dir
+	opts.ValueDir = dir
+	opts.SyncWrites = false
+
+	db, err := Open(opts)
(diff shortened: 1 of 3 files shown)


db_test.go
```diff
@@ -1697,6 +1697,59 @@ func ExampleTxn_NewIterator() {
...
+	opts := DefaultOptions
+	opts.Dir = dir
+	opts.ValueDir = dir
+	opts.SyncWrites = false
+
+	db, err := Open(opts)
+	require.NoError(t, err)
+	defer db.Close()
+
+	closeChan := make(chan struct{})
```
mangalaman93: use WaitGroup instead here.

PR #761, opened 2019-04-10
Re-wrap comments in txn.go to 80 columns

jarifibrahim:
Signed-off-by: Ibrahim Jarif <ibrahim@dgraph.io>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/761)

### DIFF
--- a/txn.go
+++ b/txn.go
@@ -33,7 +33,8 @@ import (
 )
 
 type oracle struct {
-	// A 64-bit integer must be at the top for memory alignment. See issue #311.
+	// A 64-bit integer must be at the top for memory alignment. See issue
+	// https://github.com/dgraph-io/badger/issues/311.
 	refCount  int64
 	isManaged bool // Does not change value, so no locking required.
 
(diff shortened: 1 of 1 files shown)


txn.go
```diff
@@ -168,15 +172,15 @@ func (o *oracle) newCommitTs(txn *Txn) uint64 {
 	}
 
 	var ts uint64
-	if !o.isManaged {
-		// This is the general case, when user doesn't specify the read and commit ts.
+	if o.isManaged {
```
jarifibrahim: This is the only *non comment* change.

mangalaman93:
@jarifibrahim A couple of quick comments -
 * 80 chars seems low, I think we are following 100 chars (check once). Even then, it should be taken as a guideline instead of a rule to ensure comments are readable.
 * The commit message does not explain the non-comment change that you have made. Consider breaking down the patch into two separate commits.

jarifibrahim:
Thanks @mangalaman93 and @manishrjain .
I didn't realise we are using 100 characters limits. I'll send a new PR for the minor `if` condition change.
Closing this one.

PR #762, opened 2019-04-10
Throttle writers while Batch Write and DB Restore

ashish-goswami:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/762)

### DIFF
--- a/backup.go
+++ b/backup.go
@@ -22,7 +22,6 @@ import (
 	"context"
 	"encoding/binary"
 	"io"
-	"sync"
 
 	"github.com/dgraph-io/badger/pb"
 	"github.com/dgraph-io/badger/y"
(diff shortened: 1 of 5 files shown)


batch.go
```diff
@@ -36,7 +38,18 @@ type WriteBatch struct {
...
+		txn:      db.newTransaction(true, true),
+		throttle: y.NewThrottle(100),
+	}
+}
+
+// SetMaxPendingTxns sets a limit on maximum number of pending transactions
+// while writing batches. Default limit is 100. This function should be called
+// at the initialisation of WriteBatch.
+func (wb *WriteBatch) SetMaxPendingTxns(max int) {
+	wb.throttle = y.NewThrottle(max)
```
mangalaman93: Can this leak memory by any chance? Was reading about timers leaking memory recently, seemed similar. I can't think of reason that it would though.

batch.go
```diff
@@ -47,13 +60,14 @@ func (db *DB) NewWriteBatch() *WriteBatch {
 //
 // Note that any committed writes would still go through despite calling Cancel.
 func (wb *WriteBatch) Cancel() {
-	wb.wg.Wait()
+	// TODO: what to do with error returned by finish
+	wb.throttle.Finish()
```
mangalaman93: Is there an option for logging in badger?

backup.go
```diff
@@ -140,28 +139,8 @@ func writeTo(entry *pb.KV, w io.Writer) error {
 func (db *DB) Load(r io.Reader) error {
 	br := bufio.NewReaderSize(r, 16<<10)
 	unmarshalBuf := make([]byte, 1<<10)
+	throttle := y.NewThrottle(100) // TODO: take this value as arg?
```
mangalaman93: yeah, 100 needs to be justified.

Issue #311, opened 2017-11-08
ARMv7 segmentation fault in oracle.readTs when calling loadUint64


gq0:
I am facing an issue running badger on an ARMv7 architecture. The minimal test case below works quite fine on an amd64 machine but, unfortunately, not on ARMv7 32bit.

The trace below shows that the issue originates in ``atomic.loadUint64()`` but I also run basic atomic operations tests against the golang runtime, and they work fine on this architecture. 

It looks to me that the underlying memory of ``oracle.curRead`` somehow vanishes but I am not sure.

Below you find also a strace trace. There the segmentation fault happens after the ``madvise``, but I am not sure if this is related. 

Badger version: 1.0.1 (89689ef36cae26ae094cb5ea79b7400d839f2d68)
golang version: 1.8.5 and 1.9.2

Test case:
```golang
func TestPersistentCache_DirectBadger(t *testing.T) {
	dir, err := ioutil.TempDir("", "")
	if err != nil {
		t.Fatal(err)
	}
	defer os.RemoveAll(dir)

	config := badger.DefaultOptions
	config.TableLoadingMode = options.MemoryMap
	config.ValueLogFileSize = 16 << 20
	config.LevelOneSize = 8 << 20
	config.MaxTableSize = 2 << 20
	config.Dir = dir
	config.ValueDir = dir
	config.SyncWrites = false

	db, err := badger.Open(config)
	if err != nil {
		t.Fatalf("cannot open db at location %s: %v", dir, err)
	}

	err = db.View(func(txn *badger.Txn) error { return nil })

	if err != nil {
		t.Fatal(err)
	}

	if err = db.Close(); err != nil {
		t.Fatal(err)
	}
}
```

```golang
=== RUN   TestPersistentCache_DirectBadger
--- FAIL: TestPersistentCache_DirectBadger (0.01s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x4 pc=0x1150c]

goroutine 5 [running]:
testing.tRunner.func1(0x10a793b0)
        /usr/lib/go/src/testing/testing.go:711 +0x2a0
panic(0x3e4bd8, 0x6bb478)
        /usr/lib/go/src/runtime/panic.go:491 +0x204
sync/atomic.loadUint64(0x10a483cc, 0x200000, 0x0)
[...]

PR #766, opened 2019-04-12
Force flush memtable on log file rotate

### LINKED ISSUE #549: Force flush memtable when log file is rotated

manishrjain:
Add a new option which allows a memtable to be force flushed after N log files have been rotated. This is useful in write loads where there are fewer keys and larger values. In those cases, memtables can take a long time to get filled up. But, the value log files fill up quickly and rotate. This causes the value log header in the LSM tree to not get persisted. When a crash happens, a lot of log files need to be replayed to get back to current state, causing slowdown in Open.

This PR avoids that scenario by allowing user to specify how many log files can be rotated before a memtable must be force flushed. Note that the counter gets reset every time a flush happens.

Fixes #549

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/766)

### DIFF
--- a/db.go
+++ b/db.go
@@ -65,18 +65,22 @@ type DB struct {
 	// nil if Dir and ValueDir are the same
 	valueDirGuard *directoryLockGuard
 
-	closers    closers
-	elog       trace.EventLog
-	mt         *skl.Skiplist   // Our latest (actively written) in-memory table
-	imm        []*skl.Skiplist // Add here only AFTER pushing to flushChan.
-	opt        Options
-	manifest   *manifestFile
-	lc         *levelsController
-	vlog       valueLog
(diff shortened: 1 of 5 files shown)


ashish-goswami:
@manishrjain now we are checking whether log file has been rotated inside **ensureRoomForWrite** function, which is called for every request iteration. Suppose log file has been rotated, then while iterating for the first request we will clear **LogRotatesToFlush** and store last value of head(which is the head before insertion of all requests). Ideally we should store updated head value which is after all requests has been written to LSM tree.

ashish-goswami:
> Atomic variables are generally suggested to keep in the beginning of the struct to avoid alignment issues. For example, #311

Went through #311, looks like it is valid while using 64bit functions.

Issue #684, opened 2019-01-15
Live stream Badger updates


manishrjain:
We should have a way by which we can live stream updates to Badger as they happen. There can be an option to restrict it to certain prefixes as well. This would be useful in Dgraph for giving users access to updates as they happen in the Dgraph cluster.

P.S. If we put some ads in between the updates, we can build YouTube TV. The longer the ads, the closer we get to the goal.

gedw99:
this would be really useful !

campoy:
Definitely not a P0 (nor a P1, imho) but let's see what benefits this could bring and figure out whether it makes sense to implement it as part of v2.1 or later.

gedw99:
sure.

For the Record, yes CDC and Subscriptions are very similar and doing it at Badger level first would be great underpinning.
For CDC you want to get a subscription feed of the transaction changes happening to Bager.
See Cockroach CDC for great example of the API etc.

ashish-goswami:
@gedw99 we have implemented subscription support in Badger recently. Please have a look at https://github.com/dgraph-io/badger/pull/771.

gedw99:
@ashish-goswami  thank you !  It will definitely help lots of edge cases i had with Badger.

ashish-goswami:
@gedw99 closing it for now. Feel free to open it if required.

PR #771, opened 2019-04-17
Add new API Subscribe to listen on update for the specified prefix

sch00lb0y:
subscribe API to listen on live updates #684 
Signed-off-by: பாலாஜி <rbalajis25@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/771)


publisher.go
```diff
@@ -0,0 +1,154 @@
...
+	opts        Options
+	sync.RWMutex
+}
+
+type callback func(kv *pb.KV)
+
+func newPublisher(opts Options) *publisher {
+	return &publisher{
+		subscribers: make(map[string]chan<- *pb.KV),
+		batchSize:   opts.MaxPendingSubscriberUpdates,
```
sch00lb0y: I can remove the batching logic, If there is batch size using buffered channel

sch00lb0y:
PTAL @ashish-goswami @mangalaman93

db.go
```diff
@@ -1370,3 +1375,14 @@ func (db *DB) DropPrefix(prefix []byte) error {
 	db.opt.Infof("DropPrefix done")
 	return nil
 }
+
+// Subscribe can be used watch key changes for the given key prefix
+func (db *DB) Subscribe(prefix string, cb callback) func() {
+	if cb == nil {
+		panic("callback can't be nil") // should we panic or return error?
```
mangalaman93: I would prefer returning an error. We can live without panicking.

options.go
```diff
@@ -110,6 +110,9 @@ type Options struct {
 
 	// DB-specific logger which will override the global logger.
 	Logger Logger
+
+	// Maximum pending updates for the subscribers
+	MaxPendingSubscriberUpdates int
```
mangalaman93: Name could be just `MaxPendingUpdates`

publisher.go
```diff
@@ -0,0 +1,154 @@
...
+
+	"github.com/dgraph-io/badger/pb"
+	"github.com/dgraph-io/badger/y"
+)
+
+type publisher struct {
+	subscribers map[string]chan<- *pb.KV
+	batchSize   int
+	opts        Options
+	sync.RWMutex
```
mangalaman93: This should be the first element in the struct.

publisher.go
```diff
@@ -0,0 +1,154 @@
...
+// runSubscriber spins two go rotuine, one for processing callback and another for batching the incomming
+// updates. closer will first stop further incoming updates and wait for all the updates to get consumed
+// by the subscriber's callback
+// do we have to close the subscribers while closing db? or upto user to deal?
+func (p *publisher) runSubscriber(prefix string, cb callback) *y.Closer {
+	p.Lock()
+	defer p.Unlock()
+
+	c := y.NewCloser(2)
+	listen := func(updateRecv chan *pb.KV) {
```
mangalaman93: This logic seems really complex, simplify it if possible.

sch00lb0y:
@mangalaman93 Thanks for reviewing. 

Since we're limiting maximum batch size, I don't find any reason in keeping the batch logic. So, I've removed it. (`runSubscriber` also simplified)

I've addressed all the comments. Please let me know, If anything more to be taken care :)

publisher.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+	}
+}
+
+func (p *publisher) newSubscriber(prefix []byte) (<-chan *pb.KVList, uint64) {
+	p.Lock()
+	defer p.Unlock()
+	ch := make(chan *pb.KVList, 1000)
+	// increment last ID
+	p.lastID++
```
mangalaman93: We are gonna miss allocating the id=0. Is that by design?

publisher.go
```diff
@@ -0,0 +1,142 @@
...
+	return ch, p.lastID
+}
+
+// cleanSubscribers stops all the subscribers. Ideally, It should be called while closing DB
+func (p *publisher) cleanSubscribers() {
+	p.Lock()
+	defer p.Unlock()
+	for id, s := range p.subscribers {
+		close(s.sendCh)
+		delete(p.subscribers, id)
```
mangalaman93: instead of deleting one entry at a time, we could just set the map to nil or a new map.

publisher.go
```diff
@@ -0,0 +1,142 @@
...
+listen:
+	for {
+		reqs := []*request{}
+		select {
+		case <-c.HasBeenClosed():
+			break listen
+		case r := <-p.pubCh:
+			reqs = append(reqs, r...)
+		}
+	drainer:
```
mangalaman93: I'm not sure whether this is really a good idea. It delays sending the updates, possibly blocking forever if rate of publish > rate of drain. We should potentially cap the size of updates we send.

db.go
```diff
@@ -1370,3 +1387,63 @@ func (db *DB) DropPrefix(prefix []byte) error {
 	db.opt.Infof("DropPrefix done")
 	return nil
 }
+
+// Subscribe can be used watch key changes for the given key prefix
+func (db *DB) Subscribe(prefix []byte, cb callback) (func(), error) {
+	if cb == nil {
+		return nil, errors.New("callback can't be nil")
```
mangalaman93: It is generally a good idea to declare the error once and return the same pointer every time.

sch00lb0y:
@mangalaman93 About the callback. I do not know whether giving a channel or using callback will make differences in embedded DB. But I do agree 

>  A callback will limit the things that you could do inside it or will require complex logic to deal with complex scenarios.

In gcp pub/sub, they are using callback and passing ctx to cancel the subscription.
https://github.com/GoogleCloudPlatform/golang-samples/blob/602d0b609e4c5cfdee8fada906a100062712279f/pubsub/subscriptions/main.go#L119

would love to hear input from your side before approaching this

publisher.go
sch00lb0y: writes are getting batched with a cap at `doWrites`. I think no need to drain since it is already batched. I could just push after receiving. what do you think? should I set a cap or just push.

publisher.go
sch00lb0y: I iterated it to close the channel. so that it'll block the callbackrunner loop
```
		for {
			select {
			case <-c.HasBeenClosed():
				// close the subscriber to avoid further update
				db.pub.deleteSubscriber(id)
				//stop listening and drain all the updates
				break runner
			case batch, ok := <-recvCh:
				if !ok {
					// callback reciver is closed to buffer overflow, so wait for the user to
					// close the callback runner
					<-c.HasBeenClosed()
					return
				}
				slurp(batch)
			}
```

mangalaman93:
@sch00lb0y need couple of days to respond to your comments. While, I look into those, see if it is possible to cleanup the code further, specially the labels.

publisher.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (p *publisher) publishUpdates(reqs requests) {
+	p.Lock()
+	defer func() {
+		p.Unlock()
+
+		// release all the request
+		reqs.DecrRef()
+	}()
+	for _, s := range p.subscribers {
```
mangalaman93: @manishrjain will it make sense to keep some kind of Trie to directly figure out the list of subscribers a req should go to. We can keep the Trie divided based on parts of prefixes, compared to doing it byte by byte. @sch00lb0y may be for now, add a todo here mentioning that we can find a way to quickly identify the subscribers for a prefix.

db.go
```diff
@@ -1371,3 +1388,49 @@ func (db *DB) DropPrefix(prefix []byte) error {
...
+		for {
+			select {
+			case kvs := <-recvCh:
+				batch.Kv = append(batch.Kv, kvs.Kv...)
+			default:
+				return
+			}
+		}
+	}
+	cbRunner := func(recvCh <-chan *pb.KVList, id uint64) error {
```
mangalaman93: we don't need to declare the function. We can just execute the logic as it is.

db.go
```diff
@@ -1371,3 +1388,49 @@ func (db *DB) DropPrefix(prefix []byte) error {
...
+	recvCh, id := db.pub.newSubscriber(c, prefixes...)
+	slurp := func(batch *pb.KVList) {
+		defer func() {
+			if len(batch.GetKv()) > 0 {
+				cb(batch)
+			}
+		}()
+		for {
+			select {
+			case kvs := <-recvCh:
```
mangalaman93: My previous question still stands whether draining would cause callback to never get executed. You said that there is a cap in doWrites. Please point me to that.

sch00lb0y: the cap is there for request batching so I removed the draining part in publisher goroutine  https://github.com/dgraph-io/badger/blob/master/db.go#L694 if all the DB writes points to the same prefix then there may be a chance of callback never get executed

mangalaman93: I think this only limits the size of each element in the channel, not the size of the channel. In certain high load scenarios, it is possible that channel never becomes empty and we keep increasing the size of the batch while draining (this step). Let's take a step back and help me understand the point of doing the batching of `reqs`, why not call `cb` on one `reqs`?

manishrjain: @mangalaman93 -- would be worth reading this article: https://mechanical-sympathy.blogspot.com/2011/10/smart-batching.html

publisher_test.go
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package badger
+
+import (
```
mangalaman93: The imports don't look correctly formatted. The `net/http` package is below the external imports.

publisher_test.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		time.Sleep(1 * time.Millisecond)
+		cancel()
+		for i := 0; i < 5; i++ {
+			require.Equal(t, fmt.Sprintf("value%d", i), order[i])
+		}
+	})
+}
+
+func TestPublisherBatching(t *testing.T) {
```
mangalaman93: This test is unnecessary I think. And this test will take 10 seconds to run, increasing total test running time, every time we run the tests

db.go
mangalaman93: But aren't we already doing batching, because we have group of `req` coming in? And, if I am missing something and this batching is still useful, we should have a similar cap on the size of batching as the one Balaji points to.

publisher_test.go
sch00lb0y: I've removed this import. I forgot to remove pprof

sch00lb0y:
I've removed sleep and used waitgroup for blocking. 

Sorry for the late reply. stuck between school and exams :(

I'll add bank test over this weekend

sch00lb0y:
> badger.Close() wait for running the callbacks?

Yes

About the bank test, some updates are missing due to the cap. It's working for high cap
https://github.com/dgraph-io/badger/pull/771/files#diff-ac088ec1dcd15c668ff332efe1660bceR41
```go
      select {
	case p.pubCh <- reqs:
	default:
	}
```
I can find some cap on the trial and error method to make it work. But I'm afraid that it won't work for different number of account and goroutine config. 

Should I add that bank test or not?

mangalaman93:
@sch00lb0y I will look into it more, not sure about the exact issue. Though, one context here. One of the use cases of this API would be to be to replicate the database in real time to another node/cluster (such as Cross Data Center Replication). In such a case, we should be able to ensure that all the data is correctly replicated.

mangalaman93:
@sch00lb0y Is it possible to push the incomplete (failing) bank test? We can try look into it from our end.

sch00lb0y:
@mangalaman93 updated

mangalaman93:
The test seems to pass when I run it using `badger bank test --dir d`. Am I missing something here?

sch00lb0y:
https://github.com/dgraph-io/badger/pull/771/commits/1791ea8f1d740ea75c9062196eb27a057e3157ff#diff-ac088ec1dcd15c668ff332efe1660bceL41

I increased the buffer size, that's why it's working now. test with low buffer size. 

I'm saying that there is no delivery guarantee. 

>I will look into it more, not sure about the exact issue. Though, one context here. One of the use cases of this API would be to be to replicate the database in real time to another node/cluster (such as Cross Data Center Replication). In such a case, we should be able to ensure that all the data is reliably & correctly replicated.

I assume, we can't replicate using this API. replication has a lot of corner case. like consistency, slow runner...

If we have to make it reliable, we have to avoid this select statement
```go
      select {
	case p.pubCh <- reqs:
	default:
	}
```
have to block if the buffer is full. but It'll affect the performance on high load.

mangalaman93:
This makes sense. Let me discuss this, and I also want to look into mysql, as to how it handles it.

mangalaman93:
The PR looks good, just fix one minor typo. I will get it merged then. @sch00lb0y

Issue #767, opened 2019-04-12
Deleting values and running GC doesn't reclaim space


magik6k:
I was trying to get Badger GC in go-ipfs to reclaim space, but it didn't seem to work, so I've wrote this rather basic test case to see if it works in the simple case of keys being added, then deleted, and GC ran:

https://gist.github.com/magik6k/8c379cc02b443495e4809170fb8803a9

EDIT: gist / results updated as I discovered that I was calling Delete on wrong data...

These are my (reproducible) results:
DiscardRatio=0.5
```
non-deletable put: 101 MB(101 MB); sst: 0 B; vlog: 101 MB
data put: 3.1 GB(3.1 GB); sst: 0 B; vlog: 3.1 GB
put closed: 3.1 GB(3.1 GB); sst: 20 MB; vlog: 3.1 GB
del-open: 3.1 GB(3.1 GB); sst: 20 MB; vlog: 3.1 GB
after del: 3.2 GB(3.2 GB); sst: 20 MB; vlog: 3.1 GB
gc-step: 3.2 GB(3.2 GB); sst: 20 MB; vlog: 3.1 GB
gc: 3.2 GB(3.2 GB); sst: 20 MB; vlog: 3.1 GB
close-gc: 3.2 GB(3.2 GB); sst: 36 MB; vlog: 3.1 GB // expected ~100-120M
```

Cases below are wrong because of a bug in my code:
<s>
DiscardRatio=0.01
```
non-deletable put: 101 MB(101 MB); sst: 0 B; vlog: 101 MB //put 100M, looks good
data put: 3.1 GB(3.1 GB); sst: 0 B; vlog: 3.1 GB // put 3G, looks good
put closed: 3.1 GB(3.1 GB); sst: 20 MB; vlog: 3.1 GB // close DB, looks good
del-open: 3.1 GB(3.1 GB); sst: 20 MB; vlog: 3.1 GB // reopen db, looks good
after del: 9.1 GB(9.1 GB); sst: 3.0 GB; vlog: 6.1 GB // delete the 3G keys put before, sst explodes, vlog grows by 3G
gc-step: 9.1 GB(9.1 GB); sst: 3.0 GB; vlog: 6.1 GB // call RunValueLogGC, nothing changes
gc: 9.1 GB(9.1 GB); sst: 3.0 GB; vlog: 6.1 GB // ErrNoRewrite, even with 3 mostly rewritable vlogs
[...]

PR #778, opened 2019-04-23
Set write batch readTs to the readMark.DoneUntil() value

jarifibrahim:
Every Transaction stores the latest value of `readTs` it is aware of. When the transaction is discarded (which happens even when we commit), the global value of `readMark` is updated.
https://github.com/dgraph-io/badger/blob/1fcc96ecdb66d221df85cddec186b6ac7b6dab4b/txn.go#L501-L503
Previously, the `readTs` of transaction inside the write batch struct was set to 0. So the global value of `readMark` would also be set to 0 (unless someone ran a transaction after using write batch).
Due to the 0 value of the global `readMark`, the compaction algorithm would skip all the values which were inserted in the write batch call. 
https://github.com/dgraph-io/badger/blob/1fcc96ecdb66d221df85cddec186b6ac7b6dab4b/levels.go#L480-L484
and
https://github.com/dgraph-io/badger/blob/1fcc96ecdb66d221df85cddec186b6ac7b6dab4b/txn.go#L138-L145
The `o.readMark.DoneUntil()` call would always return `0` and so the compaction wouldn't compact the newer values.

With this commit, the compaction algorithm works as expected with key-values inserted via Transaction API or via the Write Batch API.

Fixes https://github.com/dgraph-io/badger/issues/767

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/778)

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -36,7 +36,14 @@ type WriteBatch struct {
 // creating and committing transactions. Due to the nature of SSI guaratees provided by Badger,
 // blind writes can never encounter transaction conflicts (ErrConflict).
 func (db *DB) NewWriteBatch() *WriteBatch {
-	return &WriteBatch{db: db, txn: db.newTransaction(true, true)}
+	txn := db.newTransaction(true, true)
+	// If we let it stay at zero, compactions would not allow older key versions to be deleted,
+	// because the read timestamps of pending txns, would be zero. Therefore, we set it to the
+	// maximum read timestamp that's done. This allows compactions to discard key versions below
+	// this read timestamp, while also not blocking on pending txns to finish before starting this
+	// one.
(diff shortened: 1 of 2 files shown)


mangalaman93:
@jarifibrahim will it now affect the performance as @manishrjain mentioned? How are we solving that? We should probably add comments around it in the PR.

jarifibrahim:
> @jarifibrahim will it now affect the performance as @manishrjain mentioned? How are we solving that? We should probably add comments around it in the PR.

@mangalaman93 No, this change shouldn't affect performance. My previous changes would've affected the performance but this one shouldn't.

ashish-goswami:
@jarifibrahim can you check why unit tests are failing on Teamcity?

PR #787, opened 2019-04-29
Fix vlog.sync()

### LINKED ISSUE #783: Badger crashes on open if last saved head is not pointing to latest value log file
**Way to reproduce:**
1. Run **fill** command for inserting large number of entries(100M) then kill it after some time. Try to run it again.
2. Change **Close** method of **DB** to push head with {0 0 0} value in **flushTask**.

**Reason**
Currently after flushing a Memtable, vlog.sync() is called. sync() tries to sync latest value log file(file with maximum ID). After restart, badger tries to replay all the entries from last saved head. If head is not pointing to latest file, replay will start from here and latest file wont be opened till then. If Memtable is flushed during this time, calling sync() will return error. This will create some assertion failure.

ashish-goswami:
vlog.sync() syncs only latest value log file(vlog.maxFid). There can
be case when vlog.maxFid is not opened and vlog.sync() is called(during
replay when head is not pointing to vlog.maxFid). Hence it will return
error. This causes assertion y.AssertTrue(ft.mt == db.imm[0]) in db.go
to fail, resulting in crash of badger. This commit fixes it by accepting
an argument fid in vlog.sync method. If fid < vlog.maxFid, it returns
without doing anything.

Fixes #783 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/787)

### DIFF
--- a/db.go
+++ b/db.go
@@ -433,7 +433,7 @@ const (
 // Sync syncs database content to disk. This function provides
 // more control to user to sync data whenever required.
 func (db *DB) Sync() error {
-	return db.vlog.sync()
+	return db.vlog.sync(math.MaxUint32)
 }
 
 // When you create or delete a file, you have to ensure the directory entry for the file is synced
@@ -770,7 +770,7 @@ func (db *DB) ensureRoomForWrite() error {
 	case db.flushChan <- flushTask{mt: db.mt, vptr: db.vhead}:
 		db.elog.Printf("Flushing value log to disk if async mode.")
 		// Ensure value log is synced to disk so this memtable's contents wouldn't be lost.
-		err = db.vlog.sync()
+		err = db.vlog.sync(db.vhead.Fid)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


test.sh
```diff
@@ -23,9 +23,9 @@ rm -R p || true
 echo
 echo "==> Starting tests with value log mmapped..."
 sleep 5
-go test -v --vlog_mmap=true -race ./...
+go test -v --vlog_mmap=true -race -timeout 900s ./...
```
jarifibrahim: @ashish-goswami why do we need to increase the timeout? Is there a test that doesn't complete in 10 minutes (default timeout) ?

jarifibrahim:
@ashish-goswami This PR would fix https://github.com/dgraph-io/badger/issues/783, isn't it?

test.sh
ashish-goswami: Once my build failed with message `tests ran too long` and hence changed it. I have removed the commit now. It was just for debugging.

PR #796, opened 2019-05-03
Make getting key count optional in db.Tables()

ashish-goswami:
Currently db.Tables() returns list of TableInfo objects. These objects
also contain keys count for the tables. Getting keys count for tables
can be expensive as it requires iterating all the tables, espcially in
large databases. db.KeySplits function(used for steaming) also uses the
same function, which does not require keys count for tables. Hence
in case of large databases, stream can take lot of time just for getting
key splits. This commit adds support for getting keys count optionally
in TableInfo objects.

I compared time for KeySplits() in with and without keys count case.
Environment:
* Ubuntu 18.10 system with 16 core, 32 GB RAM and SSD storage
* Created DB using fill command containing 500M records and 473 SSTs

I tried to create stream on DB created above (stream internally uses
db.KeySplits()). Time taken by db.KeySplits():
with keys count: ~49 sec
without keys count: ~500 microsec

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/796)

### DIFF
--- a/badger/cmd/info.go
+++ b/badger/cmd/info.go
@@ -223,7 +223,8 @@ func dur(src, dst time.Time) string {
 }
 
 func tableInfo(dir, valueDir string, db *badger.DB) {
-	tables := db.Tables()
+	// we want all tables with keys count here.
+	tables := db.Tables(true)
 	fmt.Println()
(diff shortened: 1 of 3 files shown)


campoy:
In the future, please make sure that any breaking changes to the API or data format of backups or database storage is signaled with `[BREAKING CHANGE]` on the title.

Issue #790, opened 2019-05-02
Improve Badger startup time


jarifibrahim:
We should improve the startup time on a big badger directory.
Right now, it needs to read through all the tables. Instead, we should do it so that we move the table boundaries to MANIFEST file.
Then, we can use that to create our table.Table objects.
Then, on demand, we'd load up tables as necessary (or do it in the background)

jarifibrahim:
I did some simple benchmarks and it looks like we can gain significant performance improvement by removing the `t.LoadToRAM()` function call from
https://github.com/dgraph-io/badger/blob/3374ec21baac7d10ae162c694dbfddf670c7927e/table/table.go#L144-L147

Setup:
1. Copy and apply the following diff to master branch using `git apply -v <patch file>` 
```diff
diff --git a/db_test.go b/db_test.go
index 1d66b09..d6cb860 100644
--- a/db_test.go
+++ b/db_test.go
@@ -1736,3 +1736,13 @@ func TestMain(m *testing.M) {
 	}()
 	os.Exit(m.Run())
 }
+
+func TestDBStartup(t *testing.T) {
+	dir := "/home/ibrahim/m2_drive/500m"
+	opt := DefaultOptions
+	opt.Dir = dir
+	opt.ValueDir = dir
+	db, err := Open(opt)
+	require.NoError(t, err, "unable to open DB")
+	require.NoError(t, db.Close())
+}
diff --git a/table/table.go b/table/table.go
index 9650c08..ce80797 100644
--- a/table/table.go
+++ b/table/table.go
@@ -17,7 +17,6 @@
 package table
 
 import (
-	"bytes"
 	"crypto/sha256"
 	"encoding/binary"
 	"fmt"
@@ -142,17 +141,17 @@ func OpenTable(fd *os.File, mode options.FileLoadingMode, cksum []byte) (*Table,
 	t.tableSize = int(fileInfo.Size())
 
 	// We first load to RAM, so we can read the index and do checksum.
-	if err := t.loadToRAM(); err != nil {
-		return nil, err
-	}
+	// if err := t.loadToRAM(); err != nil {
+	// 	return nil, err
+	// }
 	// Enforce checksum before we read index. Otherwise, if the file was
 	// truncated, we'd end up with panics in readIndex.
-	if len(cksum) > 0 && !bytes.Equal(t.Checksum, cksum) {
-		return nil, fmt.Errorf(
[...]

PR #805, opened 2019-05-08
Store table boundaries in manifest file 

jarifibrahim:
With this commit, we store the table boundaries (smaller and biggest key) in the manifest file. The `OpenTable()` function accepts `Smallest` and `Biggest` as params but if `Smallest` and `Biggest` are `nil`, we figure out the smallest and the biggest on the fly (by iterating).

Related to https://github.com/dgraph-io/badger/issues/790

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/805)

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -69,6 +69,8 @@ type levelManifest struct {
 type TableManifest struct {
 	Level    uint8
 	Checksum []byte
+	Smallest []byte
+	Biggest  []byte
 }
 
 // manifestFile holds the file pointer (and other info) about the manifest file, which is a log
@@ -99,7 +101,8 @@ const (
 func (m *Manifest) asChanges() []*pb.ManifestChange {
 	changes := make([]*pb.ManifestChange, 0, len(m.Tables))
 	for id, tm := range m.Tables {
-		changes = append(changes, newCreateChange(id, int(tm.Level), tm.Checksum))
+		changes = append(changes, newCreateChange(
+			id, int(tm.Level), tm.Checksum, tm.Smallest, tm.Biggest))
 	}
 	return changes
 }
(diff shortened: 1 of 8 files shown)


manifest.go
```diff
@@ -69,6 +69,8 @@ type levelManifest struct {
 type TableManifest struct {
 	Level    uint8
 	Checksum []byte
+	Smallest []byte
```
mangalaman93: Do we have a limit on size of the key?

table/table.go
```diff
@@ -117,7 +117,8 @@ func (b block) NewIterator() *blockIterator {
 // entry.  Returns a table with one reference count on it (decrementing which may delete the file!
 // -- consider t.Close() instead).  The fd has to writeable because we call Truncate on it before
 // deleting.
-func OpenTable(fd *os.File, mode options.FileLoadingMode, cksum []byte) (*Table, error) {
+func OpenTable(
+	fd *os.File, mode options.FileLoadingMode, cksum, smallest, biggest []byte) (*Table, error) {
```
mangalaman93: Break it like this instead -
`func OpenTable(fd *os.File, mode options.FileLoadingMode, cksum, smallest, biggest []byte) (
*Table, error) {`

table/table.go
```diff
@@ -157,18 +160,24 @@ func OpenTable(fd *os.File, mode options.FileLoadingMode, cksum []byte) (*Table,
...
-	if it.Valid() {
-		t.smallest = it.Key()
+	if t.smallest == nil {
+		it := t.NewIterator(false)
+		defer it.Close()
+		it.Rewind()
+		if it.Valid() {
+			t.smallest = it.Key()
+		}
+
```
mangalaman93: unnecessary newline

PR #807, opened 2019-05-08
Change Stream.Backup to marshal pb.list

ashish-goswami:
Backup tool uses Stream to read DB entries and writing it to a
file. It implements Stream.Send to write entry one by one. This
commit modifies Stream.Send to write whole batch of entries in
one go by marshaling pb.list instead of pb.KV. This improves
backup speed by ~10x.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/807)

### DIFF
--- a/backup.go
+++ b/backup.go
@@ -105,11 +105,8 @@ func (stream *Stream) Backup(w io.Writer, since uint64) (uint64, error) {
 			if maxVersion < kv.Version {
 				maxVersion = kv.Version
 			}
-			if err := writeTo(kv, w); err != nil {
-				return err
-			}
 		}
-		return nil
+		return writeTo(list, w)
 	}
 
 	if err := stream.Orchestrate(context.Background()); err != nil {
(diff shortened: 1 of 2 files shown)


campoy:
I'm reviewing all the PRs merged since our latest release and checking whether they impact our users.

This one seems to be quite serious, as it means that someone upgrading to v2.0.0 will not be able to restore backups done with previous versions

Is that really the case? And if so, should we provide a tool to transform old backups into the new format?

connorgorman:
@campoy Any update on the possibility of a tool? Right now, backups taken on 1.5 are not compatible with 1.6 and 2.0.0. This seems less ideal and also pretty confusing as a 0.x change typically just implies an API change as opposed to a breaking format change

campoy:
Yup @connorgorman !

We're working on it and it will be released with badger v2 sometime this month :)

Feel free to add any requirements or wishes about the migration tool here: #966

PR #809, opened 2019-05-13
Use batchSetAsync instead of transaction in merge compactions

jarifibrahim:
When merge compactions run very frequently, there might be multiple transaction conflicts. With this PR, we use batchSetAsync instead of transaction API to insert new discard entries. This allows us to insert into DB while the merge requests are coming in.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/809)


manishrjain:
The question is, why are we running so frequently?

jarifibrahim:
> The question is, why are we running so frequently?

I had made the change so that the `discard entry` added by the `merge compaction` can be removed by the main compactions. The issue was that if the `discard entry` was added by the `m.stop()` call https://github.com/dgraph-io/badger/blob/master/merge_test.go#L134 then it would have a `ts` > `discardTs` and the entries would never be removed

But if the `discard entry` was added by the periodic merge compactions, it would have a `ts` <= `discardTs` and then all the older entries would be removed.

jarifibrahim:
Thanks @manishrjain. I've made the requested changes.

merge.go
```diff
@@ -73,51 +75,48 @@ func (op *MergeOperator) iterateAndMerge(txn *Txn) (val []byte, err error) {
...
+	val, version, err := op.iterateAndMerge()
 	if err == ErrKeyNotFound || err == errNoMerge {
-		// pass.
+		return nil
 	} else if err != nil {
 		return err
 	}
-	return nil
+	entries := []*Entry{
+		&Entry{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

jarifibrahim: Done

Issue #817, opened 2019-05-20
When truncating an entire vlog, it should also get deleted.


danielmai:
I saw this issue with Dgraph v1.0.15-rc7.

```
   2019/05/20 20:17:55 Error while creating badger KV posting store error: invalid argument
Unable to map file
github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger/y.Wrapf
        /ext-go/1/src/github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger/y/error.go:82
github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger.(*logFile).openReadOnly
        /ext-go/1/src/github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger/value.go:90
github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger.(*valueLog).open
        /ext-go/1/src/github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger/value.go:794
github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger.Open
        /ext-go/1/src/github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger/db.go:301
github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger.OpenManaged
        /ext-go/1/src/github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger/managed_db.go:26
github.com/dgraph-io/dgraph/edgraph.(*ServerState).initStorage
        /ext-go/1/src/github.com/dgraph-io/dgraph/edgraph/server.go:172
github.com/dgraph-io/dgraph/edgraph.InitServerState
        /ext-go/1/src/github.com/dgraph-io/dgraph/edgraph/server.go:77
github.com/dgraph-io/dgraph/dgraph/cmd/alpha.run
        /ext-go/1/src/github.com/dgraph-io/dgraph/dgraph/cmd/alpha/run.go:482
github.com/dgraph-io/dgraph/dgraph/cmd/alpha.init.1.func1
        /ext-go/1/src/github.com/dgraph-io/dgraph/dgraph/cmd/alpha/run.go:82
github.com/dgraph-io/dgraph/vendor/github.com/spf13/cobra.(*Command).execute
        /ext-go/1/src/github.com/dgraph-io/dgraph/vendor/github.com/spf13/cobra/command.go:702
[...]

PR #823, opened 2019-05-22
Delete vlog file instead of truncating it completely

jarifibrahim:
With this commit, we delete the vlog file which is completely corrupted.
Earlier we would truncate such vlog file and that meant we had vlog
files with size 0 and mmap would complain about it.

Fixes https://github.com/dgraph-io/badger/issues/817

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/823)

### DIFF
--- a/value.go
+++ b/value.go
@@ -726,8 +726,10 @@ func (vlog *valueLog) replayLog(lf *logFile, offset uint32, replayFn logEntry) e
 	if err != nil {
 		return errFile(err, lf.path, "Unable to replay logfile")
 	}
-	// The entire file should be truncated (i.e. it should be deleted)
-	if endOffset == 0 {
+	// The entire file should be truncated (i.e. it should be deleted).
+	// If fid == maxFid then it's okay to truncate the entire file since will be
+	// used for future additions
+	if endOffset == 0 && lf.fid != vlog.maxFid {
 		return errDeleteVlogFile
 	}
 	if int64(endOffset) == fi.Size() {
@@ -792,11 +794,8 @@ func (vlog *valueLog) open(db *DB, ptr valuePointer, replayFn logEntry) error {
 		// Replay and possible truncation done. Now we can open the file as per
 		// user specified options.
 		if err := vlog.replayLog(lf, offset, replayFn); err != nil {
-			if err != errDeleteVlogFile {
-				return err
-			}
 			// Log file is corrupted. Delete it
-			if vlog.maxFid != fid {
+			if err == errDeleteVlogFile {
 				delete(vlog.filesMap, fid)
 				path := vlog.fpath(lf.fid)
 				if err := os.Remove(path); err != nil {
@@ -804,6 +803,7 @@ func (vlog *valueLog) open(db *DB, ptr valuePointer, replayFn logEntry) error {
 				}
 				continue
 			}
+			return err
 		}
 		vlog.db.opt.Infof("Replay took: %s\n", time.Since(now))
 
--- a/value.go
+++ b/value.go
@@ -88,7 +88,7 @@ func (lf *logFile) openReadOnly() error {
 
 	if err = lf.mmap(fi.Size()); err != nil {
 		_ = lf.fd.Close()
-		return y.Wrapf(err, "Unable to map file")
+		return y.Wrapf(err, "Unable to map file: %q", fi.Name())
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


value.go
```diff
@@ -725,6 +726,10 @@ func (vlog *valueLog) replayLog(lf *logFile, offset uint32, replayFn logEntry) e
 	if err != nil {
 		return errFile(err, lf.path, "Unable to replay logfile")
 	}
+	// The entire file should be truncated (i.e. it should be deleted)
+	if endOffset == 0 {
+		return errDeleteVlogFile
```
ashish-goswami: Once this error is returned, we delete the vlog file. We should return this error once we check DB was opened with vlog.opt.Truncate true.

value_test.go
```diff
@@ -935,3 +935,40 @@ func BenchmarkReadWrite(b *testing.B) {
...
+	opts := DefaultOptions
+	opts.Dir = dir
+	opts.ValueDir = dir
+	opts.Truncate = true
+
+	db, err := Open(opts)
+	require.NoError(t, err)
+	// Insert 1 entry so that we have valid data in first vlog file
+	require.NoError(t, db.Update(func(txn *Txn) error {
+		txn.Set([]byte("foo"), nil)
```
golangcibot: Error return value of `txn.Set` is not checked (from `errcheck`)

jarifibrahim: Fixed

value.go
jarifibrahim: Yeah. Makes sense. Fixed in https://github.com/dgraph-io/badger/pull/823/commits/7104d7d8aedf3eee6c3a07c5e7ca841e8c6bfa1d

value_test.go
```diff
@@ -935,3 +935,51 @@ func BenchmarkReadWrite(b *testing.B) {
 		}
 	}
 }
+
+// Regression test for https://github.com/dgraph-io/badger/issues/817
+func TestValueLogTruncate(t *testing.T) {
+	dir, err := ioutil.TempDir(".", "badger-test")
+	y.Check(err)
```
ashish-goswami: require.NoError would be nice here.

value_test.go
```diff
@@ -935,3 +935,51 @@ func BenchmarkReadWrite(b *testing.B) {
...
+	opts := DefaultOptions
+	opts.Dir = dir
+	opts.ValueDir = dir
+	opts.Truncate = true
+
+	db, err := Open(opts)
+	require.NoError(t, err)
+	// Insert 1 entry so that we have valid data in first vlog file
+	require.NoError(t, db.Update(func(txn *Txn) error {
+		require.NoError(t, txn.Set([]byte("foo"), nil))
```
ashish-goswami: you can simply return `txn.Set([]byte("foo", nil)`

value_test.go
```diff
@@ -935,3 +935,51 @@ func BenchmarkReadWrite(b *testing.B) {
...
+		return nil
+	}))
+
+	fileCountBeforeCorruption := len(db.vlog.filesMap)
+	require.NoError(t, err)
```
ashish-goswami: what is this error?

value_test.go
```diff
@@ -935,3 +935,51 @@ func BenchmarkReadWrite(b *testing.B) {
...
+	zeroFile, ok := db.vlog.filesMap[2]
+	require.True(t, ok)
+	fileStat, err := zeroFile.fd.Stat()
+	require.NoError(t, err)
+	require.Zero(t, fileStat.Size())
+
+	// -1 because the file with id=2 will be completely truncated. It won't be deleted.
+	// There would be two files. fid=0 with valid data, fid=2 with zero data (truncated).
+	fileCountAfterCorruption := len(db.vlog.filesMap) - 1
+	require.Equal(t, fileCountBeforeCorruption, fileCountAfterCorruption)
```
ashish-goswami:
```suggestion
	require.Equal(t, fileCountBeforeCorruption+1, fileCountAfterCorruption)
```
is more clear to understand. Instead of comparing fileCountBeforeCorruption to fileCountAfterCorruption-1

value_test.go
ashish-goswami: Also check if it is possible to wrap this in `runBadgerTest`, which takes care of creating directory and initializing DB.

value_test.go
jarifibrahim: Fixed

value_test.go
jarifibrahim: Removed it. Wasn't needed.

value_test.go
jarifibrahim: Thanks! Changed it.

value_test.go
jarifibrahim: Using `runBadgerTest` wouldn't be useful since we want to manually open and close DB twice.

PR #824, opened 2019-05-22
Fix a bug in StreamWriter which causes key overlap in levels.

manishrjain:
Badger currently put the `!badger!head` key in one of the tables at the last level that StreamWriter writes to.  This can cause issues if any other table in that level has an overlapping key range, considering this key did not come from the Stream Framework iteration.

This PR clears out level 1 for this head key. All other keys start from level 2. At the end, this key is written to level 1, so there's no chance of overlap.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/824)

### DIFF
--- a/stream_writer_test.go
+++ b/stream_writer_test.go
@@ -249,3 +249,47 @@ func TestStreamWriter4(t *testing.T) {
 		require.NoError(t, sw.Flush(), "sw.Flush() failed")
 	})
 }
+
+func TestStreamWriter5(t *testing.T) {
+	runBadgerTest(t, nil, func(t *testing.T, db *DB) {
+		list := &pb.KVList{}
+
+		left := make([]byte, 6)
+		left[0] = 0x00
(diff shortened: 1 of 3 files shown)


stream_writer_test.go
```diff
@@ -249,3 +249,47 @@ func TestStreamWriter4(t *testing.T) {
...
+		// require.NoError(t, err)
+
+		sw := db.NewStreamWriter()
+		require.NoError(t, sw.Prepare(), "sw.Prepare() failed")
+		require.NoError(t, sw.Write(list), "sw.Write() failed")
+		require.NoError(t, sw.Flush(), "sw.Flush() failed")
+		require.NoError(t, db.Close())
+
+		var err error
+		db, err = Open(db.opt)
```
golangcibot: SA4006: this value of `db` is never used (from `staticcheck`)

Issue #830, opened 2019-05-27
db.Open returns "unexpected end of JSON input"


mjentsch:
on 3 of 4 servers. Any idea where this may came from? Databases are ~5TiB.

ashish-goswami:
@mjentsch What is the Badger version? Also, tell us the system configuration on which Badger is running.

mjentsch:
Current master. Server are CentOS 7 x86_64, Go version is 1.11.5.
What else would be useful to know?

ashish-goswami:
I wanted to know about RAM, CPU and disks info. Just to confirm, 5TB is size of single Badger instance?

mjentsch:
5TiB is the first of 2 DB dirs and fails to be opened. We have small keys and relative large values.
32G RAM,  8 core, 20TiB filesystem where Badger dirs are located, 4TiB free. Nothing else beside OS and the DB prog running on that box.

mjentsch:
Using ./badger info on such DB ends with

[Summary]
Level 0 size:        64 MB
Level 1 size:       252 MB
Level 2 size:       2.6 GB
Level 3 size:       267 MB
Total index size:   3.2 GB
Value log size:     6.1 TB

Abnormalities: None.
0 extra files.
0 missing files.
0 empty files.
0 truncated manifests.
badger 2019/05/27 15:09:56 INFO: 28 tables out of 51 opened in 3.045s
badger 2019/05/27 15:09:59 INFO: All 51 tables opened in 5.722s
Error: failed to open database: unexpected end of JSON input

mjentsch:
I added a line in value.go

        if err := json.Unmarshal(vs.Value, &statsMap); err != nil {
                vlog.opt.Debugf("error unmarshal: %v", statsMap)
                return err
        }

and the compiled badger binary ends with

[Summary]
Level 0 size:        64 MB
Level 1 size:       252 MB
Level 2 size:       2.6 GB
Level 3 size:       267 MB
Total index size:   3.2 GB
Value log size:     6.1 TB

Abnormalities: None.
0 extra files.
0 missing files.
0 empty files.
0 truncated manifests.
badger 2019/05/27 15:43:11 INFO: 23 tables out of 51 opened in 3.103s
badger 2019/05/27 15:43:14 INFO: 47 tables out of 51 opened in 6.113s
badger 2019/05/27 15:43:14 INFO: All 51 tables opened in 6.333s
[...]

PR #835, opened 2019-05-29
Drop discard stat if we cannot unmarshal it

jarifibrahim:
Under certain circumstances, the discard stat map stored in the SST
could get corrupted. In this case the DB would fail to start. We should
be able to start the DB even when the discard stats are corrupted.

With this commit, we do not end badger with failure when we cannot read
the discard stats from the file. We set the discard stats to empty value
since the discard stats will be rebuilt during compactions.

Fixes https://github.com/dgraph-io/badger/issues/830

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/835)

### DIFF
--- a/value.go
+++ b/value.go
@@ -1388,7 +1388,11 @@ func (vlog *valueLog) populateDiscardStats() error {
 
 	var statsMap map[uint32]int64
 	if err := json.Unmarshal(vs.Value, &statsMap); err != nil {
-		return err
+		// It's safe to ignore the error since the discard stat might have been corrupted and
+		// it would be rebuilt during compactions.
+		vlog.opt.Debugf("Failed to unmarshal: %s. Dropping Value Log Discard stats.", err)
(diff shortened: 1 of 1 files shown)


value.go
```diff
@@ -1388,7 +1388,11 @@ func (vlog *valueLog) populateDiscardStats() error {
 
 	var statsMap map[uint32]int64
 	if err := json.Unmarshal(vs.Value, &statsMap); err != nil {
-		return err
```
ashish-goswami: now we are not returning error from here and this function is used internally. we can remove error from its return type.

jarifibrahim: https://github.com/dgraph-io/badger/blob/e9447c910efd3c67c5453ea1f65d2f355544dd82/value.go#L1376-L1396

We return error from the `db.get()` call.

jarifibrahim:
This PR doesn't fix the underlying issue, see https://github.com/dgraph-io/badger/issues/830#issuecomment-497386681 for details. Closing this PR.

PR #840, opened 2019-05-31
Making the loader API public

gitlw:
so that the cross data center replication can make use of the loader api for populating data into badger.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/840)


campoy:
This has been reverted by #887

PR #841, opened 2019-05-31
Merge function should process items in order

jarifibrahim:
With this commit, merge operator processes the items in the order of which they were inserted. Earlier it would be processed by the merge function in reverse order.

For instance, if `A` , `B`, `C` were inserted, and merge function was simple `append` operation, they would be passed to the merge function as
```
mergeFunc(C, B) => result CB
mergeFunc(CB, A) => result CBA
```

With this change, they're passed to the merge function as
```
mergeFunc(B, C) => result BC
mergeFunc(A, BC) => result ABC
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/841)

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,26 +278,16 @@ for {
 ```
 
 ### Merge Operations
-Badger provides support for unordered merge operations. You can define a func
+Badger provides support for ordered merge operations. You can define a func
 of type `MergeFunc` which takes in an existing value, and a value to be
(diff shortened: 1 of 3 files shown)


campoy:
I'm reviewing all the PRs merged since our latest release and checking whether they impact our users.
I'd say this one doesn't, please let me know if that's not the case.

PR #858, opened 2019-06-10
Store discardMap in vlog file when it's too large

jarifibrahim:
The discardMap stores the discard status for all value log files in the
db and this could get too large when there are many value log files in
badger. When there are many vlog files, the size of discard map could be
more than what we can keep in the LSM tree.

With this commit, we insert the discardMap into badger using the write
channel which would automatically put the discard map into the value log
file and use it's pointer in the LSM tree if the LSM tree cannot hold it.

Fixes https://github.com/dgraph-io/badger/issues/830

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/858)

### DIFF
--- a/db.go
+++ b/db.go
@@ -354,7 +354,9 @@ func (db *DB) Close() error {
 func (db *DB) close() (err error) {
 	db.elog.Printf("Closing database")
 
-	db.vlog.FlushDiscardStats()
+	if err := db.vlog.FlushDiscardStats(); err != nil {
+		return errors.Wrap(err, "failed to flush discard stats")
+	}
 
 	atomic.StoreInt32(&db.blockWrites, 1)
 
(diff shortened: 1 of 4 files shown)


db.go
```diff
@@ -353,6 +353,9 @@ func (db *DB) Close() error {
 
 func (db *DB) close() (err error) {
 	db.elog.Printf("Closing database")
+
+	db.vlog.FlushDiscardStats()
```
golangcibot: Error return value of `db.vlog.FlushDiscardStats` is not checked (from `errcheck`)

campoy:
@manishrjain said he was leaving final approval to @martinmr, so it seems like this is ready to be merged

Could you merge it? Thanks!

Issue #769, opened 2019-04-15
TestSyncForRace test fails on windows with Access is Denied error


jarifibrahim:
Running the `TestSyncForRace` test
```
 go test -timeout 30s github.com/dgraph-io/badger -run TestSyncForRace -v
```
on a windows machine fails with
```
--- FAIL: TestSyncForRace (8.50s)
    db_test.go:1759: 
        	Error Trace:	db_test.go:1759
        	            				asm_amd64.s:1333
        	Error:      	Received unexpected error:
        	            	sync badger-test974813326\000000.vlog: Access is denied.
        	Test:       	TestSyncForRace
```

The test works fine on linux but fails on windows.
See build logs https://ci.appveyor.com/project/manishrjain/badger/builds/23850505?fullLog=true

jarifibrahim:
The access denied error was returned by `db.Sync()` because `vlog.MaxFid` would be read to find the file which should be synced, but before the file sync could start, it would be marked as read only by `vlog.write` method.
If we try to fsync a read only file, windows returns a Access Denied error.

jarifibrahim:
This is fixed via https://github.com/dgraph-io/badger/pull/923.

PR #860, opened 2019-06-12
Acquire lock before updating maxFid

jarifibrahim:
**NOTE** - The windows build is still failing because there are multiple test failures. This PR fixes only `TestSyncForRace` test on windows.

---

This commit fixes failing TestSyncForRace test on windows. The test was
failing because of race condition on `vlog.maxFid` variable.
With this commit, we acquire lock before updating the value of maxFid.

The test `TestSyncForRace` fails on windows for 2 reasons (I don't know why none of them show up on linux)
#### 1. Deadlock in the test. 
The test code looks something like this
```go
func TestSyncForRace(t *testing.T) {
    ...
    go func() {
        for {
            select { // After a while none of the select cases are executed
            case <-ticker.C:
                if err := db.Sync(); err != nil {
                    require.NoError(t, err)
                }
            case <-closeChan:
                close(doneChan)
                return
            }
        }
    }()

    for i := 0; i < 10000; i++ {
        ...
    }
    require.NoError(t, txn.Commit())

    close(closeChan)
    <-doneChan // The test gets blocked here
}
```

Notice that we close the `closeChan` and wait for `doneChan` to close. The interesting part is that `doneChan` doesn't close most of the times and the test goes into deadlock. The simple fix was to use wait groups instead of a channel to ensure we don't end up in a deadlock.

#### 2. The `vlog Access Denied` error.
This error was returned by `db.Sync()` because `vlog.MaxFid` would be read to find the file which should be synced, but before the file sync could start, it would be marked as `read only` by `vlog.write` method.
If we try to `fsync` a read only file, windows returns a `Access Denied` error.

Fixes https://github.com/dgraph-io/badger/issues/769

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/860)


jarifibrahim:
We have multiple test failures on windows because mmap isn't working as expected. I'm looking into the failures.

jarifibrahim:
I've reverted https://github.com/dgraph-io/badger/pull/860/commits/f69b1c32e5432d4c37442ff45c74a3444c760336 because it shouldn't be part of this PR. I'll send a separate to fix Value Log truncation failure on windows.

jarifibrahim:
Closing this PR in favour of https://github.com/dgraph-io/badger/pull/923. Locking all the files isn't a great approach and we should prevent it. https://github.com/dgraph-io/badger/pull/923 prevents it by opening all files in read-write mode.

Issue #867, opened 2019-06-16
Inaccurate prompt


lengrongfu:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.12.4 darwin/amd64
</pre>

### What version of Badger are you using?
v1.5.4

### Does this issue reproduce with the latest master?
no

### What are the hardware specifications of the machine (RAM, OS, Disk)?
disk

### What did you do?

I set opts.ValueThreshold = 65520

### What did you expect to see?

I hope tips "Invalid ValueThreshold, must be lower than 65520".

### What did you see instead?
  but tips "Invalid ValueThreshold, must be lower than uint16."

PR #877, opened 2019-06-18
Use actual value of uint16-16 in ErrValueThreshold

jarifibrahim:
The ErrValueThreshold doesn't really show what should be the max value
of value threshold. This commit adds the actual value of value threshold
to the error message.

Fixes https://github.com/dgraph-io/badger/issues/867

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/877)

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -30,7 +30,7 @@ var (
 	// ErrValueThreshold is returned when ValueThreshold is set to a value close to or greater than
 	// uint16.
 	ErrValueThreshold = errors.Errorf(
-		"Invalid ValueThreshold, must be lower than %d", math.MaxUint16-16)
+		"Invalid ValueThreshold, must be less than %d", math.MaxUint16-16+1)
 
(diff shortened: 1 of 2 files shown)


errors.go
```diff
@@ -27,7 +29,8 @@ var (
 
 	// ErrValueThreshold is returned when ValueThreshold is set to a value close to or greater than
 	// uint16.
-	ErrValueThreshold = errors.New("Invalid ValueThreshold, must be lower than uint16")
+	ErrValueThreshold = errors.Errorf(
+		"Invalid ValueThreshold, must be less than %d", math.MaxUint16-16+1)
```
jarifibrahim: I've used `math.MaxUint16-16+1` because of 
https://github.com/dgraph-io/badger/blob/6d3b67d75431ba874d6658c916103070cba3c1f8/db.go#L195-L198

PR #879, opened 2019-06-18
test osx in all Go versions, not only the first one

campoy:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/879)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,10 @@ go:
     - "1.11"
     - "1.12"
 
-matrix:
-  include:
-    - os: osx
+os:
+    - linux
+    - osx
+
 notifications:
   email: false
   slack:


martinmr:
Sorry, didn't mean to approve.

campoy:
In that case, @danielmai  can you confirm on what platforms and with which Go versions we're testing our changes?

danielmai:
@campoy In TeamCity the tests run on Linux and Go 1.12.5. In TeamCity it's not so straightforward to configure multi-platform and multi-Go versions on TeamCity. If that's necessary then we can keep Travis around.

jarifibrahim:
We should keep Travis around. Doesn't hurt to run on additional platforms :grinning:

danielmai:
> Does Teamcity do the right thing? Can we remove this Travis config if we don't use it? @danielmai

We can remove this TravisCI config from Badger.

TeamCity tests Badger on Linux and (currently) with Go 1.12.4—we use the go version matching Dgraph release builds. TeamCity also runs other types of tests like a long-running bank-transaction-transfer test overnight and tests with the race detector.

campoy:
Does TeamCity test Mac and other versions of Go too, @danielmai ?

danielmai:
Right now our TeamCity set up is only Linux and a single version of Go (1.12.4).

campoy:
OK, in that case I'll close this PR and create a new issue to handle dropping Travis and using only TeamCity but including Windows and MacOS platforms plus go1.11, go1.12, and tip.

See PR #888

PR #882, opened 2019-06-20
release badger v2.0.0

### LINKED ISSUE #758: Support Go modules for v2
Could you add a `go.mod` and update imports for v2 so we can use the v2 releases release candidates with modules?

https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher

campoy:
This change releases badger v2 for Go modules.

*WARNING*:

This is not compatible with versions of Go before Go 1.11, so support for Go 1.10 is also being dropped.

Fixes #758 

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/882)


manishrjain:
Ok

manishrjain:
Didn't realize that this PR changes ALL the import paths. That doesn't seem like the right thing to do.

PR #895, opened 2019-06-24
Add options to transform backup data back and forth to a different format

martinmr:
In order to support backwards-compatible backups in Dgraph, the data
needs to be stored in some other format than just the raw key and
values. This change introduces the option to set methods to transform
the data to and forth a separate format.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/895)

### DIFF
--- a/backup.go
+++ b/backup.go
@@ -69,10 +69,28 @@ func (stream *Stream) Backup(w io.Writer, since uint64) (uint64, error) {
 				}
 			}
 
+			// Convert key and/or values to a different format if requested before writing
+			// the key-value pair.
+			keyCopy := item.KeyCopy(nil)
+			if stream.db.opt.BackupKeyFn != nil {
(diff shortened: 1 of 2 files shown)


martinmr:
Some work is still needed to use the new Options API. I'll do that once that PR is merged. However, the code in backup.go shouldn't change so it can be reviewed now.

manishrjain:
This PR still needed my approval before submission.

manishrjain:
I don't understand why this change exists here. It was something very specific that we needed in Dgraph, which somehow is being pushed all the way into options in Badger. It is not generic enough to be put here.

manishrjain:
As per our discussion, this change would be reverted.

martinmr:
PR to revert the change https://github.com/dgraph-io/badger/pull/900

PR #896, opened 2019-06-25
Removing support for Go Modules

campoy:
As explained in https://discuss.dgraph.io/t/go-modules-on-badger-and-dgraph/4662,
Go Modules force an import path rename that currently breaks support for most other
dependency management systems, including dep.

It seems an effort is being made to add support for this import path renaming by
patching the go tool itself, but until the community hasn't settled and Go Modules
is an easier option to adopt, we are simply not supporting them.

For those willing to use badger v2 from Go modules it is still very simple, you might
notice that once you set the version to be v2.0.0 a +incompatible will be added to it
indicating the dependency has not opted-in into Go Modules. That's normal and not to
be a concern.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/896)


hsanjuan:
> breaks support for most other
dependency management systems, including dep

By rewriting tags this managed to break absolutely every system now. Repeatedly pushing to existing tags was the worst option on the table. The damage is done now but please, consider not rewriting tags ever in the future. :/

PR #905, opened 2019-06-27
Adding support for Go Modules

campoy:
Adding support for Go Modules now that we have decided we will not follow semantic versioning on badger releases.

Another PR will be coming soon explaining the versioning schema of Badger.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/905)


campoy:
Exactly, Badger will support Go Modules but in a way that doesn't force Dgraph or anyone else to do so too.

Issue #907, opened 2019-06-28
Value log gc doesn't work due to too large txn


simonxw:
### What version of Go are you using (`go version`)?

<pre>
$ go version 
go1.12.1
</pre>

### What version of Badger are you using?
1.5.4

### Does this issue reproduce with the latest master?
yes

### What are the hardware specifications of the machine (RAM, OS, Disk)?
Linux, 32GB, SDD

### What did you do?
With default db options,  vlog file size 1GB, memtablesize = 64MB, keep writing keys with a certain ratio of overwrites to badger db. There are about 32 vlog files and 1GB per file.

### What did you expect to see?
vlog gc works.

### What did you see instead?

vlog gc keeps failed with error:

 vlog gc not executed    {"error": ": Txn is
 too big to fit into one request", "errorVerbose": "Txn is too big to fit into one request\ngithub.com/dgraph-io/badger.ini
t\n\t/root/go/pkg/mod/github.com/dgraph-io/badger@v1.5.4/errors.go:38\natlas/internal/pkg/badgerdb.init\n\t<autogenerated>:
1\natlas/cmd/atlas/config.init\n\t<autogenerated>:1\natlas/cmd/atlas/bootstrap.init\n\t<autogenerated>:1\natlas/cmd/atlas/c
lustersvc.init\n\t<autogenerated>:1\nmain.init\n\t<autogenerated>:1\nruntime.main\n\t/usr/lib64/go/1.11/src/runtime/proc.go
:189\nruntime.goexit\n\t/usr/lib64/go/1.11/src/runtime/asm_amd64.s:1333\n\ngithub.com/dgraph-io/badger/y.Wrap\n\t/root/go/p
kg/mod/github.com/dgraph-io/badger@v1.5.4/y/error.go:71\ngithub.com/dgraph-io/badger.(*valueLog).iterate\n\t/root/go/pkg/mo
d/github.com/dgraph-io/badger@v1.5.4/value.go:323\ngithub.com/dgraph-io/badger.(*valueLog).rewrite\n\t/root/go/pkg/mod/gith
ub.com/dgraph-io/badger@v1.5.4/value.go:411\ngithub.com/dgraph-io/badger.(*valueLog).doRunGC\n\t/root/go/pkg/mod/github.com
/dgraph-io/badger@v1.5.4/value.go:1176\ngithub.com/dgraph-io/badger.(*valueLog).runGC\n\t/root/go/pkg/mod/github.com/dgraph
[...]

PR #911, opened 2019-07-01
Ensure rewrite in vlog is within transactional limits

jarifibrahim:
With this commit the temporary list of entries built during value log
iteration is committed before it overflows transactional limits
(maxBatchSize and maxBatchCount).

Fixes https://github.com/dgraph-io/badger/issues/907

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/911)

### DIFF
--- a/value.go
+++ b/value.go
@@ -397,7 +397,7 @@ func (vlog *valueLog) rewrite(f *logFile, tr trace.Trace) error {
 			}
 
 			ne.Value = append([]byte{}, e.Value...)
-			es := int64(e.estimateSize(vlog.opt.ValueThreshold))
+			es := int64(ne.estimateSize(vlog.opt.ValueThreshold))
 			// Ensure length and size of wb is within transaction limits.
 			if int64(len(wb)+1) > vlog.opt.maxBatchCount ||
 				size+es > vlog.opt.maxBatchSize {
--- a/value.go
+++ b/value.go
@@ -397,9 +397,10 @@ func (vlog *valueLog) rewrite(f *logFile, tr trace.Trace) error {
 			}
 
 			ne.Value = append([]byte{}, e.Value...)
-			wb = append(wb, ne)
-			size += int64(e.estimateSize(vlog.opt.ValueThreshold))
-			if size >= 64*mi {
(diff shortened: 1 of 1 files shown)


value.go
```diff
@@ -397,16 +397,19 @@ func (vlog *valueLog) rewrite(f *logFile, tr trace.Trace) error {
 			}
 
 			ne.Value = append([]byte{}, e.Value...)
-			wb = append(wb, ne)
-			size += int64(e.estimateSize(vlog.opt.ValueThreshold))
-			if size >= 64*mi {
+			es := int64(e.estimateSize(vlog.opt.ValueThreshold))
```
ashish-goswami:
```suggestion
			es := int64(ne.estimateSize(vlog.opt.ValueThreshold))
```

jarifibrahim: Fixed in https://github.com/dgraph-io/badger/pull/911/commits/294f9d583bf8a5418b89c0eae7a7a0c2a91df345

jarifibrahim:
> Is there any easy way to add some unit test to verify the behavior of the code?
> 
> > _[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/911#-:-Liikx8s26zBgNEcbEfl:b3t887j)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ashish-goswami and @manishrjain)

@campoy Not really. I tried a few things but couldn't get a test that could replicate the failure. The test should be able to pick up the correct vlog file and that vlog file should have more than 64 MB of useful entries.

campoy:
Ok, I'm a bit concerned because this code is not necessarily obvious, so if you find a way to test it, even if it's not perfect, please do.

Thanks!

jarifibrahim:
> Ok, I'm a bit concerned because this code is not necessarily obvious, so if you find a way to test it, even if it's not perfect, please do.
> 
> Thanks!

I spend a lot of time yesterday trying to get a working test but it didn't work. The problem is we perform sampling inside vlog file from a random start point. The sampling should find at least some entries that can be deleted. The other prerequisite for the test is that it should have at least some amount of valid data in the same vlog file. It's tough to get both the things at the same time.

Issue #793, opened 2019-05-02
TestDropAllRace is failing on Teamcity with data race


ashish-goswami:
Line to build https://teamcity.dgraph.io/viewLog.html?buildId=11071&buildTypeId=Badger_UnitTests&tab=buildLog
Stack trace attached.
```
[13:34:02][Step 2/2] WARNING: DATA RACE
[13:34:02][Step 2/2] Write at 0x00c0007a2190 by goroutine 209:
[13:34:02][Step 2/2]   runtime.closechan()
[13:34:02][Step 2/2]       /usr/local/go/src/runtime/chan.go:327 +0x0
[13:34:02][Step 2/2]   github.com/dgraph-io/badger.(*DB).doWrites()
[13:34:02][Step 2/2]       /home/pawan0201/go/src/github.com/dgraph-io/badger/db.go:710 +0x6b1
[13:34:02][Step 2/2] 
[13:34:02][Step 2/2] Previous read at 0x00c0007a2190 by goroutine 275:
[13:34:02][Step 2/2]   runtime.chansend()
[13:34:02][Step 2/2]       /usr/local/go/src/runtime/chan.go:140 +0x0
[13:34:02][Step 2/2]   github.com/dgraph-io/badger.(*DB).sendToWriteCh()
[13:34:02][Step 2/2]       /home/pawan0201/go/src/github.com/dgraph-io/badger/db.go:660 +0x32a
[13:34:02][Step 2/2]   github.com/dgraph-io/badger.(*Txn).commitAndSend()
[13:34:02][Step 2/2]       /home/pawan0201/go/src/github.com/dgraph-io/badger/txn.go:547 +0x783
[13:34:02][Step 2/2]   github.com/dgraph-io/badger.(*Txn).CommitWith()
[13:34:02][Step 2/2]       /home/pawan0201/go/src/github.com/dgraph-io/badger/txn.go:651 +0x218
[13:34:02][Step 2/2]   github.com/dgraph-io/badger.(*Txn).CommitAt()
[13:34:02][Step 2/2]       /home/pawan0201/go/src/github.com/dgraph-io/badger/managed_db.go:56 +0xa8
[13:34:02][Step 2/2]   github.com/dgraph-io/badger.TestDropAllRace.func1()
[13:34:02][Step 2/2]       /home/pawan0201/go/src/github.com/dgraph-io/badger/managed_db_test.go:317 +0x3a2
[13:34:02][Step 2/2] 
[...]

PR #915, opened 2019-07-02
Acquire lock before sending to write channel

jarifibrahim:
Fixes https://github.com/dgraph-io/badger/issues/793

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/915)

PR #918, opened 2019-07-04
stop compaction after flushing the memtable and don't close the channel while closing the doWrite

sch00lb0y:
FIX for 2 bugs
̍1) datarace on closechannel

channel is threadsafe only for read and write but not for closing. 
Anyways we are blocking the write, So there is no need to close the channel in doWrite 
By this solution we can avoid the lock. refer #915 

2)  handleFlush is stalling on DropPrefix
It's because we are stopping the compaction before flushing the memtable. 
so l0 are not pushing to the bottom level and we're stalling indefinitely for getting space to flush the memtable to l0

So, I'm moving the stop compaction after flushing the memtable

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/918)

### DIFF
--- a/db.go
+++ b/db.go
@@ -20,7 +20,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
-	"encoding/hex"
 	"expvar"
 	"io"
 	"math"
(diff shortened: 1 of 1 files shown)


jarifibrahim:
This PR fixes https://github.com/dgraph-io/badger/issues/793

sch00lb0y:
I ran badger using the script which crashes frequently. 
ran it for 
```
166m34.215s
̍```
No panic so far

sch00lb0y:
ran the code which crash and runs badger fill in a loop.

It ran successfully for one day. So,I'm merging this

PR #925, opened 2019-07-09
[breaking/format] Use varint for header encoding

jarifibrahim:
**This is a breaking change**.

Value log stores an entry for each key-value paired stored in it. The existing format looks like

| Header (fixed size: 18 bytes) 	| Key 	| Value 	| Checksum 	|
|---------------------------	|-----	|-------	|----------	|

We don't always use up all the 18 Bytes of the header. This PR proposes varint encoding for the individual attributes of the header. The varint encoded header can have length in the range of 5-21 Bytes.

The proposed entry in value log looks like

| Header length (1 byte) 	| Header (variable size) 	| Key 	| Value 	| Checksum 	|
|------------------------	|------------------------	|-----	|-------	|----------	|

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/925)

### DIFF
--- a/db2_test.go
+++ b/db2_test.go
@@ -61,7 +61,7 @@ func TestTruncateVlogWithClose(t *testing.T) {
 
 	// Close the DB.
 	require.NoError(t, db.Close())
-	require.NoError(t, os.Truncate(path.Join(dir, "000000.vlog"), 4096))
+	require.NoError(t, os.Truncate(path.Join(dir, "000000.vlog"), 4090))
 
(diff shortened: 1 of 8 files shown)


jarifibrahim:
Adding @campoy as a reviewer since this is a breaking change.

jarifibrahim:
Closing this in favour of https://github.com/dgraph-io/badger/pull/935. https://github.com/dgraph-io/badger/pull/935 implements the same varint encoding for the header without storing its length.

Issue #926, opened 2019-07-11
panic: failed to read value pointer from vlog file


256dpi:
I just ran into this when I restarted a service a bunch of times:

```
panic: failed to read value pointer from vlog file: {Fid:143 Len:277 Offset:337322920}: Unable to find log file. Please retry
```

Since I'm running badger with `SyncWrites=false` is it possible that `RunValueLogGC` deleted a vlog that still was referenced by a persisted table?

I'm using `go1.12` and badger `v1.6.0`.

jarifibrahim:
@256dpi Can you check if your badger directory has vlog file with id `143`? Also can you share your badger directory? Did the DB crash by any chance? 

>Since I'm running badger with SyncWrites=false is it possible that RunValueLogGC deleted a vlog that still was referenced by a persisted table?

This can't be the case. We fix the reference before removing the vlog fie.
Here's how GC works
1. Find list of entries to move.
2. Insert the new entries into DB (these will be pushed to LSM tree and vlog).
3. Delete the old vlog file.

256dpi:
I wasn't able to recover any data as I purged the directory right away to get the service running again... Next time I'll do that first! Here are the logs tough:

```
I 2019-07-10T15:01:30 badger: Flushing memtable, mt.size=65796570 size of flushChan: 0
I 2019-07-10T15:01:30 badger: Storing value log head: {Fid:146 Len:49 Offset:128069992}
I 2019-07-10T15:26:44 badger: Storing value log head: {Fid:148 Len:49 Offset:15097884}
I 2019-07-10T15:26:44 badger: Flushing memtable, mt.size=67109213 size of flushChan: 0
I 2019-07-10T15:26:46 badger: Got compaction priority: {level:0 score:1 dropPrefix:[]}
I 2019-07-10T15:26:46 badger: Running for level: 0
I 2019-07-10T15:26:49 badger: LOG Compact. Added 2056568 keys. Skipped 172200 keys. Iteration took: 2.339905392s
I 2019-07-10T15:27:17 badger: LOG Compact. Added 2499284 keys. Skipped 0 keys. Iteration took: 28.577111521s
E 2019-07-10T15:28:18 panic: send on closed channel
--- CRASH => RESTART            
I 2019-07-10T15:28:22 badger: 11 tables out of 15 opened in 3.212s
I 2019-07-10T15:28:23 badger: All 15 tables opened in 4.151s
[...]

PR #929, opened 2019-07-13
Fix discard stats moved by GC bug

jarifibrahim:
The discard stats are stored like normal keys in badger, which means if
the size of the value of discard stats key is greater the value threshold it
will be stored in the value log. When value log GC happens, we insert
the same key in badger with a !badger!move prefix which points to the
new value log file. So when searching for the value of a key, if the value
points to a value log file which doesn't exist, we search for the same
key with `!badger!move` prefix.

The above logic was missing from the populateDiscardStats function. The
earlier implementation would never search for the key with `!badger!move`
prefix. It would return error on the first attempt to find the key. This
commit ensures we search for a key with prefix badger move if the value
for the existing key pointed to a value log file that no longer exists.

Fixes - https://github.com/dgraph-io/badger/issues/926

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/929)

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -36,7 +36,6 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/require"
-	"golang.org/x/net/trace"
 
 	"github.com/dgraph-io/badger/options"
 	"github.com/dgraph-io/badger/pb"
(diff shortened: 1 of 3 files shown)


connorgorman:
Will this PR be cherry-picked on top of 1.6 and released as v1.6.1? Currently without this change, v1.6.0 has an unrepairable issue.

jarifibrahim:
> Will this PR be cherry-picked on top of 1.6 and released as v1.6.1? Currently without this change, v1.6.0 has an unrepairable issue.

@campoy How should we release this fix?

PR #936, opened 2019-07-17
Drop discard stats if we can't unmarshal it

jarifibrahim:
We don't want to prevent badger from starting if we can't read/unmarshal
discard stats.

Fixes https://github.com/dgraph-io/dgraph/issues/3669

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/936)

### DIFF
--- a/value.go
+++ b/value.go
@@ -800,7 +800,9 @@ func (vlog *valueLog) open(db *DB, ptr valuePointer, replayFn logEntry) error {
 		return errFile(err, last.path, "Map log file")
 	}
 	if err := vlog.populateDiscardStats(); err != nil {
-		return err
+		// Print the error and continue. We don't want to prevent value log open if there's an error
+		// with the fetching discards stats.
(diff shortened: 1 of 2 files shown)


value.go
```diff
@@ -800,7 +800,9 @@ func (vlog *valueLog) open(db *DB, ptr valuePointer, replayFn logEntry) error {
 		return errFile(err, last.path, "Map log file")
 	}
 	if err := vlog.populateDiscardStats(); err != nil {
-		return err
+		// Print the error and continue. We don't want to prevent value log open if there's an error
+		// with the fetching discards stats.
+		db.opt.Infof("Failed to populate discard stats: %s", err)
```
ashish-goswami:
```suggestion
		db.opt.Errorf("Failed to populate discard stats: %s", err)
```

jarifibrahim: Fixed.

PR #937, opened 2019-07-22
Optimize compareKeys function

jarifibrahim:
The existing implementation would perform two comparisons if the keys
differed only in timestamps. The proposed implementation performs only
one comparison even if the keys are different.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/937)

### DIFF
--- a/y/y.go
+++ b/y/y.go
@@ -121,16 +121,18 @@ func ParseTs(key []byte) uint64 {
 	return math.MaxUint64 - binary.BigEndian.Uint64(key[len(key)-8:])
 }
 
-// CompareKeys checks the key without timestamp and checks the timestamp if keyNoTs
-// is same.
-// a<timestamp> would be sorted higher than aa<timestamp> if we use bytes.compare
-// All keys should have timestamp.
-func CompareKeys(key1, key2 []byte) int {
-	AssertTrue(len(key1) > 8 && len(key2) > 8)
-	if cmp := bytes.Compare(key1[:len(key1)-8], key2[:len(key2)-8]); cmp != 0 {
-		return cmp
+// CompareKeys checks the key with timestamp if the lengths are equal, otherwise checks only the
+// keys.
+// a<timestamp> would be sorted higher than aa<timestamp>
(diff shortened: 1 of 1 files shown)


jarifibrahim:
I compared the proposed implementation against the one in master and there isn't a conclusive performance improvement.
```go
func BenchmarkCompareKeys(b *testing.B) {
    testData := []struct {
        name  string
        key1  []byte
        key2  []byte
        equal bool
    }{
        {"equal with ts", []byte("johndoe00000003"), []byte("johndoe00000003"), true},
        {"equal with diff ts", []byte("johndoe123456429"), []byte("johndoe123456427"), false},
        {"different length", []byte("johndoe00000042"), []byte("johnddd00000042"), false},
    }

    for _, item := range testData {
        b.Run(fmt.Sprintf("%s-master", item.name), func(b *testing.B) {
            for i := 0; i < b.N; i++ {
                result := CompareKeysMaster(item.key1, item.key2)
                if item.equal {
                    require.Zero(b, result)
                } else {
                    require.NotZero(b, result)
                }
            }
        })
        b.Run(fmt.Sprintf("%s-new", item.name), func(b *testing.B) {
            for i := 0; i < b.N; i++ {
                result := CompareKeysNew(item.key1, item.key2)
                if item.equal {
                    require.Zero(b, result)
                } else {
                    require.NotZero(b, result)
                }
            }
        })

    }
}

func CompareKeysMaster(key1, key2 []byte) int {
    AssertTrue(len(key1) > 8 && len(key2) > 8)
    if cmp := bytes.Compare(key1[:len(key1)-8], key2[:len(key2)-8]); cmp != 0 {
        return cmp
    }
    return bytes.Compare(key1[len(key1)-8:], key2[len(key2)-8:])
}

func CompareKeysNew(key1 []byte, key2 []byte) int {
    if len(key1) == len(key2) {
        return bytes.Compare(key1, key2)
    }
    // If the lengths are different, we don't need to compare the
    // timestamps since the keys itself are different.
    return bytes.Compare(key1[:len(key1)-8], key2[:len(key2)-8])
}
```

PR #954, opened 2019-08-01
Regenerate proto files

mangalaman93:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/954)


jarifibrahim:
@mangalaman93 Why do we have to regenerate the proto file? How did you generate them?
Also, please make necessary changes to https://github.com/dgraph-io/badger/blob/master/pb/gen.sh

mangalaman93:
I used only `gen.sh` to regenerate the protobuf files. May be, run the Makefile in Dgraph to update the libraries.

PR #955, opened 2019-08-01
Move Options to options package

ashish-goswami:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/955)


ashish-goswami:
This is not required now as we options at the table level.

Issue #820, opened 2019-05-21
Transaction error on backup restoration


awishformore:
We are running badger in a memory-constrained environment of less than 200 MB. When reducing the memory table size too much, the `db.Load(file)` call will fail with `Txn is too big to fit into one request`. I'm assuming this is due to the fact that it tries to squeeze the entire loading into a single transaction, which can not be bigger than the memory table limit? Would it be a good idea to change the code to batch this into several transactions, as no other operations should be executed on the database during loading anyway?

ashish-goswami:
@awishformore This looks like an issue at our end. We will check and get back to you.

jarifibrahim:
The `transaction too big` is being returned because we have hard coded the transaction size to be 1000 entries. The following code should take in to account the size of the tables and calculate the appropriate transaction size
https://github.com/dgraph-io/badger/blob/0ce1d2e26af1ba8b8a72ea864145a3e1e3b382cd/backup.go#L159-L161

PR #957, opened 2019-08-01
 Fix transaction too big issue in restore 

jarifibrahim:
When restore happens, we bunch entries together and perform batch
write. The existing implementation would batch 1000 entries. The batch
of 1000 entries can overflow the transactional limits in certain
circumstances (eg: when table size is too small). This commit fixes the
issue. We flush the entries once the batch reaches its max capacity

Fixes https://github.com/dgraph-io/badger/issues/820

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/957)


backup.go
```diff
@@ -175,7 +183,8 @@ func (l *KVLoader) send() error {
 		return err
 	}
 
-	l.entries = make([]*Entry, 0, 1000)
+	l.entries = l.entries[:0]
```
ashish-goswami: We should not do this as this will modify entries slice already passed to db.batchSetAsync.

PR #959, opened 2019-08-01
Making the stream writer APIs goroutine-safe

gitlw:
Adding locks around the StreamWriter APIs so that multiple go routines can call the APIs simultaneously without race conditions.
I've verified this change in bulk loader, which used to have RACE condition warnings reported by the race detector. With this change, those warnings are no longer showing up.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/959)

### DIFF
--- a/stream_writer.go
+++ b/stream_writer.go
@@ -79,7 +79,8 @@ func (sw *StreamWriter) Prepare() error {
 }
 
 // Write writes KVList to DB. Each KV within the list contains the stream id which StreamWriter
-// would use to demux the writes. Write is not thread safe and it should NOT be called concurrently.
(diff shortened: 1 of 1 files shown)


stream_writer.go
```diff
@@ -40,6 +41,7 @@ const headStreamId uint32 = math.MaxUint32
 // StreamWriter should not be called on in-use DB instances. It is designed only to bootstrap new
 // DBs.
 type StreamWriter struct {
+	writeLock sync.Mutex
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	writeLock  sync.Mutex
```

PR #964, opened 2019-08-05
Improve value log GC logic

jarifibrahim:
The current implementation of GC would sample a vlog even when we know how much
data can be discarded from it. The proposed implementation would GC a
file if the amount of stale data in it is greater than `discardRatio`.
Otherwise, it will pick two files, one from discard stats and the other
one randomly. It performs the sampling on both these files and tries to clean up
at least one of them.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/964)


manishrjain:
We still need to try a random file as well, in case the file with the highest chance of getting discarded, doesn't get discarded due to inadequate sample. In that case, every loop of ValueLogGC would retry the same file indefinitely, until it actually can. Random file allows us to still keep making progress.

jarifibrahim:
Closing this PR since the new implementation isn't any better than the existing one.

PR #965, opened 2019-08-05
Use pointers instead of binary encoding

jarifibrahim:
The change improves encoding/decoding performance by **4X**. See https://gist.github.com/jarifibrahim/30237927ff3a4b200d4907c97bd93f41 for benchmark code.
```
name                        time/op
EncodingSingle/Binary-16    4.02ns ± 1%
EncodingSingle/Slice-16     0.83ns ± 0%
Encodingu32Slice/Binary-16  9.52µs ± 1%
Encodingu32Slice/Slice-16   0.95ns ± 0%
```

The change also improves the performance of existing benchmarks.

```
name             old time/op  new time/op  delta
Read-16           153ms ± 1%   145ms ± 1%  -5.09%  (p=0.008 n=5+5)
ReadAndBuild-16   1.01s ± 1%   0.97s ± 3%  -4.04%  (p=0.008 n=5+5)
ReadMerged-16     980ms ± 1%   980ms ± 1%    ~     (p=1.000 n=5+5)
RandomRead-16    2.73µs ± 0%  2.65µs ± 1%  -2.73%  (p=0.008 n=5+5)
```

Inspired by https://github.com/coocood/badger/commit/0b03ef08d27de34e7cf907b4ba642d2b22f3f81e

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/965)


table/builder.go
```diff
@@ -281,3 +277,37 @@ func (b *Builder) writeChecksum(data []byte) {
 	_, err = b.buf.Write(buf[:])
 	y.Check(err)
 }
+
+func u32ToBytes(v uint32) []byte {
```
ashish-goswami: We can move these functions to y package so that these can be reused.

jarifibrahim: Moved to y.

PR #971, opened 2019-08-07
Prevent unnecessary safecopy in iterator parseKV

jarifibrahim:
The previous implementation of parseKV would copy over the value slice
every time, which is not necessary. The value slice should be copied only
when we return the result of point lookups. The proposed patch remove
safecopy call from the iterator.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/971)

### DIFF
--- a/table/iterator.go
+++ b/table/iterator.go
@@ -164,7 +164,7 @@ func (itr *blockIterator) parseKV(h header) {
 			itr.pos, h.klen, h.vlen, len(itr.data), h)
 		return
 	}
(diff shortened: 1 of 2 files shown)


jarifibrahim:
@ashish-goswami Review, please.

ashish-goswami:
Just check why travis is failing.

jarifibrahim:
The build on mac OS is failing on travis. It is failing for all PRs.

PR #972, opened 2019-08-08
Reuse block iterator inside table iterator

ashish-goswami:
Idea is to minimise allocations for block iterators by allocating block iterator only once and then reusing the same, inside table iterator.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/972)


jarifibrahim:
Please fix the PR title and description.

PR #982, opened 2019-08-13
In the bank tool, do not print all bank transfers by default.

martinmr:
The bank tool is currently printing each bank transfer by default, which
results in a lot of output, specially when running the test for a while.
This PR makes this optional and controllable by a --verbose flag.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/982)


manishrjain:
Good to go.

PR #987, opened 2019-08-16
Fix boundaries on GC batch size

connorgorman:
In the value log rewrite, the check to determine if the commit was going to be within transaction limits only checked if the value would be `>` than the max count. However, the check within the transaction is:

```
   if count >= db.opt.maxBatchCount || size >= db.opt.maxBatchSize {
        return nil, ErrTxnTooBig
    }
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/987)


jarifibrahim:
@ashish-goswami Please review this PR.

jarifibrahim:
The change looks harmless. I'm merging it.

PR #989, opened 2019-08-17
Pick necessary memtables for the key iterator.

balajijinnah:
This PR, will pick the right memtable for the prefix iterator.

It'll check weather the keys with give prefix exist in the memtable or not. If present, then It'll be given to iterator.

In order to find the biggest and smallest key of the memtable.
smallest- first entry of the lowest level.
largest - last entry of the lowest level.
last entry will be tracked will inserting data to the skiplist

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/989)

### DIFF
--- a/db.go
+++ b/db.go
@@ -508,6 +508,35 @@ func (db *DB) getMemTables() ([]*skl.Skiplist, func()) {
 	}
 }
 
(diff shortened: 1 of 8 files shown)


balajijinnah:
Closing this

Issue #992, opened 2019-08-19
Using badger.IteratorOptions.Prefix can hide data if certain keys have been written


jtacoma:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.11.6 linux/amd64
</pre>

### What version of Badger are you using?

I have tried with v1.6.0 as well as latest master (7a7dd17559ae0262a8ce9ec0bc1ce91d6f7ec942 at time of writing).

### Does this issue reproduce with the latest master?

Yes.

### What are the hardware specifications of the machine (RAM, OS, Disk)?

16 GiB RAM
Linux machine 4.19.0-5-amd64
400 GiB SSD

### What did you do?

I wrote two items to a badger database. The first of these items has the key `[]byte{0}`. Then I closed the database, re-opened it, and tried to find the second item using a prefix iterator, and cannot find it. Here is a test that fails, showing the issue:

```go
package main

import (
	"io/ioutil"
	"os"
	"testing"

	"github.com/dgraph-io/badger"
)

func TestBadger(t *testing.T) {
	dir, err := ioutil.TempDir("", "test-badger")
	if err != nil {
		t.Fatal(err)
	}
	defer os.RemoveAll(dir)
	func() {
		db, err := badger.Open(badger.DefaultOptions(dir))
		if err != nil {
			t.Fatal(err)
		}
		defer db.Close()
		txn := db.NewTransaction(true)
		defer txn.Discard()

		// Setting a value for key []byte{0} is necessary to trigger the bug.
		if err := txn.Set([]byte{0}, nil); err != nil {
			t.Fatal(err)
		}

		// This is the item we're looking for later. It will remain available
		// within the scope of this *badger.DB connection, but becomes partially
		// unavailable to later connections.
		if err := txn.Set(testKey(), []byte("42")); err != nil {
			t.Fatal(err)
		} else if err := txn.Commit(); err != nil {
			t.Fatal(err)
		}

		// Verify that the item we're looking for is available.
		got := CountAll(db)
		if err != nil {
			t.Fatal(err)
		} else if got != 1 {
			t.Fatalf("want 1, got %v", got)
		}
	}()
	func() {
		db, err := badger.Open(badger.DefaultOptions(dir))
		if err != nil {
			t.Fatal(err)
		}
		defer db.Close()

		got := CountAll(db)
		if err != nil {
			t.Fatal(err)
		} else if got != 1 {
			// The test fails here. CountAll is the same function used to verify the
[...]

PR #983, opened 2019-08-14
binary search based table picker

balajijinnah:
Signed-off-by: பாலாஜி <rbalajis25@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/983)

PR #997, opened 2019-08-21
Fix table.Smallest/Biggest and iterator Prefix bug

jarifibrahim:
The pick table compares the smallest and biggest key of each table
against the iteratorOptions.Prefix. The smallest and biggest key contain
timestamps while the prefix does not contain a timestamp. So pickTable
would return not return the correct tables.

For example
Table has Smallest [0 255 255... ] and prefix [0, 0, 1], the comparison
of these two values would return true implying that the smallest is
greater than the prefix which is incorrect because prefix doesn't
contains a timestamp and smallest does.

With this commit, we compare the smallest/biggest keys without
timestamps to the opt.Prefix.

Fixes https://github.com/dgraph-io/badger/issues/992

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/997)

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -349,10 +349,10 @@ func (opt *IteratorOptions) pickTable(t table.TableInterface) bool {
 		}
 		return key
 	}
-	if bytes.Compare(trim(t.Smallest()), opt.Prefix) > 0 {
+	if bytes.Compare(trim(y.ParseKey(t.Smallest())), opt.Prefix) > 0 {
 		return false
 	}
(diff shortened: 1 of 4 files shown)


ashish-goswami:
Just fixed this in https://github.com/dgraph-io/badger/pull/983

ashish-goswami:
Just check why travis is failing for Go1.11.

jarifibrahim:
@manishrjain Added a regression test.

jarifibrahim:
> Just check why travis is failing for Go1.11.

The test for checksum mismatch was failing but I reran and it passed. Looks like we have a flaky test.

PR #1034, opened 2019-09-13
Handle nil memtable gracefully on Close

connorgorman:
This is related to crashes on Close. I think it's a fair expectation that if you have multiple goroutines reading/writing a call to Close (on a sigterm for example) will not cause a panic. This fixes an issue where the memtable is nil on Close. This will cause Get's to return errors and not panic

This is an example of the panic
```
goroutine 85845 [running]:
github.com/dgraph-io/badger/skl.(*Skiplist).IncrRef(...)
	skl/skl.go:86
github.com/dgraph-io/badger.(*DB).getMemTables(0xc000491c00, 0x0, 0x0, 0x0, 0x0)
	db.go:484 +0xd7
github.com/dgraph-io/badger.(*DB).get(0xc000491c00, 0xc015c7c3c0, 0x38, 0x38, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	db.go:516 +0x66
github.com/dgraph-io/badger.(*Txn).Get(0xc00dfc5500, 0xc013083da0, 0x30, 0x30, 0x24, 0x30, 0xc017628378)
	txn.go:407 +0xe7
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1034)

### DIFF
--- a/db.go
+++ b/db.go
@@ -492,6 +492,10 @@ func (db *DB) getMemTables() ([]*skl.Skiplist, func()) {
 
 	tables := make([]*skl.Skiplist, len(db.imm)+1)
 
+	if db.mt == nil {
+		return nil, nil
+	}
+
 	// Get mutable memtable.
 	tables[0] = db.mt
 	tables[0].IncrRef()
@@ -527,6 +531,9 @@ func (db *DB) getMemTables() ([]*skl.Skiplist, func()) {
 // to ensure that we pick the highest version of the movekey present.
 func (db *DB) get(key []byte) (y.ValueStruct, error) {
 	tables, decr := db.getMemTables() // Lock should be released.
+	if len(tables) == 0 {
+		return y.ValueStruct{}, errors.New("No tables found. DB may be closed")
+	}
 	defer decr()
 
 	var maxVs *y.ValueStruct


ashish-goswami:
@connorgorman Thanks for the PR. Can you tell us, how did you end up with above crash? Also, if possible, please create issue for this.

connorgorman:
@ashish-goswami It's actually related to this one: https://github.com/dgraph-io/dgraph/issues/3873. Basically, that stacktrace shows two panics. It panics once on the getMemTables because it is getting a value, but then there is an iterator panic in a defer function because the DB has been closed

This scenario occurs when you Close the DB, but have a txn in progress (new txns get an error about the DB being closed)

connorgorman:
@jarifibrahim If you are trying to gracefully shutdown by calling Close, but have other readers then I believe they are not synchronized and will panic. https://github.com/dgraph-io/badger/blob/master/db.go#L407 is where db.mt is set to nil.

In the case where `len(tables) == 0`, based on the code in getMemTables, the only time when it would have 0 tables would be when `db.mt == nil` based on the change I made. Otherwise, there is always at least 1

connorgorman:
Circling back on this change, I think the correct Close behavior would be to wait for all txns including Views to complete while not allowing any new txns. Thoughts?

jarifibrahim:
>Circling back on this change, I think the correct Close behavior would be to wait for all txns including Views to complete while not allowing any new txns. Thoughts?

I had the same thing in mind. Instead of crashing with a nil pointer exception, it would be nice to complain to the user about the open transactions while closing the DB.

connorgorman:
@manishrjain Does it seem like it'd be better to just wait for View txns to close? The issue is that closing the DB does not wait for all txns to finish

manishrjain:
Hey @jarifibrahim , what's the status of this?

connorgorman:
@manishrjain @jarifibrahim  After looking at this and testing, it seems like this will not be sufficient for fixing the issue completely. e.g. if you use a merge operator, then the Get will succeed, but then the iterator will panic. It seems like the only way to fix this issue would be to ensure that all txns complete (both view and update) and then closing the DB (while not allowing new txns to begin). I took a stab at this, but it's fairly problematic due to the NewTransaction function being relied upon to always succeed and have a non-nil return value, which makes synchronizing difficult

Do you guys have any thoughts about how we could synchronize txns?

jarifibrahim:
Hey @connorgorman, I don't think there's any way we can handle all the `db closed` cases. For instance, even if we managed to handle nil memtables, there will be `nil` tables, also the `value log` buffers will be `nil`. The crash could happen at so many places. There's no way we can handle all the cases easily.

The crash happened in https://github.com/dgraph-io/dgraph/issues/3873 because we were accessing badger after closing it. A user/system should never do that. I think it's a reasonable expectation from the user to not access the DB after they have closed the database.

@connorgorman I'd like to close this PR since the issue we're trying to fix here is not really an issue with badger.

PR #1035, opened 2019-09-15
Add EventLogging option

### LINKED ISSUE #938: make event logging configurable

<pre>
$ go version
1.12.7
</pre>

v1.6.0

### What did you do?

Ran benchmarks using badgerdb.

### What did you expect to see?

I *didn't* expect to see logging appear on the CPU profile.

### What did you see instead?

It did, and there's no way to turn it off.

In some of my benchmarks ~10% of the time spent inside (*DB).doWrites (and more elsewhere!) is spent printing to the event log. ISTM there's no way to disable this.

mjgarton:
This introduces an option to disable or enable trace.EventLog logging of
events.  For backwards compatability this defaults to true.

The reason for making this configurable is that under some workloads the
cost of event logging is significant and so can cause performance
issues.

Fixes #938

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1035)


nil_event_log.go
```diff
@@ -0,0 +1,15 @@
+package badger
```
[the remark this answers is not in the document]
ashish-goswami: We can call this file just `event_log.go`.

ashish-goswami:
@mjgarton We need to propagate above change https://github.com/dgraph-io/badger/blob/86a77bb28c44052be16ce878ca1115ec0b83824e/value.go#L795 here also.

value.go
```diff
@@ -787,12 +787,15 @@ func (vlog *valueLog) replayLog(lf *logFile, offset uint32, replayFn logEntry) e
 	return nil
 }
 
-func (vlog *valueLog) open(db *DB, ptr valuePointer, replayFn logEntry) error {
+func (vlog *valueLog) open(db *DB, ptr valuePointer, replayFn logEntry, eventLogging bool) error {
```
ashish-goswami: No need to pass extra argument. It should be accessible from from `db.opt`.

mjgarton:
I missed another event logger in `y/watermark.go` so I need to add that.  I'll push some changes soon.

mjgarton:
Please advise whether `y` was the right place to put the null event logging code.  It isn't quite clear to me what the scope of the `y` package is supposed to be and I didn't find it documented.

y/event_log.go
```diff
@@ -0,0 +1,31 @@
...
+	NoEventLog trace.EventLog = nilEventLog{}
+)
+
+type nilEventLog struct{}
+
+func (nel nilEventLog) Printf(format string, a ...interface{}) {}
+
+func (nel nilEventLog) Errorf(format string, a ...interface{}) {}
+
+func (nel nilEventLog) Finish() {}
```
ashish-goswami: You might want to implement `SetMaxEvents` function also. As it is used below.
https://github.com/dgraph-io/badger/blob/c6681c181835f3846b859d6b546e5da8d9f86ea9/levels.go#L845

jarifibrahim:
Thank you for fixing this @mjgarton :tada:

PR #1038, opened 2019-09-17
Fix windows build for badger

jarifibrahim:
This commit fixes windows build issues. There were 3 issues

1. Trying to truncate a mmapped file.
    On linux, we can truncate a mmapped file but on windows this
    operation is invalid. We should unmap the file before truncating it.
2. Removing a file that has open file descriptor.
    On windows we cannot delete a file if it has an open file descriptor.
    The following functions would increment the ref for a table but never
    decrease it.
     - getTableInfo function in levels.go
    	The getTableInfo function would create an iterator but never
	close it and because of this we would have issues trying to
	delete the file.
     - CreateTable function in stream writer
	The createTable function calls OpenTable which sets the ref of a
	table to 1 but we would never decrement this ref.
3. AppVeyor file checksum mismatch error (iterator errors)
These errors were caused because we were trying to create a file with random name but on appVeyor the `rand.Int63()` would return same value for multiple calls. The problem was that `rand.Seed` was called with a `time.Now()` timestamp which wasn't always unique on AppVeyor 🤷‍♂ 
3. AppVeyor out of disk space error.
    In some of the tests we weren't closing the DB and trying to remove the directory. This meant the directory would never get deleted and thus AppVeyor would run out of disk space.

With this PR, the appVeyor build should succeed.

Badger ❤️ Windows

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1038)


levels.go
```diff
@@ -1000,6 +1000,7 @@ func (s *levelsController) getTableInfo(withKeysCount bool) (result []TableInfo)
 			var count uint64
 			if withKeysCount {
 				it := t.NewIterator(false)
+				defer it.Close()
```
ashish-goswami: having `defer` in `for` loop is not good idea. Iterator can be closed after counting.

table/table_test.go
```diff
@@ -59,10 +59,10 @@ func buildTable(t *testing.T, keyValues [][]string) *os.File {
 	opts := Options{BlockSize: 4 * 1024, BloomFalsePositive: 0.01}
 	b := NewTableBuilder(opts)
 	defer b.Close()
-	rand.Seed(time.Now().UnixNano())
 	// TODO: Add test for file garbage collection here. No files should be left after the tests here.
 
 	filename := fmt.Sprintf("%s%s%d.sst", os.TempDir(), string(os.PathSeparator), rand.Int63())
+	fmt.Println("creating file", filename)
```
ashish-goswami: This can be removed.

ashish-goswami:
Nice work @jarifibrahim. Feels very nice after seeing everything green :smile:

levels.go
jarifibrahim: Yeah. Thanks! Fixed it.

table/table_test.go
jarifibrahim: Removed it.

Issue #1049, opened 2019-09-22
Badger open data directory without any error with vlog file what contains broken value data


nordicdyno:
### What version of Go are you using (`go version`)?

<pre>
$ go version
1.13
</pre>

### What version of Badger are you using?

* v1.6
* v2.0.0-rc2

### Does this issue reproduce with the latest master?

yes

### What are the hardware specifications of the machine (RAM, OS, Disk)?

MacBook Pro (15-inch, 2018), 2,2 GHz i7, 16Gb, APPLE SSD AP0256M

### What did you do?

I've spoiled data in vlog file in data directory (emulate real world issues with hardware) and badger just starts with this directory with any error. 

### What did you expect to see?

Error on database open.

### What did you see instead?

Badger works without any error if spoiled data are in values body.

## Extra

More info and reproducible tests are here:
https://github.com/nordicdyno/badger-spoiler

I'm not sure, but probably it's related to #601

jarifibrahim:
@nordicdyno The test you're referring to corrupts a bit in the middle of the file. This is an unlikely scenario. AFAIK, a file cannot be corrupted in the middle. It can be corrupted only from the end.

If the corruption was at the end, badger would've recognized it. We perform checksum verification of the entries in vlog after a specific point (called as `head` point).

nordicdyno:
I disagree. It's very likely scenario if we speak about disk corruption after badger stop*. I don't see any reason why it could NOT happen. And problem is what we never know about it as badger users.  

 _* I don't want to speak about probability here, but real world could be a messy place_

nordicdyno:
Why not just add checksum for every item in vlog and checking it against replayed values on start? (or smth like that)

jarifibrahim:
This should be fixed by https://github.com/dgraph-io/badger/pull/1052 .

Thanks for writing a test @nordicdyno :)

afiskon:
For the record - this might be not a problem if the data  is corrupted before the checkpoint (meaning the record that means that everything before it was synced to the disk), e.g. the record will not be used anyway during the recovery process after the restart. If the vlog entry is used during the recovery then I agree with @nordicdyno it's checksum should be checked.

While working on PostgreSQL I had an unpleasant experience of recovering data from databases/backups which random entries were corrupted. Sadly sometimes this really happens in real world due to hardware failures.

jarifibrahim:
[...]

PR #1052, opened 2019-09-24
Support checksum verification for values read from vlog

jarifibrahim:
The values are written with its checksum but we were not validating it
for every read. This commit enables checksum validation for every read.

If `VerifyValueChecksum` is set to true, the checksum will be calculated for
each entry read from the value log. This option has no effect if the
value is stored completely in SST.

Fixes https://github.com/dgraph-io/badger/issues/1049

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1052)


value.go
```diff
@@ -1165,6 +1165,16 @@ func (vlog *valueLog) Read(vp valuePointer, s *y.Slice) ([]byte, func(), error)
 		return nil, cb, err
 	}
 
+	if vlog.opt.VerifyValueChecksum {
+		hash := crc32.New(y.CastagnoliCrcTable)
+		hash.Write(buf)
```
golangcibot: Error return value of `hash.Write` is not checked (from `errcheck`)

jarifibrahim: Fixed.

nordicdyno:
Is any hope this fix would land in 1.6.x branch?

afiskon:
```
// Feel free to modify these to suit your needs with the WithX methods.
func DefaultOptions(path string) Options {
	return Options{
// ...
		VerifyValueChecksum:     false,
```

As I mentioned in https://github.com/dgraph-io/badger/issues/1049 it's worth considering making it a default behavior.

jarifibrahim:
> Is any hope this fix would land in 1.6.x branch?

We should be able to cherry-pick this into the 1.6.x branch.

jarifibrahim:
> ```
> // Feel free to modify these to suit your needs with the WithX methods.
> func DefaultOptions(path string) Options {
> 	return Options{
> // ...
> 		VerifyValueChecksum:     false,
> ```
> 
> As I mentioned in #1049 it's worth considering making `VerifyValueChecksum` true by default.

I am not very sure if the default behavior should be to verify the checksum of every entry read from value log. This would significantly affect the performance. Keeping it to `false` allows us to keep the existing behavior (and performance).

The chances of getting corrupted data somewhere in the middle of a value log file are anyway pretty slim. If there was some corruption, it might have affected SSTs too. In that case, Badger would recognize corruption when DB is opened. We perform checksum verification for every SST file when we open the DB. So I am not sure if we should keep it enabled by default.

afiskon:
> I am not very sure if the default behavior should be to verify the checksum of every entry read from value log. This would significantly affect the performance. Keeping it to false allows us to keep the existing behavior (and performance). The chances of getting corrupted data somewhere in the middle of a value log file are anyway pretty slim.

I would argue that the default behavior of the storage engine is to provide consistency, not to look good on benchmarks. Users interested in performance can change the default value if they like. Users interested in consistency may not even know that by default Badger sometimes can loose data, even if chances are pretty slim.

PR #1054, opened 2019-09-25
Set datakey for first table in stream writer

jarifibrahim:
If encryption is enabled on a DB and stream writer is used then the
first file SST would not be encrypted, only the SSTs after the first one
would be encrypted. This PR fixes it.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1054)

### DIFF
--- a/stream_writer.go
+++ b/stream_writer.go
@@ -128,7 +128,11 @@ func (sw *StreamWriter) Write(kvs *pb.KVList) error {
 	for streamId, req := range streamReqs {
 		writer, ok := sw.writers[streamId]
 		if !ok {
-			writer = sw.newWriter(streamId)
+			var err error
+			writer, err = sw.newWriter(streamId)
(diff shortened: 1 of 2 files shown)


stream_writer.go
```diff
@@ -206,10 +213,16 @@ type sortedWriter struct {
 	head     valuePointer
 }
 
-func (sw *StreamWriter) newWriter(streamId uint32) *sortedWriter {
+func (sw *StreamWriter) newWriter(streamId uint32) (*sortedWriter, error) {
```
golangcibot: method parameter `streamId` should be `streamID` (from `golint`)

jarifibrahim: Fixed.

jarifibrahim:
Merging this PR since it's a simple change and doesn't require @manishrjain's review.

PR #1069, opened 2019-10-07
Change zstd compression algorithm

jarifibrahim:
Currently, we use https://github.com/DataDog/zstd implementation for ZSTD compression and https://github.com/valyala/gozstd seems to be faster than https://github.com/DataDog/zstd . This PR 
proposes use of `valyala/gozstd` instead of `DataDog/zstd`.

```
benchstat zstd.txt gozstd.txt 
name                       old time/op  new time/op  delta
Compression/compress-16     213ms ± 6%   116ms ± 3%  -45.79%  (p=0.000 n=9+10)
Compression/decompress-16  48.5ms ± 3%  45.0ms ± 1%   -7.30%  (p=0.000 n=10+10)
```

The following script was used for benchmarking
```go
var a []byte
func BenchmarkCompression(b *testing.B) {
	opts := Options{BlockSize: 4 * 1024, BloomFalsePositive: 0.01}
	count := uint64(1e6)
	builder := NewTableBuilder(opts)
	// Approx 64 MB table size
	for i := uint64(0); i < count*2; i++ {
		k := fmt.Sprintf("%016x", i)
		v := fmt.Sprintf("%016d", i)
		builder.Add([]byte(k), y.ValueStruct{Value: []byte(v)})
	}
	// Build a dummy table so that we can use the same buffer for benchmarks
	buf := builder.Finish()

	var x []byte
	var err error
	// Used for decompression benchmarks
	// cbuf, err := zstd.Compress(nil, buf) 	// Use this buffer for zstd.
	// require.NoError(b, err)
	cbuf1 := gozstd.Compress(nil, buf)

	fmt.Println("len of buf", humanize.Bytes(uint64(len(buf))))
	b.ResetTimer()
	b.Run("compress", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			x = gozstd.Compress(nil, buf) // change this for zstd.
		}
		a = x
	})
	b.Run("decompress", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			x, err = gozstd.Decompress(nil, cbuf1) // change this for zstd.
			if err != nil {
				panic(err)
			}
		}
		a = x
	})
}
```

The following is the benchmark of time taken to build a table in badger (https://github.com/dgraph-io/badger/blob/f50343ff404d8198df6dc83755ec2eab863d5ff2/table/builder_test.go#L116)
```
benchstat master-builder.txt new-builder.txt                       
name        old time/op  new time/op  delta
Builder-16   1.02s ± 1%   0.73s ± 2%  -27.77%  (p=0.000 n=9+10)
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1069)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/spf13/cobra v0.0.5
 	github.com/stretchr/objx v0.2.0 // indirect
 	github.com/stretchr/testify v1.4.0
-	github.com/valyala/gozstd v1.6.2 // indirect
+	github.com/valyala/gozstd v1.6.2
 	golang.org/x/net v0.0.0-20190620200207-3b0461eec859
 	golang.org/x/sys v0.0.0-20190626221950-04f50cda93cb
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/dgraph-io/badger
 go 1.12
 
 require (
-	github.com/DataDog/zstd v1.4.1
 	github.com/cespare/xxhash v1.1.0
 	github.com/cespare/xxhash/v2 v2.1.0 // indirect
 	github.com/dgraph-io/ristretto v0.0.0-20190916120426-cd2835491e0e
@@ -16,6 +15,7 @@ require (
 	github.com/spf13/cobra v0.0.5
 	github.com/stretchr/objx v0.2.0 // indirect
 	github.com/stretchr/testify v1.4.0
+	github.com/valyala/gozstd v1.6.2 // indirect
 	golang.org/x/net v0.0.0-20190620200207-3b0461eec859
 	golang.org/x/sys v0.0.0-20190626221950-04f50cda93cb
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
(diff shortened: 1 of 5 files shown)


ashish-goswami:
Benchmarks are surely better for `gozstd` than `zstd`. But we should also consider stability and support for both. Just try to find who are users of both libraries(if they are public) and using those in production.

jarifibrahim:
Benchmarks show LZ4 is faster than zstd and so we don't need this PR anymore.
```
go test -v -run ^$ -bench BenchmarkTableCompression
data size 4056
goos: linux
goarch: amd64
pkg: github.com/dgraph-io/badger/table
BenchmarkTableCompression/Compression/Snappy-16 	   	   	    	  148005	      8103 ns/op
BenchmarkTableCompression/Compression/LZ4-16      	   	   	   	  	  192206	      5589 ns/op
BenchmarkTableCompression/Compression/ZSTD_-_Datadog-16         	   24697	     47433 ns/op
BenchmarkTableCompression/Decompression/Snappy-16               	  308114	      3952 ns/op
BenchmarkTableCompression/Decompression/LZ4-16                  	  379030	      2728 ns/op
BenchmarkTableCompression/Decompression/ZSTD_-_Datadog-16       	  100808	     11431 ns/op
PASS
ok  	github.com/dgraph-io/badger/table	8.708s
```
https://gist.github.com/jarifibrahim/50b596c233aa9d82706aa0df65a546ec used for benchmarking.

arl:
I know this is closed but there's another reason to switch zstd library, a serious memory corruption bug in DataDog/zstd, that can either to segfaults, lower compression ratio, or nothing depending on your usage of the library.

Please see https://github.com/DataDog/zstd/issues/22

.

Issue #1032, opened 2019-09-10
deadlock when writing into Badger using multiple routines


tony2001:
go version go1.13 linux/amd64
github.com/dgraph-io/badger v2.0.0-rc3+incompatible
I have to admit I didn't try latest master, but I've seen this issue since early versions of Badger 2.0 and it's still there with rc3.
The server is running on 56-core Intel Xeon E5-2660, no idea about the disk.

I'm reading several multi-gigabyte files in their own binary format and "converting" them into Badger, writing data encoded with Protobuf. The initial data is organized in batches (of arbitrary size, from 1 to 1000+ records in a batch), so I'm using WriteBatches to write the actual data and then badger.Update() to update the counters (stored in the same badger db). To speed up the process I'm using 64 goroutines that listen on channel with data and do the actual encoding/writing.

The problem is that after some time the badger deadlocks and stops writing anything, the process is stuck with the backtraces provided below.
As you can see from the backtrace 3, I tried to use Txn.Set/Commit directly instead of using badger.Update, but it didn't help.
Unfortunately, I'm unable to provide a reproduce code so far, as the issue seems to be reproducible only on really large amounts of data after a couple of hours of running with no error messages whatsoever. To make it more complicated, I'm unable to reproduce it in 100% of cases even using the same initial data and the same Go code.

My code looks very much like this (in simplified go "pseudocode"):
```
//this is the function that's called from 64 worker routines
func LoadData(records []*Record) error { 
    wb := badger.NewWriteBatch()
    defer wb.Cancel()
    for ... {
           err = wb.Set(key, value)
           if err != nil {
               return err
           }
    }
    err = wb.Flush()
    if err == nil {
        updateStats(..)
    }
    return err
}

func updateStats(...) error {
    statsLock.Lock()
    defer statsLock.Unlock()
    stats += value
    //...serialize value into val
    txn := badger.NewTransaction(true)
    defer txn.Discard()

    err = txn.Set(key, val)
    if err != nil {
        return err
    }

    err = txn.Commit()
    if err != nil {
        return err
    }
    return err
}
```

Backtraces:
1.
```
6 @ 0x43d600 0x44dcc0 0x44dcab 0x44d912 0x4762e4 0x976709 0x474c83 0x975cb2 0x975c5e 0x985549 0xa16b0b 0xa800ac 0x46b381
#	0x44d911	sync.runtime_Semacquire+0x41					/home/tony/go/src/runtime/sema.go:56
[...]

PR #1070, opened 2019-10-07
Fix deadlock in discard stats

ashish-goswami:
Fixes https://github.com/dgraph-io/badger/issues/1032

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1070)

### DIFF
--- a/db.go
+++ b/db.go
@@ -381,7 +381,7 @@ func (db *DB) close() (err error) {
 	db.elog.Printf("Closing database")
 
 	// Before closing writes, stop flushing discardStats.
-	db.vlog.closeFlushDiscardStats()
+	db.vlog.stopFlushDiscardStats()
 
 	atomic.StoreInt32(&db.blockWrites, 1)
 
--- a/db.go
+++ b/db.go
@@ -380,9 +380,6 @@ func (db *DB) Close() error {
 func (db *DB) close() (err error) {
 	db.elog.Printf("Closing database")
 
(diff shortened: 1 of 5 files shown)


balajijinnah:
LGTM. But, if we have flusher go routine in the db level. It fits well with the existing mechanism(all go routine are spined in the db.go).

Also, It'll easy to merge this change to the WAL and vlog separation.

connorgorman:
@jarifibrahim Is the description of the deadlock in https://github.com/dgraph-io/badger/issues/1032 correct or was there a different issue at hand? This change is pretty complex so it's difficult to determine when this was introduced and what versions may be affected

jarifibrahim:
Hey @connorgorman the description of issue #1032 was correct and this PR has fixed it.

When we perform compaction, we flush discard stats back into badger. These discard stats follow the same path as normal keys (memtable -> L0 -> L1 -> ... ). In the reported issue, compaction was triggered because L0 and memtable were full. When the compaction completed, it tried to push the discard stats back into the DB but since memtable was full, we could not push the entry into badger. This was causing the deadlock. It should have been fixed by this PR.

See https://github.com/dgraph-io/badger/commit/c1cf0d7e3d9f893561c847f0e5a62f759adfb4bb for some more explanation.

If you're looking for when this issue was introduced, it was introduced in https://github.com/dgraph-io/badger/commit/fa0679c19e7a37bc35cf5c0333420018a7668021 .

PR #1082, opened 2019-10-15
Fix wrong namespace in example

fleaz:
Hey,

just a small fix so the example can be copy/pasted and will compile directly.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1082)


fleaz:
The code snippet two above contains the "base" (imports, main func, opening db) so I´m not sure how useful it would be to copy it into this example. All the other examples also just contain their single snippet and not a full code example what is imho better for readability.

PR #1091, opened 2019-10-21
Enable encryption in bank test

jarifibrahim:
This PR moves the `--encryption` flag from `bank disect` to `bank test` command. This PR also adds a `decryption-key` flag to `badger disect` to allow opening a previously encrypted DB.

`badger test` now also dumps the encryption key in logs so that the same can be used by `badger disect`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1091)


badger/cmd/bank.go
```diff
@@ -339,16 +347,14 @@ func runTest(cmd *cobra.Command, args []string) error {
...
-		_, err := rand.Read(key)
-		if err != nil {
-			return err
-		}
-		opts = opts.WithEncryptionKey(key)
+	if encryptionKey != "" {
+		opts = opts.WithEncryptionKey([]byte(encryptionKey))
+		// The following comment is intentional as we would need the encryption key in hex in case
+		// we want to run disect tool on the directory generated by bank test tool.
+		log.Printf("Using encryption key %s\n", key)
```
golangcibot: printf: Printf format %s arg key is a func value, not called (from `govet`)

Issue #1096, opened 2019-10-22
checkptr issue found


campoy:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version devel +7d891d5e4d Tue Oct 22 21:23:03 2019 +0000 linux/amd64
</pre>

### What version of Badger are you using?

Tip (be6402f83d3126230c7fc51543abe7c691e94ed6)

### Does this issue reproduce with the latest master?

Yes

### What are the hardware specifications of the machine (RAM, OS, Disk)?

Linux, N/A

### What did you do?

I ran the tests with `checkptr` enabled.
New in go tip is a checkptr flag: https://groups.google.com/forum/#!msg/golang-dev/SzwDoqoRVJA/IvtnBW5oDwAJ

### What did you expect to see?

All tests should pass

### What did you see instead?

Two unsafe pointer conversions were found. See logs below.

```bash
gotip test -gcflags=all=-d=checkptr ./...
badger 2019/10/22 15:00:36 INFO: All 0 tables opened in 0s
badger 2019/10/22 15:00:36 INFO: DB.Backup Created batch of size: 50 B in 34.925µs.
badger 2019/10/22 15:00:36 INFO: DB.Backup Sent 2 keys
badger 2019/10/22 15:00:36 DEBUG: Storing value log head: {Fid:0 Len:29 Offset:101}
panic: runtime error: unsafe pointer conversion

goroutine 68 [running]:
github.com/dgraph-io/badger/table.(*header).Decode(...)
	/home/francesc/src/github.com/dgraph-io/badger/table/builder.go:57
github.com/dgraph-io/badger/table.(*blockIterator).setIdx(0xc0009a2010, 0x2)
	/home/francesc/src/github.com/dgraph-io/badger/table/iterator.go:82 +0x20a
github.com/dgraph-io/badger/table.(*blockIterator).seekToLast(...)
	/home/francesc/src/github.com/dgraph-io/badger/table/iterator.go:141
github.com/dgraph-io/badger/table.(*Iterator).seekToLast(0xc0009a2000)
	/home/francesc/src/github.com/dgraph-io/badger/table/iterator.go:217 +0x18d
github.com/dgraph-io/badger/table.(*Iterator).Rewind(0xc0009a2000)
	/home/francesc/src/github.com/dgraph-io/badger/table/iterator.go:378 +0x34
github.com/dgraph-io/badger/table.(*Table).initBiggestAndSmallest(0xc000998000, 0x0, 0x0)
[...]

PR #1097, opened 2019-10-23
Fix unsafe usage in Decode

jarifibrahim:
We use unsafe pointers to convert slices to structs. It might be possible that the struct and byte slice are not aligned in the same way.  This PR fixes the alignment issues

The command
```
gotip test -gcflags=all=-d=checkptr
```
no longer complains about unsafe pointer conversion.

The changed implementation is as fast as the existing one.

Fixes https://github.com/dgraph-io/badger/issues/1096

```go                                                                                                                           
// Implicit copy
func (h *header) Decode(buf []byte) int {
	*h = *(*header)(unsafe.Pointer(&buf[0]))
	return h.Size()
}

// Explicit copy
func (h *header) DecodeWithCopy(buf []byte) int {
   copy((*[headerSize]byte)(unsafe.Pointer(h))[:], buf[:headerSize])
   return int(headerSize)    
}    
      
```

```go
var x header
 
func BenchmarkHeader(b *testing.B) {
 h := &header{
     diff:    100,
     overlap: 023,
 }
 
 buf := h.Encode()
 b.ResetTimer()
 b.Run("implicit copy", func(b *testing.B) {
     for i := 0; i < b.N; i++ {
         x.Decode(buf)
     }
 })
 b.Run("explicit copy", func(b *testing.B) {
     for i := 0; i < b.N; i++ {
         x.DecodeWithCopy(buf)
     }
 })
}
```
```
gtf -run ^$ -bench=Header -benchmem
goos: linux
goarch: amd64
pkg: github.com/dgraph-io/badger/v2/table
BenchmarkHeader/implicit_copy-16         	1000000000	         0.602 ns/op
BenchmarkHeader/explicit_copy-16         	1000000000	         0.712 ns/op
PASS
ok  	github.com/dgraph-io/badger/v2/table	1.467s
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1097)


campoy:
I can confirm this fixes the issue for this test, unfortunately now it fails somewhere else.

```bash
--- FAIL: TestTruncateVlogWithClose (0.06s)
panic: runtime error: unsafe pointer conversion [recovered]
	panic: runtime error: unsafe pointer conversion

goroutine 53572 [running]:
testing.tRunner.func1(0xc0003a3e00)
	/home/francesc/src/github.com/golang/go/src/testing/testing.go:881 +0x3a3
panic(0xa43860, 0xc088c4b760)
	/home/francesc/src/github.com/golang/go/src/runtime/panic.go:679 +0x1b2
github.com/dgraph-io/badger.(*valuePointer).Decode(...)
	/home/francesc/src/github.com/dgraph-io/badger/structs.go:42
github.com/dgraph-io/badger.Open(0xc087f24f80, 0x19, 0xc087f24f80, 0x19, 0x1, 0x2, 0x2, 0x1, 0x100, 0xb5e4c0, ...)
	/home/francesc/src/github.com/dgraph-io/badger/db.go:350 +0x16c5
github.com/dgraph-io/badger.TestTruncateVlogWithClose(0xc0003a3e00)
	/home/francesc/src/github.com/dgraph-io/badger/db2_test.go:72 +0x416
testing.tRunner(0xc0003a3e00, 0xab3c18)
	/home/francesc/src/github.com/golang/go/src/testing/testing.go:916 +0xc9
created by testing.(*T).Run
```

jarifibrahim:
@campoy Both the errors are related. I'm still trying to understand why this fix works. For instance, I expected the code below to complain but it doesn't
```go
type header struct {
	a uint16
	b uint16
}

func (h header) encode() []byte {
	var b [4]byte
	*(*header)(unsafe.Pointer(&b[0])) = h
	return b[:]
}

func (h *header) decode(buf []byte) {
	*h = *(*header)(unsafe.Pointer(&buf[0]))
}

func TestChkptr(t *testing.T) {

	h := header{1, 2}
	b := h.encode()
	var y header
	y.decode(b)

	fmt.Println(h, y)
}

```
The code is exactly similar to one in badger.

jarifibrahim:
@campoy the errors are related. I've fixed them both in this PR.

Do you think we should add the `checkPtr` flag to all our builds?

campoy:
I agree that @mangalaman93's solution seems more efficient (although quite ugly?)

If that one fixes the warnings let's use it and avoid the performance hit.

mangalaman93:
@campoy Such conversions are pretty common in badger as far as I know. Let me know which part of the solution you thought could be improved/avoided.

jarifibrahim:
@mangalaman93 Thank you so much. The patch seems to work :+1:

Issue #1104, opened 2019-10-29
Pure Go? Requires gcc for zstd


zanicar:
Utilizing badger master branch as of 2019-10-29

For BadgerDB to be classified as Pure Go the dependency on zstd should be removed, or optional, as it requires gcc to build. Systems with pure Go installations are not able to install or build when including "github.com/dgraph-io/badger" as a result of this dependency. I only noticed this when operating in my Windows (pure Go) environment that, unlike my Linux environment, does not have gcc and build essentials equivalent installed.

<pre>
$ go version
go version go1.13.3 windows/amd64
</pre>

<pre>
$ go get github.com/dgraph-io/badger/...
# github.com/DataDog/zstd
exec: "gcc": executable file not found in %PATH%
</pre>

<pre>
$ go get github.com/dgraph-io/badger
# github.com/DataDog/zstd
exec: "gcc": executable file not found in %PATH%
</pre>

jarifibrahim:
ZSTD/CGO is optional. You can disable it via `CGO_ENABLED=0` flag. See https://github.com/dgraph-io/badger/pull/1094.

jarifibrahim:
From https://github.com/dgraph-io/badger/pull/1094#issuecomment-547373672 
>`go get` tries to build ZSTD by default.

@zanicar You can do `go get -d ...` for now to get around the issue. I'll work on fixing this.
Thanks for reporting :+1:

vmihailenco:
Sorry to interfere, but what is the plan regarding ZSTD? Will it be removed completely or supported optionally?

Some options:
- https://github.com/valyala/gozstd which comes with pre-compiled ZSTD
- https://github.com/klauspost/compress/tree/master/zstd - pure Go variant with beta status

jarifibrahim:
> * https://github.com/valyala/gozstd which comes with pre-compiled ZSTD

gozstd also requires cgo/gcc. So it's no good than `datadog/zstd`.

> * https://github.com/klauspost/compress/tree/master/zstd - pure Go variant with beta status

The library is in beta state and it clearly says `there may still be subtle bugs, but a wide variety of content has been tested` https://github.com/klauspost/compress/tree/master/zstd#status


[...]

PR #1105, opened 2019-10-31
Remove ZSTD compression algorithm

jarifibrahim:
The ZSTD compression algorithm library [github.com/DataDog/ZSTD](https://github.com/DataDog/zstd) requires CGO. It
enforces badger users to have gcc/g++ while doing a
`go get github.com/dgraph-io/badger`. This commit removes ZSTD
algorithm from badger so that the users don't need to have gcc/g++
installed to use badger.

See https://github.com/dgraph-io/badger/issues/1104

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1105)


jarifibrahim:
Closing this PR since we've decided to keep the ZSTD compression algorithm.

Issue #1001, opened 2019-08-23
Support disk-less mode


peterbourgon:
Can BadgerDB support a pure in-memory mode, where there is no disk access at all?

[Related discuss.dgraph.io thread](https://discuss.dgraph.io/t/badgerdb-without-touching-disk/4945).

jarifibrahim:
@peterbourgon We don't support this right now. The value log acts as a WAL which has to be written to the disk.
We might support a memory-only mode in future.

manishrjain:
I think @balajijinnah is working on making it so that if writes don't need to be present in value log (values are already stored in the LSM tree), they won't be written to the log. That change along with setting very high ValueThreshold should give us the ability to not write to value log at all. Then all we need to do is to keep Tables in memory, which is already an option.

Furthermore, if no persistence is required, we can add another option to keep all levels in memory, instead of them being written to disk. This feature could also be useful to allow Dgraph to act like a search engine instead of just a DB.

AgentZombie:
I'm interested in this capability for the purpose of writing unit tests for code that uses badger and would like to avoid those tests depending on having access to write to the disk.

PR #1113, opened 2019-11-12
Introduce disk-less mode in badger

jarifibrahim:
This PR introduces disk-less mode in badger. The disk-less mode can be enabled by setting `options.DiskLess = true`. When badger is running in disk-less mode no files are created and everything is stored in memory.

On DB close, all stored data is lost.

NOTE - An existing DB cannot be opened in diskless mode.

Fixes https://github.com/dgraph-io/badger/issues/1001

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1113)

### DIFF
--- a/db.go
+++ b/db.go
@@ -224,9 +224,11 @@ func Open(opt Options) (db *DB, err error) {
 		// Do not perform compaction in read only mode.
 		opt.CompactL0OnClose = false
 	}
-	var dirLockGuard, valueDirLockGuard *directoryLockGuard
-	var manifest Manifest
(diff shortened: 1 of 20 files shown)


db.go
```diff
@@ -305,22 +307,25 @@ func Open(opt Options) (db *DB, err error) {
 		blockCache:    cache,
 	}
 
+	fmt.Println("Opened DB with uuid %s", db.uuid.String())
```
golangcibot: printf: Println call has possible formatting directive %s (from `govet`)

Issue #1126, opened 2019-11-19
Data loss while reading from DB


ForCraft2013:
Hello. I have an issue. Badger loses data while reading data when app suddenly stops, via Ctrl+C (for example). And this happenes when I read data just through **db.View** (_read only mode_).

Go: 1.13.4
Badger: 2.0.0
Windows 10 build 1903

After stop, vlog file becomes too big (2 147 483 646 bytes aka 2 GB). And if i run app again, Badger gives an error
`During db.vlog.open: Value log truncate required to run DB. This might result in data loss`
And finally, If I run Badger with **WithTruncate(true)** option, data loses. Vlog file size becomes 20 bytes.

I want to use Badger in my app, but 'cause of this problem i can't. Any error/stop while reading can cause lossing data.

The problem can be reproduced with this code (Ctrl+C while program prints keys and values into console):

**read.go**
```
func main() {
        db, err := badger.Open(badger.DefaultOptions("tmp/badger").WithSyncWrites(true))
	if err != nil {
		log.Panic(err)
	}
	defer db.Close()

	err := db.View(func(txn *badger.Txn) error {
                opts := badger.DefaultIteratorOptions
                opts.PrefetchSize = 10
		it := txn.NewIterator(opts)
		defer it.Close()

		for it.Rewind(); it.Valid(); it.Next() {
			item := it.Item()
			k := item.Key()
			err := item.Value(func(v []byte) error {
				fmt.Printf("key=%d, value=%s\n", k, v)
				return nil
			})
                        if err != nil {
				return err
			}
		}
		return nil
	})
        if err != nil {
		log.Panic(err)
	}
}
```

**write.go**
```
func main() {
        db, err := badger.Open(badger.DefaultOptions("tmp/badger").WithSyncWrites(true))
	if err != nil {
		log.Panic(err)
	}
	defer db.Close()

	seq, err := db.GetSequence([]byte("0"), 1)
	if err != nil {
		log.Panic(err)
	}
	defer seq.Release()

        file, err := os.Open("some_big_data_file.txt")
	if err != nil {
		return err
	}
	defer file.Close()

	scanner := bufio.NewScanner(file)
	err := db.Update(func(txn *badger.Txn) error {
		for scanner.Scan() {
			line := []byte(strings.TrimSpace(scanner.Text()))
			if len(line) > 0 {
				id, err := seq.Next()
				if err != nil {
					return err
				}
				key := make([]byte, 8)
[...]

PR #1134, opened 2019-11-25
Fix windows dataloss issue

jarifibrahim:
Windows doesn't allow memory mapping a file to a size greater than the file's actual size. To circumvent this, we increase the file size by truncating it.
https://github.com/dgraph-io/badger/blob/f5b63211d7f3e2f5f8b698893313b2a54e4df7de/y/mmap_windows.go#L41-L48

When badger would re-open, we try to replay this "truncated" file. Since this truncated file consists of all zeros, the replay would return the last offset as `zero` and then we would truncate the original file to size `zero`.

The replay function would return `zero` as the last valid offset which was wrong. The last valid offset is start offset plus the forward movements of the file offset.
So instead of 
https://github.com/dgraph-io/badger/blob/f5b63211d7f3e2f5f8b698893313b2a54e4df7de/value.go#L433
```go
var validEndOff uint32 // notice we're starting from zero, not the start point.
```
we should be doing
```go
var validEndOff uint32 = offset
```
Fixes - https://github.com/dgraph-io/badger/issues/1126

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1134)

### DIFF
--- a/db2_test.go
+++ b/db2_test.go
@@ -27,6 +27,8 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"runtime"
+	"sync"
 	"testing"
 
 	"github.com/dgraph-io/badger/v2/options"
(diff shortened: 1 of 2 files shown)


zhenruyan:
I see that this issue has been resolved, but currently I am encountering the same phenomenon on Windows 10 Golang 1.21.0. When the program interrupts abnormally and fails to close properly, the vlog size becomes 1kb and prompts "Value log truncate required to run DB. This might result in data loss"

PR #1143, opened 2019-12-04
Fixed Txn size calculation and ensure write after first Set call in case of ErrTxnTooBig

adamstruck:
This PR address a scenario that causes an OOM error.  

The Txn size is underestimated due to the calculation here: https://github.com/dgraph-io/badger/blob/407e5bde06c44e5503dc1eaf5a11cc9f2f348ac4/structs.go#L153

Since pending writes are stored in `*Txn.pendingWrites`, if you try to write thousands of large entries (~16mb in my use case) during a transaction (or using WriteBatch) you will encounter an OOM. 

I encountered this issue while using WriteBatch.  The code snippet below approximates my use case.
```
package main

import (
	"math/rand"
	"os"
	"time"

	"github.com/dgraph-io/badger"
	"github.com/sirupsen/logrus"
)

var logger *logrus.Logger
var log *logrus.Entry

func init() {
	logger = logrus.New()
	logger.SetLevel(logrus.DebugLevel)
	logger.SetFormatter(&logrus.JSONFormatter{
		TimestampFormat: time.RFC3339,
	})
	log = logger.WithFields(logrus.Fields{"ns": "main"})
}

func makeRandBytes(size int) []byte {
	val := make([]byte, size)
	_, err := rand.Read(val)
	if err != nil {
		log.Error("generating random bytes:", err)
		return val
	}
	return val
}

func write(tx *badger.WriteBatch, n int, valueSize int) {
	for j := 0; j < n; j++ {
		key := makeRandBytes(16)
		val := makeRandBytes(valueSize)
		err := tx.Set(key, val)
		if err != nil {
			log.Error("set:", err)
			continue
		}
		if j%100 == 0 {
			log.Infof("set %d items", j)
		}
	}
}

func main() {
	sub := logger.WithFields(logrus.Fields{"ns": "badger"})
	path := "./badger.db"
	opts := badger.DefaultOptions(path)
	opts = opts.WithLogger(sub)
	db, err := badger.Open(opts)
	if err != nil {
		log.Error("open badger:", err)
		os.Exit(1)
	}

	bt := db.NewWriteBatch()
	defer bt.Cancel()
        write(bt, 20000, 16000000)
	log.Infof("flush")
	err = bt.Flush()
	if err != nil {
		log.Error("flush err:", err)
	}
	log.Infof("done")
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1143)

### DIFF
--- a/txn.go
+++ b/txn.go
@@ -295,11 +295,8 @@ func (txn *Txn) newPendingWritesIterator(reversed bool) *pendingWritesIterator {
 func (txn *Txn) checkSize(e *Entry) error {
 	count := txn.count + 1
 	// Extra bytes for the version in key.
-	// Choossing not to use e.estimateSize here since the value returned is inaccurate when e.Value > ValueThreshold.
-	// The entire entry ends up being stored in txn.pendingWrites so this is a more conservative approach that avoids OOM
-	// errors in the case that the user is writing lots of large values in a txn.
-	size := txn.size + int64(len(e.Key)+len(e.Value)+2) + 10
-	if count >= txn.db.opt.maxBatchCount || size >= txn.db.opt.maxBatchSize {
+	size := txn.size + int64(e.estimateSize()) + 10
+	if count >= txn.db.opt.maxBatchCount || size >= txn.db.opt.MaxBatchSize {
 		return ErrTxnTooBig
 	}
 	txn.count, txn.size = count, size
--- a/txn.go
+++ b/txn.go
@@ -295,7 +295,10 @@ func (txn *Txn) newPendingWritesIterator(reversed bool) *pendingWritesIterator {
 func (txn *Txn) checkSize(e *Entry) error {
 	count := txn.count + 1
 	// Extra bytes for the version in key.
-	size := txn.size + int64(e.estimateSize(txn.db.opt.ValueThreshold)) + 10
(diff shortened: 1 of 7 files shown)


txn.go
```diff
@@ -295,7 +295,10 @@ func (txn *Txn) newPendingWritesIterator(reversed bool) *pendingWritesIterator {
 func (txn *Txn) checkSize(e *Entry) error {
 	count := txn.count + 1
 	// Extra bytes for the version in key.
-	size := txn.size + int64(e.estimateSize(txn.db.opt.ValueThreshold)) + 10
+	// Choossing not to use e.estimateSize here since the value returned is inaccurate when e.Value > ValueThreshold.
```
golangcibot: line is 114 characters (from `lll`)

txn.go
```diff
@@ -295,7 +295,10 @@ func (txn *Txn) newPendingWritesIterator(reversed bool) *pendingWritesIterator {
...
+	// The entire entry ends up being stored in txn.pendingWrites so this is a more conservative approach that avoids OOM
```
golangcibot: line is 118 characters (from `lll`)

adamstruck:
The unit tests are passing on my local machine, not sure why they are failing on Travis-CI.

Issue #1144, opened 2019-12-05
Cannot open encrypted DB after writing with StreamWriter


danielmai:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.13.3 linux/amd64
</pre>

### What version of Badger are you using?
Badger v2.0.0

### Does this issue reproduce with the latest master?
Yes (407e5bde06c44e5503dc1eaf5a11cc9f2f348ac4)

### What are the hardware specifications of the machine (RAM, OS, Disk)?
Ubuntu Linux

### What did you do?

I'm using StreamWriter to write to badger with encryption enabled.

```
package main

import (
	"flag"
	"fmt"
	"log"

	"github.com/dgraph-io/badger/v2"
	"github.com/dgraph-io/badger/v2/pb"
)

var (
	encrypt = flag.Bool("encrypt", true, "Enable encryption")
)

func check(err error, msg string) {
	if err != nil {
		log.Fatalf("%s: %v\n", msg, err)
	}
}

func main() {
	flag.Parse()
	opts := badger.DefaultOptions("/tmp/badger-test")
	log.Printf("Using encryption: %v\n", *encrypt)
	if *encrypt {
		opts = opts.WithEncryptionKey([]byte("badgerkey16bytes"))
	}
	db, err := badger.Open(opts)
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	key := []byte("mykey")
	value := []byte("myvalue")

	list := &pb.KVList{}
	list.Kv = append(list.Kv, &pb.KV{
		Key:     key,
		Value:   value,
		Version: 20,
	})
	fmt.Printf("%+v\n", list.Kv)

	sw := db.NewStreamWriter()
	check(sw.Prepare(), "Prepare failed")
	check(sw.Write(list), "Write failed")
	check(sw.Flush(), "Flush failed")

	var val string
	err = db.View(func(txn *badger.Txn) error {
		item, err := txn.Get(key)
		if err != nil {
			return err
		}
		return item.Value(func(b []byte) error { val = string(b); return nil })
	})
	check(err, "Error while retrieving key")
	fmt.Printf("Value: %s\n", val)
}
```

### What did you expect to see?

Running the program once works. Running it a second time to re-open the existing encrypted DB gives an error. I expect to successfully open the DB the second time around.

### What did you see instead?

After a second run, the DB fails to open with an error message related to protos:

```
2019/12/05 15:00:45 proto: illegal wireType 7

[...]

PR #1146, opened 2019-12-06
Fix encryption in stream writer

jarifibrahim:
We store information about tables in the table manifest. When stream
writer would create a new table, it wouldn't store the key ID for the
respective table in the manifest file. This commit fixes it.

Fixes https://github.com/dgraph-io/badger/issues/1144

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1146)


manishrjain:
:lgtm:

PR #1149, opened 2019-12-08
Separate darwin syscalls for mmap

djdv:
Using the `unix` package by default for other systems instead of `syscall`.

This way when systems get added to `unix`, they will work without change.
And the special case is restricted only to darwin builds.

I encountered problems building on Solaris prior to this patch:
`github.com/dgraph-io/badger/y/mmap_unix.go:57:30: undefined: syscall.SYS_MADVISE`
But `unix.Madvise` works fine on that platform (and others).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1149)

### DIFF
--- a/y/mmap.go
+++ b/y/mmap.go
@@ -1,4 +1,4 @@
-// +build !windows,!darwin
+// +build !windows
 
 /*
  * Copyright 2017 Dgraph Labs, Inc. and Contributors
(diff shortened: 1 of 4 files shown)


djdv:
Missed that one, should be good now. :^)
I also changed the `_windows.go` file so that it's consistent / builds with the interface in `mmap.go` like the others do.

PR #1152, opened 2019-12-10
Backport fix for issue #1090 Data Loss during GC

drumwhizz:
This is an experiment, please tell me if this is absolutely crazy but I have run the tests and they still pass. I know I could wait until 2.0.0 is officially released.

We have been seeing data loss using 1.6.0 and wondered whether the fix #1090 would work despite it using options introduced in 2.0. 

To reproduce on 1.6.0, I create  a key with a default value, I then update the key with a different value and then call close and the key disappears. It looks like its the Garbage collection in the close that is doing it. If I update the key a second time with a different value before calling close they key persists.

	options = badger.DefaultOptions(storageDir).WithNumVersionsToKeep(0).WithMaxTableSize(32 << 20).WithValueLogFileSize(640 << 20).WithSyncWrites(true)

These are the options we are using. It would appear that if I remove WithNumVersionsToKeep(0) the data loss stops happening but the DB directory keeps growing due to the size of the vlog file.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1152)

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -899,7 +899,8 @@ func (s *levelsController) close() error {
 }
 
 // get returns the found value if any. If not found, we return nil.
-func (s *levelsController) get(key []byte, maxVs *y.ValueStruct, startLevel int) (y.ValueStruct, error) {
+func (s *levelsController) get(key []byte, maxVs *y.ValueStruct, startLevel int) (
+	y.ValueStruct, error) {
 	// It's important that we iterate the levels from 0 on upward.  The reason is, if we iterated
 	// in opposite order, or in parallel (naively calling all the h.RLock() in some order) we could
 	// read level L's tables post-compaction and level L+1's tables pre-compaction.  (If we do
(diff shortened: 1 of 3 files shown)


levels.go
```diff
@@ -899,14 +899,18 @@ func (s *levelsController) close() error {
 }
 
 // get returns the found value if any. If not found, we return nil.
-func (s *levelsController) get(key []byte, maxVs *y.ValueStruct) (y.ValueStruct, error) {
+func (s *levelsController) get(key []byte, maxVs *y.ValueStruct, startLevel int) (y.ValueStruct, error) {
```
golangcibot: line is 105 characters (from `lll`)

drumwhizz:
This has fixed the data loss issue we are seeing in v1.6.0 but with two failures on Unit tests due to tests timing out (unless you think its somethoing obvious), we will probably hang on until 2.0.0 is released.

jarifibrahim:
Hey @drumwhizz , you're losing data because you've set `number of versions to keep=0`. Compaction is removing the keys. Why are you setting it to zero?

PR #1155, opened 2019-12-10
Add Nancy to project

DarthHater:
Being we found the golang.org/x/crypto transitive issue with `nancy`, I figured we'd craft a PR for y'all that uses `nancy` if you'd like. 

There is functionality in `nancy` to exclude vulnerabilities, etc... as well, so this shouldn't cause too much burden overall, just some new additional info on whether something is vulnerable, or including packages that might be.

Thanks again for creating/working on Badger!

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1155)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,12 @@ env:
 
 before_script:
 - go get github.com/mattn/goveralls
+- cd /tmp && mkdir tools && cd -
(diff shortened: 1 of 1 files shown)


jarifibrahim:
Closing because of inactivity.

PR #1173, opened 2019-12-25
Remove global rand lock in skiplist

jarifibrahim:
The math/rand package (https://golang.org/src/math/rand/rand.go) uses a global lock to allow concurrent access to the rand object. We do not write to skiplist concurrently and hence we don't need a lock at all.

This blog has some interesting details https://blog.sgmansfield.com/2016/01/the-hidden-dangers-of-default-rand/

The PR replaces `math.Rand` with `ristretto/z.FastRand()`. `FastRand` is much faster than `math.Rand` and `rand.New(..)` generators.

The change improves concurrent writes to skiplist by ~30%
```go
func BenchmarkWrite(b *testing.B) {
	value := newValue(123)
	l := NewSkiplist(int64((b.N + 1) * MaxNodeSize))
	defer l.DecrRef()
	b.ResetTimer()
	b.RunParallel(func(pb *testing.PB) {
		rng := rand.New(rand.NewSource(time.Now().UnixNano()))
		for pb.Next() {
			l.Put(randomKey(rng), y.ValueStruct{Value: value, Meta: 0, UserMeta: 0})
		}
	})
}
```
```
name      old time/op  new time/op  delta
Write-16   657ns ± 3%   441ns ± 1%  -32.94%  (p=0.000 n=9+10)
```

This PR also fixes the `test timed out after 10 minutes` issue on teamcity (see build https://teamcity.dgraph.io/viewLog.html?buildId=36889&buildTypeId=Badger_UnitTests#testNameId521499096021883325 )
The problem here is that the test `TestValueGCManaged` tries to create many bufferes with random data
https://github.com/dgraph-io/badger/blob/ea01d380cedea306b4c9e0e04d072c60cd55e973/value_test.go#L119-L124

The `rand.Read(...)` acquires a global lock (https://golang.org/src/math/rand/rand.go) and since `N=10,000` in the above test, it leads to starvation. We would have too many `rand.Read(...)` calls waiting to acquire the same lock. At the same time, every insert to skip list also needs to acquire the same lock 
https://github.com/dgraph-io/badger/blob/ea01d380cedea306b4c9e0e04d072c60cd55e973/skl/skl.go#L168-L174

Hence, the test wouldn't complete in 10 minutes and the build fails.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1173)

### DIFF
--- a/skl/skl.go
+++ b/skl/skl.go
@@ -80,7 +80,7 @@ type Skiplist struct {
 	head   *node
 	ref    int32
 	arena  *Arena
-	rng    rand.Source
+	rand   *rand.Rand
 }
 
 // IncrRef increases the refcount
@@ -134,7 +134,7 @@ func NewSkiplist(arenaSize int64) *Skiplist {
 		head:   head,
 		arena:  arena,
 		ref:    1,
-		rng:    rand.NewSource(time.Now().UnixNano()),
+		rand:   rand.New(rand.NewSource(time.Now().UnixNano())),
 	}
 }
 
@@ -170,7 +170,7 @@ func (s *node) casNextOffset(h int, old, val uint32) bool {
 
 func (s *Skiplist) randomHeight() int {
 	h := 1
-	for h < maxHeight && uint32(s.rng.Int63()) <= heightIncrease {
+	for h < maxHeight && s.rand.Uint32() <= heightIncrease {
 		h++
 	}
 	return h
--- a/skl/skl.go
+++ b/skl/skl.go
@@ -34,12 +34,11 @@ package skl
 
 import (
 	"math"
-	"math/rand"
 	"sync/atomic"
-	"time"
 	"unsafe"
 
 	"github.com/dgraph-io/badger/v2/y"
+	"github.com/dgraph-io/ristretto/z"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


balajijinnah:
we write concurrently while updating head i guess

jarifibrahim:
> we write concurrently while updating head i guess

We don't concurrently update the same skip list. When we update head, the memtable being used for new writes is a different one. See this

https://github.com/dgraph-io/badger/blob/ea01d380cedea306b4c9e0e04d072c60cd55e973/db.go#L869-L884

Note how we created a new memtable after pushing the existing one to the flush chan.

coocood:
Why not use `runtime.fastrand`?

jarifibrahim:
> Why not use `runtime.fastrand`?

Interesting. I did some benchmarks and `fastRand` seems to be more than 2X faster than `rand.NewSource(..)`
```
goos: linux
goarch: amd64
pkg: github.com/dgraph-io/ristretto/z
BenchmarkFastRand-16      	1000000000	         0.292 ns/op
BenchmarkRandSource-16    	1000000000	         0.747 ns/op
BenchmarkRandGlobal-16    	 6822332	       176 ns/op
BenchmarkRandAtomic-16    	77950322	        15.4 ns/op
PASS
ok  	github.com/dgraph-io/ristretto/z	4.808s
```
https://github.com/dgraph-io/ristretto/pull/118/files#diff-a9ed67d3dcec1ff808f43c15fe731caeR289-R310

I'll change it. Thanks @coocood !

PR #1186, opened 2020-01-08
Do not stall Level 1

jarifibrahim:
We don't need to stall writes if Level 1 does not have enough space. Level 1 is stored on the disk and it should be okay to have more number of tables (more size) on Level 1 than the `max level 1 size`. These tables will eventually be compacted to lower levels.

This commit changes the following 
- We no longer stall writes if L1 doesn't have enough space.
- We stall L0 only if `KeepL0InMemory` is true.
- Upper levels (L0, L1, etc) get priority in compaction (previously, level with higher priority score would get preference)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1186)

### DIFF
--- a/level_handler.go
+++ b/level_handler.go
@@ -188,7 +188,9 @@ func (s *levelHandler) tryAddLevel0Table(t *table.Table) bool {
 	// Need lock as we may be deleting the first table during a level 0 compaction.
 	s.Lock()
 	defer s.Unlock()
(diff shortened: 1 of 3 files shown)


jarifibrahim:
Travis build failed because of https://github.com/dgraph-io/badger/issues/1197 and https://github.com/dgraph-io/badger/issues/1196 .

Issue #1193, opened 2020-01-14
Replace eventlog with normal logs


jarifibrahim:
Badger uses [`eventLog`](https://godoc.org/golang.org/x/net/trace#EventLog) to keep track of long-running background operations.
 https://github.com/dgraph-io/badger/blob/2a90c665f1e57fdc48256f222ed5d826c554043c/levels.go#L38-L42

https://github.com/dgraph-io/badger/blob/2a90c665f1e57fdc48256f222ed5d826c554043c/value.go#L828-L832

These `eventLogs` are expensive as seen in https://github.com/dgraph-io/badger/issues/938#issuecomment-513964707

All the trace/eventlog calls should be replaced with `logger.Debugf(...)` statements.

gabru-md:
Hi @jarifibrahim 
I've started writing Go quite recently. Can I work on it?

jarifibrahim:
Hey @gabru-md, feel free to work on it. We use https://godoc.org/golang.org/x/net/trace#EventLog which we don't really need. The task here to is to replace all the `eventlog.Printf` with `logger.Debugf(..)` statements.

PR #1203, opened 2020-01-21
changes eventLog with Logger

gabru-md:
Provides a fix for #1193 

@jarifibrahim, please review and let me know If it is correct. If it isn't then please mention what needs to be done so that I can try to do that as well.

Thanks for the issue

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1203)


db.go
```diff
@@ -68,7 +67,7 @@ type DB struct {
 	valueDirGuard *directoryLockGuard
 
 	closers   closers
-	elog      trace.EventLog
+	logger    Logger
```
jarifibrahim: You can remove this. The `options` struct has the `logger`. Look at `db.opt.Logger`.

db.go
```diff
@@ -273,9 +272,9 @@ func Open(opt Options) (db *DB, err error) {
 		}
 	}()
 
-	elog := y.NoEventLog
-	if opt.EventLogging {
-		elog = trace.NewEventLog("Badger", "DB")
+	logger := NilLogger
+	if opt.Logging {
+		logger = defaultLogger
```
jarifibrahim: You can remove the 3 lines above. The `y.NoEventLog` was added because `eventLog` would take up too much of CPU/memory. Now we don't need it.

db.go
```diff
@@ -284,7 +283,7 @@ func Open(opt Options) (db *DB, err error) {
 		writeCh:       make(chan *request, kvWriteChCapacity),
 		opt:           opt,
 		manifest:      manifestFile,
-		elog:          elog,
+		logger:        logger,
```
jarifibrahim: This would go away once you remove `logger` from the `DB` struct :)

levels.go
```diff
@@ -37,7 +37,7 @@ import (
 
 type levelsController struct {
 	nextFileID uint64 // Atomic
-	elog       trace.EventLog
+	logger     Logger
```
jarifibrahim: You can remove this `logger`. The `levelsController` contains `DB` struct which contains the `logger`. 

You can do `levelsController.kv.opt.Logger` where ever you need this.

levels.go
```diff
@@ -77,7 +77,7 @@ func newLevelsController(db *DB, mf *Manifest) (*levelsController, error) {
 	y.AssertTrue(db.opt.NumLevelZeroTablesStall > db.opt.NumLevelZeroTables)
 	s := &levelsController{
 		kv:     db,
-		elog:   db.elog,
+		logger: db.logger,
```
jarifibrahim: This would also go away once you change the `levelsController` struct.

logger.go
```diff
@@ -83,3 +83,15 @@ func (l *defaultLog) Infof(f string, v ...interface{}) {
 func (l *defaultLog) Debugf(f string, v ...interface{}) {
 	l.Printf("DEBUG: "+f, v...)
 }
+
+var NilLogger Logger = nilLogger{}
```
jarifibrahim: The `nilLogger` is of no use to use once you remove the `opt.Logging` options. You can remove everything related to `nilLogger`.

options.go
```diff
@@ -48,7 +48,7 @@ type Options struct {
 	Truncate            bool
 	Logger              Logger
 	Compression         options.CompressionType
-	EventLogging        bool
+	Logging             bool
```
jarifibrahim: This is also not necessary. The user can set the logger via the `opt.Logger` option.

value.go
```diff
@@ -827,7 +827,7 @@ type lfDiscardStats struct {
 
 type valueLog struct {
 	dirPath string
-	elog    trace.EventLog
+	logger  Logger
```
jarifibrahim: You can use `valueLog.db.opt.Logger` . We don't need this `logger`.

jarifibrahim:
I just realized that the logger in badger doesn't support leveled logging. We might have to change the default logger. The Travis build failed because we're logging too much information.

y/watermark.go
```diff
@@ -64,17 +64,13 @@ type WaterMark struct {
 	lastIndex uint64
 	Name      string
 	markCh    chan mark
-	elog      trace.EventLog
+	logger    badger.Logger
 }
 
 // Init initializes a WaterMark struct. MUST be called before using it.
-func (w *WaterMark) Init(closer *Closer, eventLogging bool) {
+func (w *WaterMark) Init(closer *Closer, opt badger.Options) {
```
gabru-md: instead of passing a boolean value I am passing the Options object so that logger can be initialized for the Watermark struct as well.
@jarifibrahim, please see if this is the correct way to do it?

gabru-md: this may cause the circular import error. I'm not clear about how would I do it otherwise. will creating the logger as a different module be a better approach?

jarifibrahim: You can move the `logger` inside the `y` package. The `y` package contains a bunch of utility types.

y/logger.go
```diff
@@ -29,45 +29,16 @@ type Logger interface {
...
-	opt.Logger.Debugf(format, v...)
-}
-
 type defaultLog struct {
 	*log.Logger
 }
 
 var defaultLogger = &defaultLog{Logger: log.New(os.Stderr, "badger ", log.LstdFlags)}
 
+func DefaultLogger() *defaultLog {
```
golangcibot: exported func DefaultLogger returns unexported type *github.com/dgraph-io/badger/v2/y.defaultLog, which can be annoying to use (from `golint`)

jarifibrahim:
This is a useful PR but we cannot merge it until we figure out a way to add leveled logging. Without leveled logging this PR would generate too much of log.

jarifibrahim:
This PR can be merged after https://github.com/dgraph-io/badger/pull/1249 is merged.

jarifibrahim:
Thanks for the PR @gabru-md :tada:

PR #1206, opened 2020-01-27
Improve log garbage collection performance in LSM-only mode

damz:
When running in LSM-only mode, purging the value log is vital, but extremely inefficient because the process requires loading each entry from the LSM Tree, just to realize that the entry doesn't actually point to the value log.

Improve this by consistently storing the `bitValuePointer` meta into the value log, and skipping the LSM tree for entries that are not pointers.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1206)

### DIFF
--- a/value.go
+++ b/value.go
@@ -510,6 +510,11 @@ func (vlog *valueLog) rewrite(f *logFile, tr trace.Trace) error {
 			tr.LazyPrintf("Processing entry %d", count)
 		}
 
+		if e.meta&bitValuePointer == 0 {
+			// Not a pointer to the value log, avoid going to the LSM Tree.
+			return nil
+		}
+
 		vs, err := vlog.db.get(e.Key)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


value.go
```diff
@@ -1408,6 +1413,11 @@ func (vlog *valueLog) write(reqs []*request) error {
 				b.Ptrs = append(b.Ptrs, valuePointer{})
 				continue
 			}
+
+			if !vlog.db.shouldWriteValueToLSM(*e) {
+				e.meta |= bitValuePointer
+			}
```
damz: RFC: maybe we should clear the bitValuePointer in an `else`?

damz:
Actually, a problem with this is that it changes the on-disk format. The `bitValuePointer` bit was usually not stored in the value log before that change (except, I think, in the `StreamWriter` code path).

PR #1213, opened 2020-02-07
Replace t.Fatal with require.NoError in tests

jarifibrahim:
We are using the following pattern in tests that can be replaced with `require.NoError(t, err)`
```go
if err != nil {
    t.Fatal(err)
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1213)


jarifibrahim:
The appveyor build failed because of the flaky `ExampleDB_Subscrbe` test. It should be fixed by https://github.com/dgraph-io/badger/pull/1214

PR #1215, opened 2020-02-07
Introduce minimal blob file

jarifibrahim:
This commit introduces a minimal blob file. All reads and writes are done through the blob file.

Earlier we used to store only value pointers in the skip list, now we store actual values.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1215)

### DIFF
--- a/blob.go
+++ b/blob.go
@@ -137,9 +137,12 @@ type blobManager struct {
 	dirPath   string
 }
 
-func (bm *blobManager) Open(db *DB, opt Options) error {
+func (bm *blobManager) Open(opt *Options) error {
+	if opt.InMemory {
+		return nil
+	}
 	// TODO - Figure out if this should be valueDir or just Dir.
-	bm.dirPath = db.opt.Dir
+	bm.dirPath = opt.Dir
 	bm.fileList = make(map[uint32]*blobFile)
 
 	fileInfos, err := ioutil.ReadDir(bm.dirPath)
--- a/blob.go
+++ b/blob.go
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2020 Dgraph Labs, Inc. and Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


db.go
```diff
@@ -340,6 +341,9 @@ func Open(opt Options) (db *DB, err error) {
 		}()
 	}
 
+	// TODO - Open BlobManager
+	db.blobManager.Open(db, db.opt)
```
golangcibot: Error return value of `db.blobManager.Open` is not checked (from `errcheck`)

blob.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// Finish finishes a blob file and opens the built blob file.
+func (bfb *blobFileBuilder) finish() (*blobFile, error) {
+	name := filepath.Join(bfb.dir, newBlobFileName(bfb.fid))
+	file, err := os.OpenFile(name, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0666)
+	if err != nil {
+		return nil, errors.Wrapf(err, "finish blob file: %d", bfb.fid)
+	}
+	if _, err := file.Write(bfb.buf.Bytes()); err != nil {
```
balajijinnah: Reuse the buffer

blob.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil, errors.Wrapf(err, "failed to write to blob file: %d", bfb.fid)
+	}
+	if err = file.Close(); err != nil {
+		return nil, err
+	}
+	return newBlobFile(file.Name(), bfb.fid)
```
balajijinnah: Do it explicit

blob.go
```diff
@@ -0,0 +1,216 @@
...
+	maxFileID uint32
+	dirPath   string
+}
+
+func (bm *blobManager) Open(opt *Options) error {
+	if opt.InMemory {
+		return nil
+	}
+	// TODO - Figure out if this should be valueDir or just Dir.
+	bm.dirPath = opt.Dir
```
balajijinnah: I would suggest to keep blob file in valueDir

db.go
```diff
@@ -898,24 +890,46 @@ func arenaSize(opt Options) int64 {
 	return opt.MaxTableSize + opt.maxBatchSize + opt.maxBatchCount*int64(skl.MaxNodeSize)
 }
 
+func (db *DB) newBlobFileBuilder() *blobFileBuilder {
+	return newBlobFileBuilder(db.blobManager.allocFileID(), db.opt.Dir)
```
balajijinnah: You can reuse the buffer here.

PR #1216, opened 2020-02-07
Fix int overflow for 32bit

### LINKED ISSUE #1182: Int Overflow Bug Reoccurs - Please Add Better Testing
This is the second time in less than two months that an issue with integer overflow happens, breaking builds on 32-bit platforms.  Given that this is the second occurrence of such a bug, I think testing needs to be improved to catch this, because if it has already happened twice in 2 months, it will probably happen a 3rd, and a 4th if testing isn't improved to catch this.

Previous bug report https://github.com/dgraph-io/badger/issues/1142, commit fixing the first time the bug happened https://github.com/dgraph-io/badger/commit/fb0cdb858c78004b9db8f08e91bb713780b20445, and the most recent commit to fix the same issue https://github.com/dgraph-io/badger/commit/465f28aba5fee9a4dac132ccbebaaae6a1d711dd

vardhanapoorv:
Fixes #1182

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1216)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ go:
   - tip
 os:
   - osx
+  - linux
 env:
   jobs:
   - GOARCH=386
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ go:
   - tip
 os:
   - osx
-  - linux
 env:
   jobs:
   - GOARCH=386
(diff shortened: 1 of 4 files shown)


jarifibrahim:
@vardhanapoorv Travis-CI didn't run on this PR. Try pushing an empty commit. 
```
git commit --allow-empty && git push
 ```

Issue #1209, opened 2020-02-04
rarely, nil pointer panic in doCompact go-routine when a badger open.


peter-kyu:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.11.5 linux/amd64
</pre>

### What version of Badger are you using?
v1.5.4

### Does this issue reproduce with the latest master?
i think it will happen too.

### What are the hardware specifications of the machine (RAM, OS, Disk)?
128g/CentOS Linux release 7.6.1810 (Core)/ssd

### What did you do?
just open a badger db as fileio mode.
i used that option.
```
		doNotCompactOpt:            false,
		levelOneSizeOpt:            400,
		levelSizeMultiplierOpt:     10,
		maxLevelsOpt:               7,
		maxTableSizeOpt:            32,
		numCompactorsOpt:           3,
		numLevelZeroTablesOpt:      5,
		numLevelZeroTablesStallOpt: 10,
		numMemtablesOpt:            1,
		valueLogFileSizeOpt:        128,
		valueThresholdOpt:          100,
		syncWritesOpt:              true,
		tableLoadingModeOpt:        int(options.FileIO), 
		valueLogLoadingModeOpt:     int(options.FileIO), 
		truncateOpt:                true,             

```
and that time vlog and sst files size were almost 100g.

### What did you expect to see?
open it successfully

### What did you see instead?
really rarely get panic in case of go routine scheduling by golang.
this is the panic log i've got.
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x94914e]
goroutine 2087487 [running]:
go/src/indexer/vendor/github.com/dgraph-io/badger.(*valueLog).updateGCStats(0xc0033359c8, 0xc028ef8060)
        /home/go/src/indexer/vendor/github.com/dgraph-io/badger/value.go:1233 +0x3e
go/src/indexer/vendor/github.com/dgraph-io/badger.(*levelsController).compactBuildTables(0xc00016a070, 0x1, 0xe7bba0, 0xc0002024e0, 0xc03edec060, 0xc03edec0c0, 
0xc0002823a8, 0x1, 0x1, 0xc012ae63a0, ...)
        /home/go/src/indexer/vendor/github.com/dgraph-io/badger/levels.go:501 +0x1460
[...]

PR #1226, opened 2020-02-24
Initialize vlog before starting compactions in db.Open

jarifibrahim:
If compaction is started before opening vlog, badger might crash with
nil pointer error when trying to push discard stats at the end of
compactions because discard stats are initialized in vlog open.

Fixes https://github.com/dgraph-io/badger/issues/1209

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1226)

### DIFF
--- a/db.go
+++ b/db.go
@@ -331,6 +331,19 @@ func Open(opt Options) (db *DB, err error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


jarifibrahim:
@manishrjain has reviewed and approved this.

Issue #1239, opened 2020-03-03
Decompression uses too much memory


jarifibrahim:
Badger uses compression to reduce disk space. When the block are decompressed, it takes up too much memory. 
https://github.com/dgraph-io/badger/blob/5b4c0a65cdf94a30f63b87c7b1db7931ee6d4f4c/table/table.go#L625-L629

The crux of the problem is that the decompression call allocates new block every time we want to decompress an existing block. An ideal fix here would be to find a way to reuses the blocks of memory.
![image](https://user-images.githubusercontent.com/12949454/75759280-4d639380-5d5b-11ea-94f5-a57df6eb6154.png)

Taken from chat: https://dgraph.slack.com/archives/C13LH03RR/p1583223673220200?thread_ts=1583216894.212400&cid=C13LH03RR

jarifibrahim:
I've tried two different approaches to reduce the memory but both of them aren't working.

**Attempt 1** - PR https://github.com/dgraph-io/badger/pull/1247 Reuse the original byte slice (the one read from the table). The problem with this approach is that the sst is mmap and if the slice of bytes read from the table is put into the sync pool, we might end up modifying the original contents of the file. This was seen in https://github.com/dgraph-io/badger/pull/1247 .

**Attempt 2** - PR https://github.com/dgraph-io/badger/pull/1308 . Attempt to reuse the byte slice after the block iterator is done processing it. This also has some strange issue which I haven't been able to figure out. In simple terms, the contents of block change if the slices are reused. This is definitely an issue with the implementation in https://github.com/dgraph-io/badger/pull/1308 but I haven't been able to fix it. I've spent significant time trying to debug this but haven't found the issue yet.

PR #1247, opened 2020-03-05
Reuse buffer for decompression

jarifibrahim:
This commit uses a slice pool to store buffers that can be reused for decompression.

Fixes https://github.com/dgraph-io/badger/issues/1239

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1247)

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -88,6 +88,7 @@ type Table struct {
 
 	fd        *os.File // Own fd.
 	tableSize int      // Initialized in OpenTable, using fd.Stat().
+	buf       []byte   // Used for decompression.
 
 	blockIndex []*pb.BlockOffset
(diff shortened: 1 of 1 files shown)


jarifibrahim:
The code in this PR won't work. The `[]byte` slice we're inserting into the sync pool is read from the mmaped sst file and any changes to the `data` byte slice leads to modification of the `mmapped` buffer of the sst.

The issue is similar to the issue seen in the following snippet. 
```go
func do(b []byte) {
	copy(b, []byte("d"))
}

func main() {
	b := []byte("foo")
	fmt.Println(string(b)) // foo
	do(b)
	fmt.Println(string(b)) // doo
}

```

Issue #1255, opened 2020-03-11
v2 much slower then v1.6.0


adwinsky:
### What version of Go are you using (`go version`)?

<pre>
$ go version
1.13.8
</pre>

### What version of Badger are you using?

v2.0.2
opts := badger.DefaultOptions(fmt.Sprintf(dir + "/" + name))
opts.SyncWrites = false
opts.ValueLogLoadingMode = options.FileIO

### Does this issue reproduce with the latest master?

Yes

### What are the hardware specifications of the machine (RAM, OS, Disk)?

Samsung nvme disk SAMSUNG MZPLL3T2HMLS-00003 
Linux: 18.04.2 LTS (Bionic Beaver)

### What did you do?

As I mentioned in other ticket https://github.com/dgraph-io/badger/issues/1228 I am unable to use Badger v2 version because it's very slow. 

### What did you expect to see?

I expect Badger v2 version to be  similarly fast as v1.6.0

### What did you see instead?

One great example is when I run RunValueLogGC function it can take up to 20 minutes. I have rather database (about 10 GB), and with v1.6.0 same operation takes up to 30s.

Have a look at debug/request output:

<pre>
When | Elapsed (s)
-- | --
2020/03/11 13:37:28.017837 | 464.393672 | GC
13:37:28.017850 | .    13 | ... Could not find candidate via discard stats. Randomly picking one.
13:37:28.017857 | .     8 | ... Randomly chose fid: 1
13:37:28.017865 | .     7 | ... Size window: 125606803.00. Count window: 10000.
13:37:28.017866 | .     2 | ... Skip first 324.58 MB of file of size: 1197 MB
13:37:28.906299 | .888433 | ... Stopping sampling after 10000 entries.
13:37:28.906306 | .     7 | ... Fid: 1. Skipped: 324.58MB Num iterations: 260627. Data  status={total:20.11733341217041 discard:19.84939193725586 count:10001}
13:37:28.906333 | .    28 | ... Rewriting fid: 1
13:43:47.537655 | 378.631322 | ... Processing entry 100000
</pre>

Processing 100000 entries takes more than 6 minutes! I noticed most of the time is spend  here:

https://github.com/dgraph-io/badger/blob/master/value.go#L517

Very often it takes 20-50ms for single lookup. I think this also explains why overall performance of my application is significantly slower when I switch from 1.6.0 to 2

Here is the output of pprof profiling 

<pre>
Showing top 10 nodes out of 49
[...]

Issue #1254, opened 2020-03-09
Performance regression 1.6 to 2.0.2


dougdonohoe:
### What version of Go are you using (`go version`)?

go version go1.12.7 darwin/amd64

### What version of Badger are you using?

2.0.2 (upgrading from 1.6.0)

### Does this issue reproduce with the latest master?

Haven't tried.

### What are the hardware specifications of the machine (RAM, OS, Disk)?

GCP 8 CPU (Intel Haswell), 32 GB RAM, 750 GB local ssd

### What did you do?

Running code which extracts data from Kafka and saves to Badger DB.  I'm running on exact same hardware, disk and my code against exact same Kafka topic.

### What did you expect to see?

Better or equal performance with Badger 2.

### What did you see instead?

Severe slowdown after writing ~1,461,000 records.  See below

#### 1.6.0 performance:

Performance in 1.6.0 takes about 300-400ms to extract 1000 messages.

```
  Up to offset 1453000 Time[330ms] Events[1453001] UrisCreated[1975] PathsCreated[0] Bytes[11.3 MiB] TotalBytes[11.7 GiB] Date[2014-11-10T07:35:07.000] EstTimeToFinish[4h17m58s]
  Up to offset 1454000 Time[360ms] Events[1454001] UrisCreated[1954] PathsCreated[0] Bytes[11.2 MiB] TotalBytes[11.7 GiB] Date[2014-11-10T11:31:43.000] EstTimeToFinish[4h18m1s]
  Up to offset 1455000 Time[340ms] Events[1455001] UrisCreated[1969] PathsCreated[0] Bytes[11.0 MiB] TotalBytes[11.7 GiB] Date[2014-11-10T15:33:31.000] EstTimeToFinish[4h18m4s]
  Up to offset 1456000 Time[360ms] Events[1456001] UrisCreated[1789] PathsCreated[0] Bytes[13.3 MiB] TotalBytes[11.7 GiB] Date[2014-11-10T20:46:14.000] EstTimeToFinish[4h18m7s]
  Up to offset 1457000 Time[320ms] Events[1457001] UrisCreated[1720] PathsCreated[0] Bytes[13.0 MiB] TotalBytes[11.7 GiB] Date[2014-11-11T06:56:07.000] EstTimeToFinish[4h18m9s]
[...]

Issue #1248, opened 2020-03-05
Potential Read Performance Regression


connorgorman:
### What version of Go are you using (`go version`)?

<pre>
$ go version
1.13.4
</pre>

### What version of Badger are you using?
master (https://github.com/dgraph-io/badger/commit/bce069c3875ec704fee689857d47d241199daf46)

### Does this issue reproduce with the latest master?
yes

### What are the hardware specifications of the machine (RAM, OS, Disk)?
8GB, Ubuntu, 100GB HDD

### What did you do?

Running 1.6.0 vs master with a mixed workload doing mostly writes and random reads
Value sizes range from 400 to 1000 bytes

### What did you expect to see?

I'd expect the performance of master to be better than the performance of 1.6.0 across the board. Instead, it seems that random reads have more variability in speed.

### What did you see instead?

 I've seen slow downs of up to 200x. Unfortunately, I haven't been able to provide an isolated code sample that I could share, but was curious if anything has shown up in any benchmarks.

jarifibrahim:
@connorgorman were you running badger with default options? You might want to disable compression, encryption, and cache and then run your benchmarks.

connorgorman:
@jarifibrahim I was running with the default options, but with NumLevelZeroTables = 2 and NumLevelZeroTablesStall = 5. I believe those other features are off by default (except the cache), but I will set the flags and comment again. Thanks

connorgorman:
@jarifibrahim So I ran it again with
```
badger.DefaultOptions(path).
		WithTruncate(true).
		WithNumLevelZeroTables(2).
		WithNumLevelZeroTablesStall(5).
		WithMaxCacheSize(0).
		WithChecksumVerificationMode(options.NoVerification).
		WithCompression(options.None).
		WithVerifyValueChecksum(false)
```
and I'm getting similar results in terms of performance regressions. In fact, it looks like reads and writes are generally slower (though some cases appear faster).

Are there any internal metrics in Badger I can use? Do all those options make sense?

jarifibrahim:
@connorgorman The options make sense. Let me try to run some benchmarks and see what results I get. Ideally, v2 should be faster than v1.6. 

It would be useful if you can share your benchmarks.

jarifibrahim:
Hey @connorgorman, I've raised https://github.com/dgraph-io/badger/pull/1256 to fix this. 
Major performance improvement comes from https://github.com/dgraph-io/badger/pull/1256 and #1257  also improves the performance.

[...]

PR #1204, opened 2020-01-21
Add support for caching bloomfilters

jarifibrahim:
This PR adds support for caching bloom filters in ristretto.
The bloom filters and blocks are removed from the cache when the table is deleted.

Fixes https://github.com/dgraph-io/badger/issues/791

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1204)

PR #1256, opened 2020-03-12
Revert "Add support for caching bloomfilters (#1204)"

jarifibrahim:
This reverts commit 4676ca96f15b9e9a3b1bc163b1a2a908bde553f0.

Reading bloom filters from the cache leads to severe performance degradation.
Result of read bench tool with `key size 32 bytes` and `value size 128 bytes` (Note: data is not compressed or encrypted)
```
1. With bloom filters in cache
	Average read speed		: 97.37 KB/s
	Total bytes read in 1 minute	: 3.7 MB

2. With bloom filters stored in memory (no cache)
	Average read speed		: 41.3833 MB/s
	Total bytes read in 1 minute	: 2.4 GB
```

The performance degradation is a result of the way ristretto handles item eviction from the cache. For instance, the following was the state of cache after opening 65 tables

```
badger 2020/03/12 18:01:41 INFO: All 65 tables opened in 1.034s

hit: 0              keys-added: 194     keys-evicted: 172              
miss: 130           keys-updated: 1     cost-added: 7911996649                
gets-kept: 0        gets-dropped: 0     cost-evicted: 6942514434 
sets-dropped: 0     sets-rejected: 0    gets-total: 130
hit-ratio: 0.00
```

The important metric here is the `keys-added: 194 and keys-evicted: 172`. This means we have only `22` items in the cache, which includes the bloom filters and the sst blocks. So most of the time the bloom filter has to be read from the SST which is very expensive.

Fixes https://github.com/dgraph-io/badger/issues/1255, https://github.com/dgraph-io/badger/issues/1254 and https://github.com/dgraph-io/badger/issues/1248

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1256)

### DIFF
--- a/table/builder_test.go
+++ b/table/builder_test.go
@@ -57,7 +57,7 @@ func TestTableIndex(t *testing.T) {
 		keysCount := 10000
 		for _, opt := range opts {
 			builder := NewTableBuilder(opt)
-			filename := fmt.Sprintf("%s%c%d.sst", os.TempDir(), os.PathSeparator, rand.Uint32())
+			filename := fmt.Sprintf("%s%c%d.sst", os.TempDir(), os.PathSeparator, rand.Int63())
 			f, err := y.OpenSyncedFile(filename, true)
 			require.NoError(t, err)
 
--- a/table/table.go
+++ b/table/table.go
@@ -18,7 +18,6 @@ package table
 
 import (
 	"crypto/aes"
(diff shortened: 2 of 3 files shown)


jarifibrahim:
This PR is no longer necessary. The issue has been fixed by https://github.com/dgraph-io/badger/pull/1260

PR #1260, opened 2020-03-16
Add separate cache for bloomfilters

jarifibrahim:
https://github.com/dgraph-io/badger/pull/1256 showed that a single
cache might not be enough to store the data blocks and the bloom
filters.
This commit adds a separate cache for the bloom filters. This
commit also adds a new flag `LoadBloomOnOpen` which determines
if the bloom filters should be loaded when the table is opened on
or not.

The default value of `MaxBfCacheSize` is `zero` and 
`LoadBloomOnOpen` is true.

This change has significant performance improvement on read speeds
because a single cache would lead to bloom filter eviction and we
would read the bloomfilter from the disk.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1260)

### DIFF
--- a/db.go
+++ b/db.go
@@ -92,6 +92,7 @@ type DB struct {
 	pub        *publisher
 	registry   *KeyRegistry
 	blockCache *ristretto.Cache
+	bfCache    *ristretto.Cache
 }
 
 const (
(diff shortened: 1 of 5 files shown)


db.go
```diff
@@ -304,7 +305,21 @@ func Open(opt Options) (db *DB, err error) {
...
+
+	if opt.MaxBfCacheSize > 0 {
+		config := ristretto.Config{
+			// Use 5% of cache memory for storing counters.
+			NumCounters: int64(float64(opt.MaxBfCacheSize) * 0.05 * 2),
+			MaxCost:     int64(float64(opt.MaxBfCacheSize) * 0.95),
+			BufferItems: 64,
+			Metrics:     true,
+		}
+		db.blockCache, err = ristretto.NewCache(&config)
```
nick-jones, line 320: Shouldn't this be `db.bfCache = ...`?

Issue #1263, opened 2020-02-06
Consider using xorfilter to replace bloomfilter


choleraehyq:
From its paper, xorfilters are faster and more concise alternative to bloomfilters.
https://github.com/FastFilter/xorfilter

jarifibrahim:
This looks like a nice improvement over the bloom filters used in badger. I'll run some benchmarks to confirm the performance.

jarifibrahim:
Using xor filters is a breaking change and the performance benefit is not worth breaking the on-disk format. See https://github.com/dgraph-io/badger/pull/1264 .

PR #1264, opened 2020-03-19
Use XorFilter instead of Bloomfilter

jarifibrahim:
Fixes https://github.com/dgraph-io/badger/issues/1263 

This PR proposes use of https://github.com/FastFilter/xorfilter instead of `bbloom` https://github.com/dgraph-io/ristretto/blob/master/z/bbloom.go .

Following are the performance benchmarks
```
name                           time/op
Set/bloom-8              23.2ms ± 3%
Set/xorFilter-8           121ms ± 4%
Get/success/bloom-8      20.9ns ± 5%
Get/success/xorFilter-8  7.41ns ± 7%
Get/failure/bloom-8      8.27ns ± 6%
Get/failure/xorFilter-8  7.52ns ± 5%
```
The size xor filter is lesser than bbloom.
```
Number of keys = 2070594 (~ 2 million)
Size of bloom filter  	= 5592436 (5.59 MB)
Size of xor filter	= 3395912 (3.39 MB)
```

Xorfilter is definitely faster than bbloom filter but there isn't a significant performance difference in `key iterator` or `get` operation.
```
name           old time/op  new time/op  delta
KeyIterator-8   123µs ± 7%   121µs ± 7%   ~     (p=0.436 n=10+10)
```

Benchmarks are stored in `builder_test.go` and `iterator_test.go`

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1264)

### DIFF
--- a/table/builder.go
+++ b/table/builder.go
@@ -18,11 +18,13 @@ package table
 
 import (
 	"crypto/aes"
+	"encoding/json"
 	"math"
 	"runtime"
 	"sync"
 	"unsafe"
 
+	"github.com/FastFilter/xorfilter"
 	"github.com/dgryski/go-farm"
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/snappy"
@@ -31,7 +33,6 @@ import (
 	"github.com/dgraph-io/badger/v2/options"
 	"github.com/dgraph-io/badger/v2/pb"
 	"github.com/dgraph-io/badger/v2/y"
-	"github.com/dgraph-io/ristretto/z"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


table/builder.go
```diff
@@ -372,12 +373,17 @@ The table structure looks like
 */
 // In case the data is encrypted, the "IV" is added to the end of the index.
 func (b *Builder) Finish() []byte {
-	bf := z.NewBloomFilter(float64(len(b.keyHashes)), b.opt.BloomFalsePositive)
-	for _, h := range b.keyHashes {
-		bf.Add(h)
+	var keySlice = make([]uint64, 0, len(b.keyHashes))
+	for i, _ := range b.keyHashes {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	for i := range b.keyHashes {
```

jarifibrahim:
This is a breaking change and the performance gain is not worth breaking the format. 
The bloom filters are used only to pick the tables which happens when `NewKeyIterator` is called or `txn.Get` is called. The change doesn't have any real performance gains.

PR #1272, opened 2020-03-24
make putNode correct 

zhiqiangxu:
`s.buf` bytes may be not aligned, so the correct implementation should take that into consideration.

**Update** 

Switched to use `makeAlignedBuf` instead.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1272)

### DIFF
--- a/skl/arena.go
+++ b/skl/arena.go
@@ -52,7 +52,7 @@ func newArena(n int64) *Arena {
 
 func makeAlignedBuf(n int64) []byte {
 	buf := make([]byte, n+int64(nodeAlign))
-	buf0Alignment := uint32(uintptr(unsafe.Pointer(&buf[0]))) & ^uint32(nodeAlign)
+	buf0Alignment := uint32(uintptr(unsafe.Pointer(&buf[0]))) & uint32(nodeAlign)
 	buf = buf[buf0Alignment : int64(buf0Alignment)+n]
 	return buf
 }
--- a/skl/arena.go
+++ b/skl/arena.go
@@ -35,9 +35,8 @@ const (
 
 // Arena should be lock-free.
 type Arena struct {
-	n             uint32
-	buf           []byte
-	buf0Alignment uint32
+	n   uint32
+	buf []byte
 }
 
 // newArena returns a new arena.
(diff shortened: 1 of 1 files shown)


skl/arena.go
```diff
@@ -73,8 +73,8 @@ func (s *Arena) putNode(height int) uint32 {
 		l, n, len(s.buf))
 
 	// Return the aligned offset.
-	m := (n - l + uint32(nodeAlign)) & ^uint32(nodeAlign)
-	return m
+	m := (uintptr(unsafe.Pointer(&s.buf[0]))+uintptr(n-l+uint32(nodeAlign))) & ^uintptr(nodeAlign) - uintptr(unsafe.Pointer(&s.buf[0]))
```
golangcibot: line is 132 characters (from `lll`)

zhiqiangxu: updated.

jarifibrahim:
Hey @zhiqiangxu, can you add some more information to the PR description? I'm trying to understand why do we need this change and what are the implications of it.

zhiqiangxu:
node address = address of first element + offset

In order to make sure node address is aligned, should consider `address of first element`, which is not necessarily aligned.

zhiqiangxu:
I missed one thing shamefully: 

The first word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned.

zhiqiangxu:
The spec only guarantees :

the first (64-bit) word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned.

So it's not guaranteed that the first element of a byte slice is 8 byte aligned.

jarifibrahim:
@zhiqiangxu The review for this PR is pending because I don't understand alignment issues properly. I'll read some stuff about it and then review this PR. Apologies for the delay.

jarifibrahim:
@zhiqiangxu I did a small benchmark to compare the performance.
This is the `BenchmarkWrite` function in the `skl` package.
```
name            old time/op    new time/op    delta
Write-8          522ns ± 8%     563ns ± 3%  +7.87%  (p=0.001 n=10+8)
```

The new changes seem to be slower. Do you know why it could be slower?

zhiqiangxu:
Maybe because of the extra [bounding check](https://go101.org/article/bounds-check-elimination.html) involved, but pity that I don't know how to optimize it:(

damz:
The rationale looks correct to me: Golang doesn't guarantee that the underlying memory block of a `[]byte` is aligned. It only guarantees that the underlying memory block of an array follows the alignment guarantees of each of the element:

From https://golang.org/ref/spec#Size_and_alignment_guarantees:

> For a variable x of array type: unsafe.Alignof(x) is the same as the alignment of a variable of the array's element type.

So a `[]uint64` will be aligned to 64 bit on platforms that requires that. But a `[]byte` will not.

Based on this, I can suggest the following:

```golang
func makeAligned(n int64) []byte {
	buf := make([]uint64, n>>3)
	head := (*reflect.SliceHeader)(unsafe.Pointer(&buf))
	head.Len = int(n)
	head.Cap = int(n)
	return *(*[]byte)(unsafe.Pointer(head))
}
```

It doesn't seem to have any measurable performance impact.

harshil-goel:
Keep open

harshil-goel:
From what I can understand we already do bit alignment while we are inserting data using putNode or putVal. They ensure alignment before inserting. Hence alignment while creating the buffer is not required. Let me know if you belive so and reopen the PR.

Issue #981, opened 2019-08-13
Question: Multiple iterators on a read-write transaction.


elliotcourant:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.12.6 darwin/amd64
</pre>

### What version of Badger are you using?
master @ 74ed6da2c776a6b4941596df9da3ff84a0703c70

What is the limiting factor of the transaction or iterator object being thread-safe? Is it the pending writes map/iterator that isn't safe if there are multiple iterators at the same time as data being written? 

I'm trying to run two iterators on a transaction after some entries have been written, but while the two iterators are active no writes will occur. If the limiting factor is writes cannot happen while multiple iterators are active would it be reasonable to simply restrict writes until iterators are deallocated?

Sorry if I'm misunderstanding the issue; trying to understand the workings with that object to better approach my problem.

Also sorry if this is not the correct venue for this question. I'm happy to move it if it is.

ashish-goswami:
@elliotcourant Currently limiting factor is pending write map in transaction. It is not thread safe.
That's the reason, you cannot have multiple iterators in RW transaction.
I am not following the second part you asked. Are you trying to run two iterators in RW transaction? Also are entries are written in same transaction?

elliotcourant:
I am trying to run two iterators but I don't think I will get the performance benefit that I thought that I would. I was mostly just curious about what the limiting factor was. Thank you very much!

manishrjain:
We could easily make the txn run iterators concurrently. Thread unsafe was just seemed simple enough for the utility when originally written.

PR #1286, opened 2020-03-31
Support multiple iterators in read-write transactions.

elliotcourant:
This adds support for multiple iterators during a read-write transaction. But iterators created in a read-write transaction will only be able to see writes that were performed before the iterator was created. Any writes that occur after the iterator is created will be invisible to the iterator.

This references an issue I opened previously: #981

I had closed this issue as I was able to work with a single iterator but I'm not trying to read multiple sequences of keys and values from different areas in the database at the same time, and the most efficient way to do this seems to be with multiple iterators.

Fixes https://github.com/dgraph-io/badger/issues/981

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1286)

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -816,17 +816,19 @@ func TestIterateParallel(t *testing.T) {
 		txn := db.NewTransaction(true)
 		itr := txn.NewIterator(DefaultIteratorOptions)
 		require.NotPanics(t, func() {
-			// Now that multiple iterators are supported in read-write transactions, make sure this does not panic
-			// anymore. Then just close the iterator.
+			// Now that multiple iterators are supported in read-write
+			// transactions, make sure this does not panic anymore. Then just
+			// close the iterator.
 			txn.NewIterator(DefaultIteratorOptions).Close()
 		})
-		// The transaction should still panic since there is still one pending iterator that is open.
+		// The transaction should still panic since there is still one pending
+		// iterator that is open.
 		require.Panics(t, txn.Discard)
 		itr.Close()
 		txn.Discard()
 
-		// (Regression) Make sure that creating multiple concurrent iterators within a read only transaction continues
-		// to work.
+		// (Regression) Make sure that creating multiple concurrent iterators
+		// within a read only transaction continues to work.
 		t.Run("multiple read-only iterators", func(t *testing.T) {
 			// Run multiple iterators for a RO txn.
 			txn = db.NewTransaction(false)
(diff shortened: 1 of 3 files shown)


jarifibrahim:
Hey @elliotcourant, Thank you so much for the PR. The change looks like a simple change. I wonder why were multiple iterators disabled in Read-Write transactions. Maybe @manishrjain would know.

elliotcourant:
Thank you very much for the feedback, I will make those changes!

elliotcourant:
I think change it so that all transactions would be RW would be a bit of a large change in this MR. I will add tests for creating and using multiple iterators in a RW transaction though.

jarifibrahim:
@elliotcourant I think there will be a race condition if we have multiple iterators here

elliotcourant:
@jarifibrahim If you are iterating in multiple go routines then you are absolutely right. I don't think adding a mutex to appending that array would be very bad. I would think that most iterators are in a single thread to begin with, or (at least as of right now) are using read only transactions for multiple iterators. So adding a mutex would not impact any existing behavior. 
As for the performance implications I'm not sure. If the performance of a mutex is a concern then I could try to find another solution?

elliotcourant:
Just pushed the updates from the code review and squashed my commits. 

@jarifibrahim I added a mutex around the addKey method, which should make concurrent reads safe. However it is still currently _possible_ that if one were to try to perform a read on thread A and try to commit on thread B at the same time that a race condition could occur? As thread B would be accessing the read array without a lock while thread A might try to append the read array.

I don't see that happening very often, and it's certainly possible that it could happen now? Since there is nothing stopping a user from having a single iterator on a different thread and trying to commit at the same time. 

I added a sub test to make sure that multiple concurrent iterators are working with a read-write transaction.

Another option to support more "lockless" reads with multiple iterators could be to store the keys read on the iterator itself, and then only when the iterator is closed will it try to push them to the read array on the transaction? This way we could have the commit action close all the active iterators before checking the read array, pushing the work to be done at the commit time, rather than as the reads are being performed?

elliotcourant:
@jarifibrahim @manishrjain Is there any more feedback you have on this PR? Or are there any further changes that you think need to be made?

jarifibrahim:
Hey @elliotcourant, I'll get this merged today. Thanks for your contribution :+1: 
>@jarifibrahim I added a mutex around the addKey method, which should make concurrent reads safe. However, it is still currently possible that if one were to try to perform a read on thread A and try to commit on thread B at the same time that a race condition could occur? As thread B would be accessing the read array without a lock while thread A might try to append the read array.

That won't happen because we acquire lock on oracle before checking for conflicts https://github.com/dgraph-io/badger/blob/9459a240bdf3cf1bf59a172558b37455c742d3bf/txn.go#L167-L174

jarifibrahim:
Thank you for your contribution @elliotcourant . I've merged your PR :tada:

Issue #1228, opened 2020-02-25
GC doesn't work? (not cleaning up SST files properly)


adwinsky:
### What version of Go are you using (`go version`)?

<pre>
$ go version
1.13.8
</pre>

### What version of Badger are you using?

v1.6.0

opts := badger.DefaultOptions(fmt.Sprintf(dir + "/" + name))
opts.SyncWrites = false
opts.ValueLogLoadingMode = options.FileIO

### Does this issue reproduce with the latest master?

With the latest master GC becomes much slower

### What are the hardware specifications of the machine (RAM, OS, Disk)?

2TB NVME drive, 128 GB RAM

### What did you do?

 I have a Kafka topic with 12 partitions. For every partition I create a database. Each database grows quite quickly (about 12*30GB per hour) and the TTL for most of the events is 1h, so the size should stay at constant level. Now for every partition I create a separate transaction and I process read and write operations sequentially, there is no concurrency, when the transaction is getting to big I commit it, in separate go-routine I start RunValueLogGC(0.5). Most of GC runs end up with ErrNoRewrite. Even tried to repeat RunValueLogGC until I have 5 errors in the row, but still I was running out of disk space quite quickly. My current fix is to patch the Badger GC, make it run on every fid that is before the head. This works fine, but eventually becomes slow when I have too many log files.

### What did you expect to see?

The size of each of twelve databases I created, should stay at constant level and has less then 20 GB

### What did you see instead?

After running it for a day, if I look at one of twelve databases, I see 210 sst files, 68 vlog files, db size is 84 GB (and these numbers keep growing). 

If I run badger histogram it shows me this stats:

Histogram of key sizes (in bytes)
Total count: 4499955
Min value: 13
Max value: 108
Mean: 22.92
                   Range     Count
[         8,         16)         2
[        16,         32)   4499939
[        64,        128)        14

Histogram of value sizes (in bytes)
Total count: 4499955
Min value: 82
Max value: 3603
Mean: 2428.16
                   Range     Count
[        64,        128)         1
[       256,        512)     19301
[       512,       1024)       459
[      1024,       2048)       569
[      2048,       4096)   4479625

2428*4479625=10GB

adwinsky:
I should also add that before I patched GC I was suffering the same issues.

jarifibrahim:
[...]

PR #1300, opened 2020-04-14
Set default value of NumVersionsToKeep to 0

jarifibrahim:
Issue https://github.com/dgraph-io/badger/issues/1228 showed that badger doesn't clean SST files even if all the keys are deleted. This is the expected behavior if `NumVersionsToKeep` is set to `1` but for the users this is counter-intuitive. When a key is deleted, it should be removed from the system. This PR fixes it.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1300)

### DIFF
--- a/options.go
+++ b/options.go
@@ -126,7 +126,7 @@ func DefaultOptions(path string) Options {
 		BloomFalsePositive:      0.01,
 		BlockSize:               4 * 1024,
 		SyncWrites:              true,
-		NumVersionsToKeep:       1,
+		NumVersionsToKeep:       0,
 		CompactL0OnClose:        true,
 		KeepL0InMemory:          true,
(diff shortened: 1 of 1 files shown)


ou05020:
@jarifibrahim Doesn't this enable infinite versions? `lastValidVersion` will never be true because `numVersions` will never be less than `1` when compared to `NumVersionsToKeep`.

[https://github.com/dgraph-io/badger/blob/master/levels.go#L579-L580](url)

jarifibrahim:
@ou05020 Yes. You are correct. I was looking at only the `deleted` entries and forgot how this would affect the valid entries. Thank you so much for looking at it :)

Closing this PR.

PR #1302, opened 2020-04-14
Confirm `badgerMove` entry required before rewrite

ou05020:
Value log `badgerMove` entries for deleted keys are always rewritten.

This leads to exponential DB size growth in heavy write/delete use cases with value pointer entries.

To prevent this, a check has been added to confirm the value is still needed for `badgerMove` entries.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1302)

### DIFF
--- a/value.go
+++ b/value.go
@@ -521,6 +521,15 @@ func (vlog *valueLog) rewrite(f *logFile, tr trace.Trace) error {
 		if discardEntry(e, vs) {
 			return nil
 		}
+		if bytes.HasPrefix(e.Key, badgerMove) {
+			// Verify the actual key entry withouth the badgerPrefix has not been deleted.
+			// If this is not done the badgerMove entry will be kept forever moving from
+			// vlog to vlog during rewrites.
+			avs, err := vlog.db.get(e.Key[len(badgerMove):])
+			if err != nil || avs.Version == 0 {
+				return err
+			}
+		}
 
 		// Value is still present in value log.
 		if len(vs.Value) == 0 {
--- a/value.go
+++ b/value.go
@@ -522,7 +522,7 @@ func (vlog *valueLog) rewrite(f *logFile, tr trace.Trace) error {
 			return nil
 		}
 		if bytes.HasPrefix(e.Key, badgerMove) {
-			// Verify the actual key entry withouth the badgerPrefix has not been deleted.
+			// Verify the actual key entry without the badgerPrefix has not been deleted.
 			// If this is not done the badgerMove entry will be kept forever moving from
 			// vlog to vlog during rewrites.
 			avs, err := vlog.db.get(e.Key[len(badgerMove):])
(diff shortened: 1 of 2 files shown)


value.go
```diff
@@ -521,6 +521,15 @@ func (vlog *valueLog) rewrite(f *logFile, tr trace.Trace) error {
 		if discardEntry(e, vs) {
 			return nil
 		}
+		if bytes.HasPrefix(e.Key, badgerMove) {
+			// Verify the actual key entry withouth the badgerPrefix has not been deleted.
```
connyay, line 525:
```suggestion
			// Verify the actual key entry without the badgerPrefix has not been deleted.
```

ou05020:
@manishrjain `movekeys` are only deleted once it is determined a vlog file satisfies the discard ratio and is rewritten.

However, the discard calculation includes movekey entry size in the total number of bytes (not the discard bytes) in the discard ratio calculation even if that movekey is for a key that has been deleted.

If your discard ratio is 50% and half of a file is used by movekey entries for deleted keys, it will never satisfy the discard ratio allowing it to be rewritten and subsequent movekey deletion.

There are only two approaches I can think of to solve the problem.

Always check if underlying key for each movekey has been deleted to determine if an entry should be discarded.

Or, automatically write a delete `movekey` entry every time a delete `key` entry is written. Although that approach will not cleanup existing DB vlog files already in that state and would require a routine similar to the first approach to perform an initial cleanup.

I added db_test.go `BenchmarkDbGrowth` to demonstrate the size growth problem with `badgerMove` keys.

If you update `value.go` `discardEntry` line 1628 to return false, it will cause the DB size to continue to grow for that test.

jarifibrahim:
The `deleteMoveKeysFor` function is supposed to delete the move keys but right now it doesn't  work https://github.com/dgraph-io/badger/blob/62b7a10a949e77b33d43cd7438979833c32cc865/value.go#L685

because the current implementation of iterators does not return duplicates (keys with same version).
So if the existing state was
```
level 1 => MoveKeyFoo{version:1, vlog:x}
```
After rewriting value log file `x` we will have
```
Memtable => MoveKeyFoo{version:1, vlog:y}
level 1 => MoveKeyFoo{version:1, vlog:x}
```
The `deleteMoveKeysFor` function will never see the key with `vlog:x` because the iterator skips duplicates keys.
https://github.com/dgraph-io/badger/blob/62b7a10a949e77b33d43cd7438979833c32cc865/table/merge_iterator.go#L153-L159
The iterator will see only the `vlog;y` key because it is at the top.

Even if the iterator were to not skip this key, we will end up losing data if `deleteMoveKeysFor` function reinserts an old move key with delete marker. Let's assume `deleteMoveKeysFor` function worked. The new state of the system would be
```
Memtable/Level 0 => MoveKeyFoo{version:1, vlog:x, DELETE} 
Memtable/Level 0 => MoveKeyFoo{version:1, vlog:y}
level 1 => MoveKeyFoo{version:1, vlog:x}
```

Compaction would drop all the keys except the one with the delete marker and a `Get(foo)` would return the wrong value.

jarifibrahim:
Thanks for fixing this @ou05020 :tada:

PR #1303, opened 2020-04-14
Simplify single version compaction

ou05020:
When `NumVersionsToKeep` is `1`, deleted and expired entries should always be removed if there is no overlap with lower levels. Otherwise, deleted entries will continue to grow for heavy write/delete use cases with many transient keys.

This PR adds simplified logic when `NumVersionsToKeep` is `1` that always sets the `skipKey` and only writes deleted or expired entries when `hasOverlap` is `true`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1303)

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -573,36 +573,59 @@ func (s *levelsController) compactBuildTables(
 			// Do not discard entries inserted by merge operator. These entries will be
 			// discarded once they're merged
 			if version <= discardTs && vs.Meta&bitMergeEntry == 0 {
-				// Keep track of the number of versions encountered for this key. Only consider the
-				// versions which are below the minReadTs, otherwise, we might end up discarding the
-				// only valid version for a running transaction.
-				numVersions++
-
-				// Keep the current version and discard all the next versions if
-				// - The `discardEarlierVersions` bit is set OR
-				// - We've already processed `NumVersionsToKeep` number of versions
-				// (including the current item being processed)
-				lastValidVersion := vs.Meta&bitDiscardEarlierVersions > 0 ||
-					numVersions == s.kv.opt.NumVersionsToKeep
-
-				if isDeletedOrExpired(vs.Meta, vs.ExpiresAt) || lastValidVersion {
-					// If this version of the key is deleted or expired, skip all the rest of the
-					// versions. Ensure that we're only removing versions below readTs.
+				// When `NumVersionsToKeep == 1` the skipKey should always be set
+				// and there is no need for `lastValidVersion`. The only required check
+				// is to determine if a deleted key should be written due to `hasOverlap`.
+				if s.kv.opt.NumVersionsToKeep == 1 {
+					// skip all the rest of the versions.
 					skipKey = y.SafeCopy(skipKey, it.Key())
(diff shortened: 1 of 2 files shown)


muXxer:
badger_gc
I guess we have the same problem in our software.
Here you can see a chart of the database size. We are deleting a lot of entries, but only the value size decreases at garbage collection.
NumVersionsToKeep is set to 1, because using 0 would never delete the entries/values in the database....

misterpickypants:
@manishrjain @jarifibrahim Any chance you could review this? This was another fix that we needed along with #1302.

jarifibrahim:
@ou05020 Thanks! I didn't realize there was an overlap.

levels.go
```diff
@@ -573,36 +573,59 @@ func (s *levelsController) compactBuildTables(
...
-				lastValidVersion := vs.Meta&bitDiscardEarlierVersions > 0 ||
-					numVersions == s.kv.opt.NumVersionsToKeep
-
-				if isDeletedOrExpired(vs.Meta, vs.ExpiresAt) || lastValidVersion {
-					// If this version of the key is deleted or expired, skip all the rest of the
-					// versions. Ensure that we're only removing versions below readTs.
+				// When `NumVersionsToKeep == 1` the skipKey should always be set
+				// and there is no need for `lastValidVersion`. The only required check
+				// is to determine if a deleted key should be written due to `hasOverlap`.
+				if s.kv.opt.NumVersionsToKeep == 1 {
```
jarifibrahim, line 579: All the changes in `levels.go` file can be replaced with the following
```diff
diff --git a/levels.go b/levels.go
index 764fcd9..0121650 100644
--- a/levels.go
+++ b/levels.go
@@ -578,6 +578,16 @@ func (s *levelsController) compactBuildTables(
 				// only valid version for a running transaction.
 				numVersions++
 
+				if isDeletedOrExpired(vs.Meta, vs.ExpiresAt) && !hasOverlap &&
+					s.kv.opt.NumVersionsToKeep == 1 {
+					skipKey = y.SafeCopy(skipKey, it.Key())
+
+					// If no overlap, we can skip all the versions, by continuing here.
+					numSkips++
+					updateStats(vs)
+					continue // Skip adding this key.
+				}
+
 				// Keep the current version and discard all the next versions if
 				// - The `discardEarlierVersions` bit is set OR
 				// - We've already processed `NumVersionsToKeep` number of versions
```

I tried this and everything seems to be working fine. I'll push some more tests on this PR to ensure we've covered all the cases. We have very few tests for compaction which were added recently.

jarifibrahim:
@ou05020 I'm going to close this PR in favor of PR https://github.com/dgraph-io/badger/pull/1354 since PR https://github.com/dgraph-io/badger/pull/1354 has simpler changes. Thank you so much for fixing this issue, we appreciate your contribution. We have cherry-picked your commit in PR https://github.com/dgraph-io/badger/pull/1354 and so you would be the co-author of that PR.

Thank you so much for your help :tada: 
I would appreciate if you can also review https://github.com/dgraph-io/badger/pull/1354.

PR #1309, opened 2020-04-17
TableInfo: Return valid Left key

jarifibrahim:
`TableInfo` function returns table boundaries. This is used in dgraph to calculate tablet size. As seen in issue https://github.com/dgraph-io/dgraph/issues/5026, badger could return `!badger!head` or `!badger!move` keys as the start of the table. These are valid start keys for badger but they're invalid for dgraph.
This PR changes `TableInfo` function so that we return only valid table start. For a table which contains the following keys
```
!badger!Head!
!badger!Move!...
!badger!Move!...
a
b
za
```
The valid starting key (left key) should be `a` and the end key should be `za`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1309)

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -1040,6 +1040,23 @@ func (s *levelsController) getTableInfo(withKeysCount bool) (result []TableInfo)
 		l.RLock()
 		for _, t := range l.tables {
 			var count uint64
+			var smallest, biggest []byte
+
+			// Create copies of smallest and biggest.
+			smallest = append(smallest, t.Smallest()...)
+			biggest = append(biggest, t.Biggest()...)
+
+			if bytes.HasPrefix(smallest, badgerPrefix) {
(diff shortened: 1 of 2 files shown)


levels.go
```diff
@@ -1040,19 +1040,35 @@ func (s *levelsController) getTableInfo(withKeysCount bool) (result []TableInfo)
 		l.RLock()
 		for _, t := range l.tables {
 			var count uint64
+			var smallest, biggest []byte
+
+			// Create copies of smallest and biggest.
+			smallest = append(smallest, t.Smallest()...)
+			biggest = append(biggest, t.Biggest()...)
```
jarifibrahim, line 1047: There could be a table with only `badgerPrefix` keys. What should we do in that case? Skip the table?

Issue #1139, opened 2019-11-28
RunValueLogGC crashed


cedricfung:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.13.4 linux/amd64
</pre>

### What version of Badger are you using?
v2.0.0

### Does this issue reproduce with the latest master?
Never tried

### What are the hardware specifications of the machine (RAM, OS, Disk)?
Linux 64 SSD

### What did you do?

```golang
	opts := badger.DefaultOptions(dir)
	opts.SyncWrites = sync
	db, err := badger.Open(opts)
	if err != nil {
		return nil, err
	}
	db.RunValueLogGC(0.1)

	go func() {
		ticker := time.NewTicker(1 * time.Minute)
		defer ticker.Stop()
		for range ticker.C {
			lsm, vlog := db.Size()
			if lsm > 1024*1024*8 || vlog > 1024*1024*32 {
				db.RunValueLogGC(0.5)
			}
		}
	}()
```

### What did you expect to see?
Run value log gc should work

### What did you see instead?

```
mixin[28404]: github.com/dgraph-io/badger/v2/y.AssertTrue
mixin[28404]:         /home/one/GOPATH/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.0/y/error.go:55
mixin[28404]: github.com/dgraph-io/badger/v2.(*valueLog).doRunGC.func2
mixin[28404]:         /home/one/GOPATH/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.0/value.go:1591
mixin[28404]: github.com/dgraph-io/badger/v2.(*valueLog).iterate
mixin[28404]:         /home/one/GOPATH/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.0/value.go:480
mixin[28404]: github.com/dgraph-io/badger/v2.(*valueLog).doRunGC
mixin[28404]:         /home/one/GOPATH/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.0/value.go:1557
mixin[28404]: github.com/dgraph-io/badger/v2.(*valueLog).runGC
mixin[28404]:         /home/one/GOPATH/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.0/value.go:1685
mixin[28404]: github.com/dgraph-io/badger/v2.(*DB).RunValueLogGC
[...]

PR #1313, opened 2020-04-19
Ensure `bitValuePointer` flag is cleared for LSM entry values written to LSM

ou05020:
When restoring a backup that was written with a lower `ValueThreshold` to a DB instance with a higher `ValueThreshold`, some entry values originally written to the value log in the backup DB will be written to the LSM along with the key when restored to the new DB.

The `meta` field for those entries is not updated to reflect the correct value storage state. Those entries still have the `bitValuePointer` flag set.

That causes `iterator.prefetch` to fail while looking up the value for those entries.

This fix ensures the `meta` `bitValuePointer` is cleared when entry values are stored in the LSM.

The original error can be reproduced by running the included `TestLSMVPClear` test after reverting the line 687 `db.go` change to clear the `bitValuePointer` flag.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1313)

### DIFF
--- a/backup_test.go
+++ b/backup_test.go
@@ -17,7 +17,6 @@
 package badger
 
 import (
-	"bufio"
 	"bytes"
 	"fmt"
 	"io/ioutil"
@@ -26,7 +25,6 @@ import (
 	"path/filepath"
 	"reflect"
 	"strconv"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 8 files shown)


cedricfung:
Will this fix #1139, without RunValueLogGC our data are increasing so fast.

jarifibrahim:
@cedricfung, the issue with #1139 was that you had some entries with invalid `bitPointerValue` set. This PR fixes the bits on invalid entries.

Did you take a backup and restore it with different `ValueThreshold` value?

jarifibrahim:
Thank you for fixing this @ou05020 :tada:

cedricfung:
@jarifibrahim sorry I don't understand this, what could I do to fix this?

jarifibrahim:
@cedricfung Take a backup of your badger directory and restore it. (Keep the original one around). Do you run badger with default options?

PR #1316, opened 2020-04-27
Always use fsync value from db.opt

jarifibrahim:
The tables were opened in `SyncMode` even if badger was running with `SyncWrites=false`. This PR fixes it.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1316)


jarifibrahim:
There might not be a significant performance improvement because this is a background task. Closing this PR.

PR #1319, opened 2020-04-28
Support fully disabling the bloom filter

damz:
This PR allows the bloom filter to be fully disabled, by setting `BloomFalsePositive` to zero.

It also fixes the description of the `Table.DoesNotHave` which is manifestly wrong.

We have a use case where we use badger as an append only database with values fully stored in the LSM tree. In this use case, we do not benefit from the bloom filter because we do not do any single-key queries, and it makes sense to just disable it. (But note that purging the value log with reasonable performance in this configuration requires #1206.)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1319)


damz:
> This won't work because the index len stores the length of block index (used for binary searching the block) and the bloom filter. This length will never be zero.

Yes, I somehow pushed an older (broken) version of this patch from my stack. Apologies. The new version uses computes a `hasBloomFilter` property in `readIndex()` based on `len(index.Bloomfilter)` .

jarifibrahim:
Hi @damz, the PR was missing tests and I've added some.

damz:
This is still relevant (but needs to be rebased).

jarifibrahim:
@NamanJain8 review, please.

PR #1336, opened 2020-05-21
use cache for storing block offsets

balajijinnah:
Signed-off-by: Tiger <rbalajis25@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1336)

### DIFF
--- a/options.go
+++ b/options.go
@@ -102,6 +102,9 @@ type Options struct {
 	// ------------------------------
 	maxBatchCount int64 // max entries in batch
 	maxBatchSize  int64 // max batch size in bytes
+
+	KeepBlockOffsetsInCache bool // if KeepBlockOffsetsInCache is true and cache is enabled then,
+	// block offsets are kept in cache.
 }
 
 // DefaultOptions sets a list of recommended options for good performance.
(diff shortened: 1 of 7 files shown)


table/table.go
```diff
@@ -45,6 +45,13 @@ import (
 const fileSuffix = ".sst"
 const intSize = int(unsafe.Sizeof(int(0)))
 
+// 1 word = 8 bytes
+// sizeOfOffsetStruct is the size of pb.BlockOffset
+const sizeOfOffsetStruct int64 = 3*8 + // key array take 3 words
```
vvbalaji-dgraph, line 50: How safe/reliable is this computation?

options.go
```diff
@@ -157,19 +163,23 @@ func DefaultOptions(path string) Options {
 		LogRotatesToFlush:             2,
 		EncryptionKey:                 []byte{},
 		EncryptionKeyRotationDuration: 10 * 24 * time.Hour, // Default 10 days.
+		KeepBlocksInCache:             false,
+		KeepBlockIndicesInCache:       false,
```
vvbalaji-dgraph, line 167: Is it the same cache that is referenced in https://github.com/dgraph-io/badger/commit/21735af853dab8942e187dddbeeb41f3ecb04071?

manishrjain:
Emm.. I'm not sure if my comments have been addressed. Two things:

1. The cache size can be configured via options, instead of just asking for a true/false and setting it to 100mb.

> This can be optimized. You can allocate 6 bytes upfront. set the first two bytes to bo, and then do the put uint32.

You missed my point, not sure why. Doing the append at the end would create a new byte slice. You can avoid that by just allocating the 6 bytes, setting the first two to have those two letters and then set the integer in the last 4.

balajijinnah:
@manishrjain 
I've made the capsize of 6 bytes and made len as 4. So, I can add the number in the begining and append the letter at the end. That hardcoded letter will be part of the stack both ways. 

I did this way to match the consistency with the other keys.

PR #1337, opened 2020-05-21
fix: Fix race condition in block.incRef

jarifibrahim:
Fixes https://github.com/dgraph-io/dgraph/issues/5456 .

This PR fixes the crash that could occur when a block was read from the cache.
There was a logical race condition. The following sequence of events could occur which would cause the crash.
1. An iterator makes `t.Block(idx)` call
2. The `t.Block` function finds the block in the cache. The newly found block has `ref=1` which means it was held only by the cache.
3. The `t.Block` function is holding the block and at the same time the block gets evicted from the cache. The existing ref of the block was `1` so the cache eviction would decrement the ref and make it `0`. When the ref becomes `0`, the block is added to the `sync.Pool` and is ready to be reused.
4. While the block got evicted from the cache, the iterator received the block and it incremented the ref from `0` to `1` and starts using this.

Since the block was inserted into the syncPool in the 3rd event, it could be modified by anyone while the iterator is using it.

The fix is to not increment the block ref if it is already zero.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1337)

### DIFF
--- a/badger/cmd/bank.go
+++ b/badger/cmd/bank.go
@@ -197,6 +197,25 @@ func diff(a, b []account) string {
 
 var errFailure = errors.New("test failed due to balance mismatch")
 
+// get function will fetch the value for the key "k" either by using tht
+// txn.Get API or the iterator.Seek API
+func get(txn *badger.Txn, k []byte) (*badger.Item, error) {
+	if rand.Int()%2 == 0 {
+		return txn.Get(k)
+	}
+
+	iopt := badger.DefaultIteratorOptions
+	// PrefectValues is expensive. We don't need it here.
+	iopt.PrefetchValues = false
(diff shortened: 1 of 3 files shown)


badger/cmd/bank.go
```diff
@@ -197,14 +197,33 @@ func diff(a, b []account) string {
 
 var errFailure = errors.New("test failed due to balance mismatch")
 
+// get function will fetch the value for the key "k" either by using tht
+// txn.Get API or the iterator.Seek API
+func get(txn *badger.Txn, k []byte) (*badger.Item, error) {
```
gja, line 202: prefer if you name this to add random

jarifibrahim:
All review comments have been addressed.

PR #1346, opened 2020-05-26
Increase default valueThreshold from 32B to 1KB

jarifibrahim:
Increase value threshold from 32 bytes to 1 KB to allow easier space
reclaimation. There have been multiple reports about value log garbage
collection being unsatisfactory. This commit increases the default value
of threshold from 32 bytes to 1 KB which would allow keys and values
upto 1 KB to be co-allocated in the LSM tree. This would allow GC to
reclaim space easily for values less than 1 KB.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1346)

### DIFF
--- a/db.go
+++ b/db.go
@@ -380,6 +380,7 @@ func Open(opt Options) (db *DB, err error) {
 		vptr.Decode(vs.Value)
 	}
 
+	db.writeCh = make(chan *request, kvWriteChCapacity)
 	replayCloser := y.NewCloser(1)
 	go db.doWrites(replayCloser)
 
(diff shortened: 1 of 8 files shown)


db.go
```diff
@@ -380,6 +380,7 @@ func Open(opt Options) (db *DB, err error) {
 		vptr.Decode(vs.Value)
 	}
 
+	db.writeCh = make(chan *request, kvWriteChCapacity)
```
jarifibrahim, line 383: There was a race condition between this line and the `flushDiscardStats` function. Moving it before starting the `doWrites` go routine seems to have fixed the race.

jarifibrahim:
The build failure will be fixed by https://github.com/dgraph-io/badger/pull/1382

PR #1354, opened 2020-06-02
levels: Expired keys and delete markers are never purged

damz:
The compaction process accidentally keeps one more version of expired keys and delete markers than it should, because of logic error in `*levelsController.compactBuildTables()`. When `NumVersionsToKeep` is 1, it means that expired keys and delete markers are never actually purged from the LSM tables.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1354)

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -593,6 +593,8 @@ func (s *levelsController) compactBuildTables(
 					skipKey = y.SafeCopy(skipKey, it.Key())
 
 					switch {
+					// Add the key to the table only if it has not expired.
+					// We don't want to add the deleted/expired keys.
 					case !isExpired && lastValidVersion:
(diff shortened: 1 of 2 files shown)


damz:
Unfortunately, this is really tricky to test, because I don't see a way to force the engine to purge the memtable. I forced a close/open of the database in the test, which does the trick but it is not extremely pretty.

jarifibrahim:
@damz The delete markers issue should be fixed by https://github.com/dgraph-io/badger/pull/1303

damz:
@jarifibrahim Thank you for pointing out the obvious :)

I cherry-picked the test from the other PR, which looks much better than my attempt. I confirm that it fails before the fix and passes after.

PR #1365, opened 2020-06-11
Avoid panic in fillTables()

ashish-goswami:
Fixes: DGRAPH-1673

We are seeing below panic in sentry:
```
panic: runtime error: makeslice: len out of range

goroutine 21775506 [running]:
github.com/dgraph-io/badger/v2.(*levelsController).fillTables(0xc0001c5180, 0xc1ea6d3b40, 0x0)
	/go/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.1-rc1.0.20200528205344-e7b6e76f96e8/levels.go:829 +0x3bd
github.com/dgraph-io/badger/v2.(*levelsController).doCompact(0xc0001c5180, 0x1, 0x3ff04f5543000000, 0x0, 0x0, 0x0, 0x0, 0x0)
	/go/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.1-rc1.0.20200528205344-e7b6e76f96e8/levels.go:923 +0x91b
github.com/dgraph-io/badger/v2.(*levelsController).runWorker(0xc0001c5180, 0xc13b7156a0)
	/go/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.1-rc1.0.20200528205344-e7b6e76f96e8/levels.go:368 +0x2fd
created by github.com/dgraph-io/badger/v2.(*levelsController).startCompact
	/go/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.1-rc1.0.20200528205344-e7b6e76f96e8/levels.go:343 +0x88
```

Above panic is occuring in `fillTables()`, which is called to pick right tables for compaction from any
level(L) and its next level(L+1). When we pick any table from level L, we pick all table from level
L+1, who are overlapping with level L's table key range. To achieve above task we call
`overlappingTables()`, which returns the range of index (left, right). After this we create a slice
of size (right-left) to copy tables from index `left` to `right` using `make()`.

Ideally this right should be >= left. But in some case we have seeing above panic and this might be
when right < left. We are not able to reproduce this.
This PR just logs the error when `right` < `left` and continues the loop. Once we see this information
in logs we can get more context around cases when right < left.
 

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1365)

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -830,6 +830,18 @@ func (s *levelsController) fillTables(cd *compactDef) bool {
 		cd.top = []*table.Table{t}
 		left, right := cd.nextLevel.overlappingTables(levelHandlerRLocked{}, cd.thisRange)
 
+		// Sometimes below line(make([]*table.Table, right-left)) panics with error
+		// (runtime error: makeslice: len out of range). One of the reason for this can be when
+		// right < left. We don't know how to reproduce it as of now. We are just logging it so
+		// that we can get more context.
+		if right < left {
+			s.kv.opt.Errorf("right: %d is less than left: %d in overlappingTables for current "+
+				"level: %d, next level: %d, key range(%s, %s)", right, left, cd.thisLevel.level,
+				cd.nextLevel.level, cd.thisRange.left, cd.thisRange.right)
+
+			continue
+		}
+
 		cd.bot = make([]*table.Table, right-left)
 		copy(cd.bot, cd.nextLevel.tables[left:right])
 


jarifibrahim:
The fill table panic is reproducible on https://github.com/dgraph-io/badger/pull/1346 . Run the `TestStreamWriter6` test.

ashish-goswami:
> The fill table panic is reproducible on #1346 . Run the TestStreamWriter6 test.

Did more digging into above PR. `TestStreamWriter6` sets keys which are of size `db.opt.MaxTableSize`. PR #1346 changes `db.opt.MaxTableSize` to `5MB` in tests. However max
key size allowed is `math.MaxUint16`(check `builder.go:addHelper()`) which is around `~64KB`. This is leading to overflow in length, and storing wrong length for keys.

Issue #1363, opened 2020-06-11
Crash replays all the value log files


jarifibrahim:
### What version of Badger are you using?
https://github.com/dgraph-io/badger/commit/14386ac9b764cb6655cb0da7c848e244369a2a27

### Does this issue reproduce with the latest master?
Yes

### What Badger options were set? 

```
opt := badger.DefaultOptions(sstDir).    
         WithValueDir(vlogDir).    
         WithTruncate(truncate).    
         WithSyncWrites(false).    
         WithKeepBlocksInCache(true).    
         WithMaxCacheSize(200 << 20).    
         WithKeepBlockIndicesInCache(true).    
         WithCompression(options.ZSTD).    
         WithCompactL0OnClose(force)    
```

### What did you do?
Run benchmark write tool 
```
go run main.go benchmark write -m 400 --dir ~/m2_drive/100m -l 2>&1 | tee 100m-logs.txt
```

### What did you expect to see?

Value log replayed to replay only a few recent files

### What did you see instead?
Badger is replaying file with id 0 offset 0, I have 1164 files
```
ls ~/m2_drive/100m/ | grep .vlog | wc -l
1164
```
```
go run main.go benchmark write -m 400 --dir ~/m2_drive/100m -l 2>&1 | tee 100m-logs.txt
Listening for /debug HTTP requests at port: 8080
badger 2020/06/11 17:28:28 INFO: 335 tables out of 1170 opened in 3.002s
badger 2020/06/11 17:28:31 INFO: 669 tables out of 1170 opened in 6.026s
badger 2020/06/11 17:28:34 INFO: 1004 tables out of 1170 opened in 9.015s
badger 2020/06/11 17:28:36 INFO: All 1170 tables opened in 10.797s
badger 2020/06/11 17:28:36 INFO: Replaying file id: 0 at offset: 0
badger 2020/06/11 17:28:37 INFO: Got compaction priority: {level:2 score:1.1463758505880832 dropPrefix:[]}
badger 2020/06/11 17:28:37 INFO: Running for level: 2
badger 2020/06/11 17:28:38 INFO: Got compaction priority: {level:3 score:1.1370206640660763 dropPrefix:[]}
badger 2020/06/11 17:28:38 INFO: Running for level: 3
badger 2020/06/11 17:28:38 INFO: Replay took: 2.271997389s
badger 2020/06/11 17:28:38 INFO: Replaying file id: 1 at offset: 0
```

jarifibrahim:
[...]

PR #1372, opened 2020-06-16
Update head while replaying

jarifibrahim:
Fixes https://github.com/dgraph-io/badger/issues/1363

The head pointer was not being updated when we perform replays. The head pointer would be updated only when the replay completes. If badger crashed between the point when replay started and replay finished, we would end up replaying all the value log files. This PR fixes this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1372)

### DIFF
--- a/db.go
+++ b/db.go
@@ -1002,7 +1002,8 @@ func (db *DB) handleFlushTask(ft flushTask) error {
 	if ft.mt.Empty() {
 		return nil
 	}
-	// vptr can be zero in-inmemory mode and that's okay because we don't have any value log files and there won't be any replays.
+	// vptr can be zero in-inmemory mode and that's okay because we don't have
+	// any value log files and there won't be any replays.
 	if !db.opt.InMemory {
 		y.AssertTrue(!ft.vptr.IsZero())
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -136,6 +136,11 @@ func (db *DB) replayFunction() func(Entry, valuePointer) error {
 		} else {
 			nv = vp.Encode()
 			meta = meta | bitValuePointer
+			// Update vhead. If the crash happens while replay was in progess
+			// and the head is not updated, we will end up replaying all the
+			// files again.
+			y.AssertTrue(!vp.Less(db.vhead))
+			db.vhead = vp
```
jarifibrahim, line 143: We could use the `db.UpdateHead` function here but using that function requires locking `db.vhead`.
https://github.com/dgraph-io/badger/blob/158d9271e1052082d6377ffa56483650129b726c/db.go#L674-L691

I could move the lock outside of the function but this would cause contention on the lock. The db.Lock would be held while we're iteration over all the pointers. I prefer to keep it as it is.

PR #1375, opened 2020-06-18
Force KeepL0InMemory to be true when InMemory is true

### LINKED ISSUE #1374: In-memory mode creates physical tables (in the current directory)
It looks like the in-memory mode creates physical tables in the current directory. This is because `*DB.handleFlushTask` does not seem to have a special case for the in-memory mode.

### Steps to Reproduce the issue

- Run `go test .` and watch SST files being created and destroyed in the current directory.

itsyashkothari:
InMemory mode used to create L0 sst tables when KeepL0InMemory was false as mentioned in #1374

This commit will keep L0 in memory if either KeepL0InMemory is set or InMemory is set.

On running ` go test .`
Before : sst files were visible for a second 
Now : sst files won't created/visible

Fixes #1374

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1375)

### DIFF
--- a/db.go
+++ b/db.go
@@ -218,8 +218,7 @@ func Open(opt Options) (db *DB, err error) {
 	if opt.Compression == options.ZSTD && !y.CgoEnabled {
 		return nil, y.ErrZstdCgo
 	}
-	// Keep L0 in memory if either it is set or if InMemory is set. We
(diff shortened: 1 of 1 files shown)


itsyashkothari:
> I think only first line is fine as comment.

Done

PR #1381, opened 2020-06-22
Rework DB.DropPrefix

### LINKED ISSUE #1380: DropPrefix breaks levels, resulting in hard crashes and data loss

<pre>
$ go version
go version go1.14.4 linux/amd64
</pre>

### What operating system are you using?

Linux

master

### Steps to reproduce the issue

This is actually pretty tricky to reproduce in a synthetic way.

You need a database with a sufficient amount of data so that on one level you have a layout similar to this:

* Table containing the `badgerMoveKey` prefix (either actual move keys, or something else), and is not full
* [...] Some number of other tables
* Table with the prefix you want to drop

In this case, the `dropPrefix` will generate a same-layer compaction request where `bot` has two non-consecutive tables:

https://github.com/dgraph-io/badger/blob/d37ce36911ae349fbedba342f693184aa434cc30/levels.go#L301-L330

If it happens that `compactTables()` generates tables without the same exact boundaries (for example because there were expired keys / deleted keys) then the result will be a layer with overlapping tables.

The assumption that `bot` has consecutive tables was broken in https://github.com/dgraph-io/badger/pull/1331

This results in data loss, and/or various assertions depending on what compactions are going to happen after that. In our case, we trigger the new assertion added in https://github.com/dgraph-io/badger/pull/1365 and various other assertions.

### How to fix?

`dropPrefix` should generate a separate compaction request for the move key prefix, instead of trying to generate a single compaction request for everything.

damz:
Fixes three issues with the current implementation:

- It can generate compaction requests that break the invariant that bottom tables need to be consecutive (issue #1380). See https://github.com/dgraph-io/badger/issues/1380#issuecomment-649448069
- It performs the same level compactions in increasing order of levels (starting from L0) which leads to old versions of keys for the prefix re-surfacing to active transactions.
- When you have to drop multiple prefixes, the API forces you to drop one prefix at a time and go through the whole expensive table rewriting multiple times.

Fixes #1380

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1381)

### DIFF
--- a/db.go
+++ b/db.go
@@ -974,7 +974,7 @@ func buildL0Table(ft flushTask, bopts table.Options) []byte {
 	defer b.Close()
 	var vp valuePointer
 	for iter.SeekToFirst(); iter.Valid(); iter.Next() {
-		if len(ft.dropPrefix) > 0 && bytes.HasPrefix(iter.Key(), ft.dropPrefix) {
+		if len(ft.dropPrefixes) > 0 && hasAnyPrefixes(iter.Key(), ft.dropPrefixes) {
 			continue
 		}
 		vs := iter.Value()
@@ -987,9 +987,9 @@ func buildL0Table(ft flushTask, bopts table.Options) []byte {
 }
 
 type flushTask struct {
-	mt         *skl.Skiplist
-	vptr       valuePointer
-	dropPrefix []byte
+	mt           *skl.Skiplist
+	vptr         valuePointer
+	dropPrefixes [][]byte
 }
 
 // handleFlushTask must be run serially.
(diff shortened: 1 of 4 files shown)


damz:
> What is the benefit of cleaning up levels in the reverse order?

@jarifibrahim Unless I am missing something, DropPrefix does not prevent new transactions and iterators from being started. If we drop prefixes in the order of the levels, we make old versions of keys or deleted keys reappear for a short while. (For example, if L1 contains a delete key for key `foo`, and an old version of the key is in L2, the compaction on L1 will make the key reappear.)

Of course, this operation is not under normal SSI guarantees, but I think it is nicer to at least not have the prefix go back in time during the DropPrefix process.

jarifibrahim:
Thanks! That makes sense. I've pushed a test and some minor variable renames to this PR. I hope you it's okay with you.

jarifibrahim:
Thanks for another awesome PR @damz  :tada:

Issue #1162, opened 2019-12-16
Use pure Go zstd implementation


narqo:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.13.4 darwin/amd64
</pre>

### What version of Badger are you using?
latest master

### Does this issue reproduce with the latest master?
Yes

As it was mentioned in #1081 #1104 BadgerDB consider switching to a pure Go implementation of zstd ([github.com/klauspost/compress](https://github.com/klauspost/compress/tree/master/zstd)) after it went out of Beta.

Version 1.9.4 is no longer marked as beta, refer to https://github.com/klauspost/compress/tree/v1.9.4/zstd#status

> STABLE - there may always be subtle bugs, a wide variety of content has been tested and the library is actively used by several projects. This library is being continuously fuzz-tested, kindly supplied by fuzzit.dev.

jarifibrahim:
We're working on this @narqo. I'll try to make this change as soon as possible.

jarifibrahim:
Hi @narqo , we've decided to not use the pure go based ZSTD because of performance issues. Please see https://github.com/dgraph-io/badger/pull/1176#issuecomment-573635053 and the benchmarks in https://github.com/dgraph-io/badger/pull/1176#issue-356949512 .


You can build badger with `CGO_ENABLED=0` to use badger without `CGO` in which case `Snappy` would be the default compression algorithm.

johanbrandhorst:
The discussion in https://github.com/dgraph-io/badger/pull/1176 implies to me that this decision should be reconsidered.

mvdan:
I'm also interested in seeing this reopened. I have a package that imports badger/v2, and the indirect dependency on DataDog/zstd is really unfortunate. It takes a long time to build all that C code, and because it's a single package, it means a huge bottleneck in my build time.

jarifibrahim:
I did some benchmarks against both the libraries using the badger benchmark write tool https://github.com/dgraph-io/badger/blob/master/badger/cmd/write_bench.go (compresion is disabled by default, you'll have to enable it) and here's what I found
```
writing 100 million key-values approx 22 gig of data
Datadog/zstd
	Elapsed (wall clock) time (h:mm:ss or m:ss): 6:23.08
Klauspost/compress
	Elapsed (wall clock) time (h:mm:ss or m:ss): 7:50.20
```
## Datadog/zstd logs 
(run master with compression enabled)
```
[...]

PR #1176, opened 2019-12-26
Use pure Go based ZSTD implementation

jarifibrahim:
Fixes https://github.com/dgraph-io/badger/issues/1162

This PR proposes to use https://github.com/klauspost/compress/tree/master/zstd instead of CGO based https://github.com/DataDog/zstd . 

This PR also removes the `CompressionLevel` options since https://github.com/klauspost/compress/tree/master/zstd supports only two levels of ZSTD Compression. The default level is `ZSTD Level 3` and the `fastest` level is `ZSTD level 1`. `ZSTD level 1` will be the default level in badger.

I've experimented will all the suggestions mentioned in https://github.com/klauspost/compress/issues/196#issuecomment-568905095 . Setting `WithSingleSegment` didn't seem to make a lot of speed difference (~ 1MB/s difference)
`WithNoEntropyCompression` seemed to have ~ 3% of speed improvement (but that could also be because of non-deterministic nature of benchmarks)
```
name                                       old time/op      new time/op (NoEntropy set)   delta
Compression/ZSTD_-_Go_-_level1-16           35.7µs ± 1%     36.9µs ± 5%                 +3.41%  (p=0.008 n=5+5)
Decompression/ZSTD_-_Go-16                  16.0µs ± 0%     15.9µs ± 1%                 -0.77%  (p=0.016 n=5+5)

name                                    old speed      new speed (NoEntropy set)      delta
Compression/ZSTD_-_Go_-_level1-16      115MB/s ± 1%   111MB/s ± 5%                -3.24%  (p=0.008 n=5+5)
Decompression/ZSTD_-_Go-16             256MB/s ± 0%   258MB/s ± 1%                 +0.78%  (p=0.016 n=5+5)
```

### Benchmarks
1. Table Data (contains some randomly generated data). 
```
Compression Ratio Datadog ZSTD level 1 3.1993720565149135
Compression Ratio Datadog ZSTD level 3 3.099619771863118

Compression Ratio Go ZSTD 3.2170481452249406
Compression Ratio Go ZSTD level 3 3.1474903474903475
```
```
name                                        time/op
Compression/ZSTD_-_Datadog-level1-16    17.6µs ± 3%
Compression/ZSTD_-_Datadog-level3-16    20.7µs ± 3%

Compression/ZSTD_-_Go_-_level1-16       27.8µs ± 2%
Compression/ZSTD_-_Go_-_Default-16      39.1µs ± 1%

Decompression/ZSTD_-_Datadog-16         7.12µs ± 2%
[...]

PR #1383, opened 2020-06-23
Replace Datadog/zstd with Klauspost/compress

jarifibrahim:
Fixes https://github.com/dgraph-io/badger/issues/1162
PR https://github.com/dgraph-io/badger/pull/1176 was the first attempt at replacing the library but we decided to not switch the library because of performance issues.

https://github.com/klauspost/compress has become much faster since the last time we benchmarked it in https://github.com/dgraph-io/badger/pull/1176. The new benchmarks in https://github.com/dgraph-io/badger/issues/1162#issuecomment-646958797 look very promising and we should get rid of the CGO dependency. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1383)


klauspost:
> The block size of 4KB was converted to a size of 5554 bytes.

Let me double check. That shouldn't be possible.

y/zstd.go
```diff
@@ -19,18 +17,38 @@
...
+	var err error
+	zstdEncOnce.Do(func() {
+		zstdEnc, err = zstd.NewWriter(
+			nil, zstd.WithZeroFrames(true),
+			zstd.WithEncoderLevel(zstd.EncoderLevelFromZstd(level)),
+			zstd.WithEncoderCRC(false))
+
+		AssertTrue(err == nil)
+	})
+	return zstdEnc.EncodeAll(src, dst), nil
```
klauspost, line 53:
```suggestion
	return zstdEnc.EncodeAll(src, dst[:0]), nil
```

This is likely the problem. Output is appended to make it easier to use in serializers, etc.

y/zstd.go
```diff
@@ -19,18 +17,38 @@
...
 
 // ZSTDDecompress decompresses a block using ZSTD algorithm.
 func ZSTDDecompress(dst, src []byte) ([]byte, error) {
-	return zstd.Decompress(dst, src)
+	var err error
+	zstdDecOnce.Do(func() {
+		zstdDec, err = zstd.NewReader(nil)
+		AssertTrue(err == nil)
+	})
+	return zstdDec.DecodeAll(src, dst)
```
klauspost, line 39:
```suggestion
	return zstdDec.DecodeAll(src, dst[:0])
```
If you don't want it appended.

ivanjaros:
Any progress with this?

klauspost:
@jarifibrahim Ping. You might want to upgrade to `v1.11.3`, there are some minor tweaks and fixes. I don't see them invalidating the tests, but might as well get them in.

emolitor:
Has this PR been abandoned? Slightly confused by the state of the ticket as reading through the comments it appears all the issues which were raised have been addressed. Are there other issues which are not documented or is this ticket just waiting for time by the author to test and commit?

NamanJain8:
@jarifibrahim closing this because the issue is addressed by #1706

jarifibrahim:
We have merged https://github.com/dgraph-io/badger/pull/1706 which adds the pure go-based ZSTD to badger. This will be released as part of the badger release in July.

@mholt @emolitor @ivanjaros @klauspost Thank you all for your feedback and helping improve badger :heart: . Apologies for the delay. This PR was lost in all the other work we're doing on badger/dgraph.

jarifibrahim:
@francislavoie Okay. I think we can set pure go as the default one. I think @klauspost already mentioned that the klauspost/compress should be able to decompress data written by datadog/zstd. So we shouldn't have a problem.

klauspost:
Yes and vice versa if that should become relevant.

PR #1393, opened 2020-06-30
Fix assertions in table/builder.go

jarifibrahim:
In builder.go we have some assertions to ensure we don't break the table
in any way. Some of these asserts were incorrect. This PR fixes it.

Assertions such as `Assert(uint32(x) <= math.MaxUint32)` will never fail
since we're casting the number into uint32.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1393)

### DIFF
--- a/table/builder.go
+++ b/table/builder.go
@@ -213,7 +213,6 @@ func (b *Builder) addHelper(key []byte, v y.ValueStruct, vpLen uint64) {
 	}
 
 	// store current entry's offset
-	y.AssertTrue(b.sz < math.MaxUint32)
 	b.entryOffsets = append(b.entryOffsets, b.sz-b.baseOffset)
 
 	// Layout: header, diffKey, value.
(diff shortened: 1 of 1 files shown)


table/builder.go
```diff
@@ -332,11 +334,11 @@ func (b *Builder) shouldFinishBlock(key []byte, value y.ValueStruct) bool {
 
 // Add adds a key-value pair to the block.
 func (b *Builder) Add(key []byte, value y.ValueStruct, valueLen uint32) {
+	y.AssertTrue(b.sz < math.MaxUint32)
```
damz, line 337: This cannot be false (sz is a uint32).

table/builder.go
```diff
@@ -313,7 +315,7 @@ func (b *Builder) shouldFinishBlock(key []byte, value y.ValueStruct) bool {
 	}
 
 	// Integer overflow check for statements below.
-	y.AssertTrue((uint32(len(b.entryOffsets))+1)*4+4+8+4 < math.MaxUint32)
+	y.AssertTrue((len(b.entryOffsets)+1)*4+4+8+4 < math.MaxUint32)
```
damz, line 318: This is not going to work on 32 bit builds.

You probably want:

```
y.AssertTrue((uint64(len(b.entryOffsets))+1)*4+4+8+4 < math.MaxUint32)
```

jarifibrahim:
Closing this PR. We don't need it anymore.

PR #1401, opened 2020-07-08
Avoid panic on multiple closer.Signal calls

jarifibrahim:
The closer.Signal function closes a channel and multiple calls to the
closing the same channel would cause a "channel already closed". This PR
avoids the panic by using a sync.Once while closing the channel.

Multiple calls to closer.Signal would be no-op with this commit.

Fixes DGRAPH-1581

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1401)

### DIFF
--- a/y/y.go
+++ b/y/y.go
@@ -191,8 +191,9 @@ func FixedDuration(d time.Duration) string {
 // to tell the goroutine to shut down, and a WaitGroup with which to wait for it to finish shutting
 // down.
 type Closer struct {
-	closed  chan struct{}
-	waiting sync.WaitGroup
+	closed    chan struct{}
+	waiting   sync.WaitGroup
+	closeOnce sync.Once
 }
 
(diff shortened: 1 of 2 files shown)


y/y.go
```diff
@@ -209,7 +210,9 @@ func (lc *Closer) AddRunning(delta int) {
 
 // Signal signals the HasBeenClosed signal.
 func (lc *Closer) Signal() {
-	close(lc.closed)
+	lc.closeOnce.Do(func() {
+		close(lc.closed)
```
jarifibrahim, line 214: Write a unit test for this.

jarifibrahim: Add a todo about changing this API when we introduce breaking changes.

jarifibrahim: Done.

jarifibrahim:
The travis build for 32 bit CPU is failing and it is an existing issue, not related to this PR.

PR #1412, opened 2020-07-10
Add a limit to the size of the batches sent over a stream.

martinmr:
Currently, there's no limit to how many KVs can be included in a single
batch. This leads to issues when the size is over a hard limit. For
example, a batch of size > 2GB will cause issues if it has to be sent
over gRPC.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1412)

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -32,6 +32,9 @@ import (
 
 const pageSize = 4 << 20 // 4MB
 
+// maxStreamSize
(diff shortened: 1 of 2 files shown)


simon28082:
maxStreamSize can changed? I want use small size, but current maxStreamSize=100M is fixed

PR #1275, opened 2020-03-24
Clean up transaction oracle as we go, take two

muXxer:
Hi badger team,

I resolved all the comments @jarifibrahim made in #1198 
Would be cool if that could finally be merged, since RAM usage is much less.

Thanks to @damz !

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1275)

PR #1413, opened 2020-07-10
txn: Do not keep one too many conflict information

damz:
This fixes an issue introduced in #1275.

If the read watermark is, say `10`, it means that all the read transactions of `readTs <= 10` have completed. All the remaining in-flight transactions have `readTs >= 11`, which cannot possibly conflict with transactions of `commitTs <= 11`. We were keeping conflict information of one too many transactions.

Unfortunately, trying to fix that revealed a causality issue in `y.WaterMark`, something like that would likely fail because the processing is asynchronous:

```
w.Begin(1)
w.Done(1)
AssertTrue(w.DoneUntil() == 1)
```

So the PR rewrites `y.WaterMark` to make in synchronous (and massively simplifying it). My (limited) benchmarks seem to show that not trying to be smart here actual leads to a measurable performance _improvement_, even under high concurrency.


damz:
Note that this reveled that the `y.WaterMark` API is used incorrectly all over the place:

1. Multiple places use `Done()` instead of `SetDoneUntil()` to set the current water mark;
2. Multiple places are trying to use `0` as a valid water mark, but the current implementation behaves incorrectly in that case (`DoneUntil()` on a zero `y.WaterMark` would incorrectly return `0`, but that index is not done)

The only way I can think of to fix (2) would be to change the API to return `int64` instead of `uint64` and make the implementation return `-1` for `DoneUntil()` until `0` is done.

backup.go
```diff
@@ -259,6 +259,6 @@ func (db *DB) Load(r io.Reader, maxPendingWrites int) error {
 	if err := ldr.Finish(); err != nil {
 		return err
 	}
-	db.orc.txnMark.Done(db.orc.nextTxnTs - 1)
+	db.orc.txnMark.SetDoneUntil(db.orc.nextTxnTs - 1)
```
jarifibrahim, line 262: This should be `txnMark.Done(..)`. What if there other transactions running at the `ts<db.orc.nextTxnTs`? We would mark transactions as done which haven't completed yet.

damz: All the cases where I have changed `Done()` into `SetDoneUntil()` are cases where `Done()` was called without a corresponding `Begin()`, as far as I know.

jarifibrahim: Is it possible to not make this change and still use your implementation? I understand that the existing behavior wasn't perfect but if possible I'd like to avoid this change since it is hard to test this.

PR #1422, opened 2020-07-14
levels: Compaction incorrectly drops delete markers

damz:
This took me a ridiculous amount of time to track down.

fd8989493 ("Compaction: Expired keys and delete markers are never purged") revealed a bug in the compaction logic that leads to delete markers being incorrectly dropped during compaction (i.e. deleted keys are reappearing).

During compaction, `*levelsController.compactBuildTables` decides to drop a delete key if there is no overlap with levels lower than the bottom layer of the compaction definition.

It does that by checking only the `top` table, thinking that proving that the `top` table doesn't overlap is sufficient to prove that the `bottom` table doesn't. Unfortunately, this is not the case. Not in general, and even less in the case of `DropPrefix()` where we run a same-level compaction and `top` is empty.

The faulty condition was introduced way back when in e597fb721 ("Discard key versions during compaction").

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1422)

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -484,11 +484,8 @@ func (s *levelsController) compactBuildTables(
 	botTables := cd.bot
 
 	// Check overlap of the top level with the levels which are not being
-	// compacted in this compaction. We don't need to check overlap of the bottom
-	// tables with other levels because if the top tables overlap with any of the lower
-	// levels, it implies bottom level also overlaps because top and bottom tables
-	// overlap with each other.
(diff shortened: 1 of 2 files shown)


levels.go
```diff
@@ -484,11 +484,8 @@ func (s *levelsController) compactBuildTables(
 	botTables := cd.bot
 
 	// Check overlap of the top level with the levels which are not being
```
NamanJain8, line 486: @jarifibrahim the comment needs to be fixed?

PR #1425, opened 2020-07-14
Add support for inserting entries with TTL in write bench tool

ahsanbarkati:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1425)

### DIFF
--- a/badger/cmd/write_bench.go
+++ b/badger/cmd/write_bench.go
@@ -65,6 +65,7 @@ var (
 	loadBloomsOnOpen    bool
 	detectConflicts     bool
 	compression         bool
+	withTTL             bool
 )
 
 const (
@@ -101,6 +102,8 @@ func init() {
 		"If true, it badger will detect the conflicts")
 	writeBenchCmd.Flags().BoolVar(&compression, "compression", false,
 		"If true, badger will use ZSTD mode")
+	writeBenchCmd.Flags().BoolVar(&withTTL, "entry-ttl", false,
+		"If true, badger will insert entries with TTL")
 }
 
 func writeRandom(db *badger.DB, num uint64) error {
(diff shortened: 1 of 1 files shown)


badger/cmd/write_bench.go
```diff
@@ -112,7 +115,14 @@ func writeRandom(db *badger.DB, num uint64) error {
 	for i := uint64(1); i <= num; i++ {
 		key := make([]byte, keySz)
 		y.Check2(rand.Read(key))
-		if err := batch.Set(key, value); err != nil {
+
+		e := badger.NewEntry(key, value)
+
+		if withTTL {
+			e = badger.NewEntry(key, value).WithTTL(1 * time.Second)
```
NamanJain8, line 122: You can have e.WithTTL(1*time.Second) instead of creating NewEntry(). The rest looks good.

ahsanbarkati: Thanks. Done

badger/cmd/write_bench.go
```diff
@@ -101,6 +102,8 @@ func init() {
 		"If true, it badger will detect the conflicts")
 	writeBenchCmd.Flags().BoolVar(&compression, "compression", false,
 		"If true, badger will use ZSTD mode")
+	writeBenchCmd.Flags().BoolVar(&withTTL, "entry-ttl", false,
```
jarifibrahim, line 105: change this to accept an integer which will be seconds. Set default to zero.

badger/cmd/write_bench.go
```diff
@@ -112,7 +115,12 @@ func writeRandom(db *badger.DB, num uint64) error {
 	for i := uint64(1); i <= num; i++ {
 		key := make([]byte, keySz)
 		y.Check2(rand.Read(key))
-		if err := batch.Set(key, value); err != nil {
+
+		e := badger.NewEntry(key, value)
+		if withTTL {
```
jarifibrahim, line 120: If the value of the flag is zero, do not set TTL, otherwise use it.

PR #1426, opened 2020-07-14
add GC flags for write benchmark

NamanJain8:
This PR adds GC flags to badger write benchmark.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1426)

### DIFF
--- a/badger/cmd/write_bench.go
+++ b/badger/cmd/write_bench.go
@@ -240,9 +240,9 @@ func writeBench(cmd *cobra.Command, args []string) error {
 
 	startTime = time.Now()
 	num := uint64(numKeys * mil)
-	c := y.NewCloser(1)
+	c := y.NewCloser(2)
 	go reportStats(c)
-	go runGC(db)
+	go runGC(c, db)
 
 	if sorted {
 		err = writeSorted(db, num)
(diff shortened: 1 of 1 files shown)


badger/cmd/write_bench.go
```diff
@@ -101,6 +105,8 @@ func init() {
 		"If true, it badger will detect the conflicts")
 	writeBenchCmd.Flags().BoolVar(&compression, "compression", false,
 		"If true, badger will use ZSTD mode")
+	writeBenchCmd.Flags().StringVarP(&gcPeriod, "gc-every", "g", "5m", "GC Period.")
+	writeBenchCmd.Flags().Float64VarP(&gcDiscardRatio, "gc-ratio", "r", 0.1, "GC discard ratio.")
```
ahsanbarkati, line 109: The recommended discard ratio is 0.5. Shouldn't we use that value?

badger/cmd/write_bench.go
```diff
@@ -263,8 +270,26 @@ func reportStats(c *y.Closer) {
...
+	period, err := time.ParseDuration(gcPeriod)
+	y.Check(err)
+	t := time.NewTicker(period)
+	defer t.Stop()
+	for {
+		select {
+		case <-c.HasBeenClosed():
+			return
+		case <-t.C:
+			if err := db.RunValueLogGC(gcDiscardRatio); err == nil {
```
ahsanbarkati, line 290: Can we handle the error here? The error can be because of many reasons like invalid discardRatio which can be good to show for better debugging in future.

NamanJain8: Logged the error in case of failure. Else we are storing the success count. Thanks .

PR #1428, opened 2020-07-14
Add support collecting stats (sst count, vlog count, file sizes, move key count, valid keys etc)

ahsanbarkati:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1428)

### DIFF
--- a/badger/cmd/write_bench.go
+++ b/badger/cmd/write_bench.go
@@ -144,7 +144,7 @@ func writeRandom(db *badger.DB, num uint64) error {
 		y.Check2(rand.Read(key))
 		e := badger.NewEntry(key, value)
 
-		if ttlPeriod != 0 && errParse == nil {
+		if ttlPeriod != 0 {
 			e.WithTTL(ttlPeriod)
 		}
 		err := batch.SetEntry(e)
--- a/badger/cmd/write_bench.go
+++ b/badger/cmd/write_bench.go
@@ -136,11 +136,14 @@ func writeRandom(db *badger.DB, num uint64) error {
 	es := uint64(keySz + valSz) // entry size is keySz + valSz
 	batch := db.NewWriteBatch()
 
+	ttlPeriod, errParse := time.ParseDuration(ttlDuration)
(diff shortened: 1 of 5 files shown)


badger/cmd/write_bench.go
```diff
@@ -247,16 +263,66 @@ func writeBench(cmd *cobra.Command, args []string) error {
...
+					}
+					if filepath.Ext(path) == ".sst" {
+						sstCount++
+					}
+					return nil
+				})
+				if err != nil {
+					log.Printf("Error while fetching directory. %v.", err)
+				} else {
+					for _, file := range files {
```
NamanJain8, line 319:
```suggestion
                                        fmt.Printf("[Content] Number of files:\n", len(files))
					for _, file := range files {
```
Something like this?
Keeping a count would be good for later processing (if automated).

badger/cmd/write_bench.go
```diff
@@ -247,16 +263,66 @@ func writeBench(cmd *cobra.Command, args []string) error {
...
+				if i.IsDeletedOrExpired() {
+					invalidKeyCount++
+				} else {
+					validKeyCount++
+				}
+			}
+
+			// fetch directory contents
+			if showDir {
+				fmt.Printf("The directory contents are:\n")
```
NamanJain8, line 303:
```suggestion
```
The contents will be empty in case of an error. Shouldn't we remove it?

badger/cmd/write_bench.go
```diff
@@ -247,16 +263,66 @@ func writeBench(cmd *cobra.Command, args []string) error {
...
+				err := filepath.Walk(sstDir, func(path string, info os.FileInfo, err error) error {
+					fileSize := strconv.FormatFloat(float64(info.Size()) / (1024.0 * 1024.0),
```
NamanJain8, line 305: Doesn't matter much but can use [humanize Bytes](https://godoc.org/github.com/dustin/go-humanize#Bytes)?

PR #1434, opened 2020-07-16
Remove vlog file if bootstrap, syncDir or mmap fails

jarifibrahim:
The `createVlogFile` function creates a new vlog file. The function apart
from creating a new file, bootstraps it, syncs it to the disk, and mmaps it.

If any of the three operations fail, we will end up with a file on the
disk but the vlog.maxFid will not be updated.

On the next attempt to create the vlog files, we will get a "File
already exists" error. This commit fixes this issue by removing the file
if `createVlogFile` encounters any error.

Fixes DGRAPH-1930

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1434)

### DIFF
--- a/value.go
+++ b/value.go
@@ -1000,14 +1000,26 @@ func (vlog *valueLog) createVlogFile(fid uint32) (*logFile, error) {
 		return nil, errFile(err, lf.path, "Create value log file")
 	}
 
+	removeFile := func() {
+		// Remove the file so that we don't get an error when createVlogFile is
+		// called for the same fid, again. This could happen if there is an
+		// transient error because of which we couldn't create a new file
+		// and the second attempt to create the file succeeds.
+		os.Remove(lf.fd.Name())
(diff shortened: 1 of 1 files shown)


value.go
```diff
@@ -1000,14 +1000,26 @@ func (vlog *valueLog) createVlogFile(fid uint32) (*logFile, error) {
 		return nil, errFile(err, lf.path, "Create value log file")
 	}
 
+	removeFile := func() {
```
jarifibrahim, line 1003: Add test for this.

jarifibrahim: I tried to add a test by setting ulimit but this will be difficult for two reasons
1. `ulimit` is set per process. If I set the `ulimit` for a test, it might affect other tests too.
2. It's hard to set the correct `ulimit` and the test wouldn't fail even if the limit is off by 1.

PR #1438, opened 2020-07-17
Add drop-prefix option in write benchmark

ahsanbarkati:
This PR adds the drop-prefix option in write benchmark. The current size of prefix is taken as `1 + 0.1*keySize`, I would request your comments on this.

### DIFF
--- a/badger/cmd/write_bench.go
+++ b/badger/cmd/write_bench.go
@@ -79,10 +79,11 @@ var (
 	vlogCount        uint32
 	files            []string
 
-	dropAllPeriod  string
-	gcPeriod       string
-	gcDiscardRatio float64
-	gcSuccess      uint64
+	dropAllPeriod    string
+	dropPrefixPeriod string
+	gcPeriod         string
+	gcDiscardRatio   float64
+	gcSuccess        uint64
 )
 
 const (
@@ -123,6 +124,8 @@ func init() {
 		"If true, the report will include the directory contents")
 	writeBenchCmd.Flags().StringVar(&dropAllPeriod, "dropall", "0s",
 		"Period of dropping all. If 0, doesn't drops all.")
+	writeBenchCmd.Flags().StringVar(&dropPrefixPeriod, "drop-prefix", "0s",
+		"Period of dropping by random prefixes. If 0, doesn't drops by prefix.")
 	writeBenchCmd.Flags().StringVar(&ttlDuration, "entry-ttl", "0s",
 		"TTL duration in seconds for the entries, 0 means without TTL")
 	writeBenchCmd.Flags().StringVarP(&gcPeriod, "gc-every", "g", "5m", "GC Period.")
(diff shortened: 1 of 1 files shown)


badger/cmd/write_bench.go
```diff
@@ -415,3 +419,38 @@ func dropAll(c *y.Closer, db *badger.DB) {
...
 }
+
+func dropPrefix(c *y.Closer, db *badger.DB) {
+	defer c.Done()
+
+	dropPeriod, err := time.ParseDuration(dropPrefixPeriod)
+	y.Check(err)
+	if dropPeriod == 0 {
+		return
+		log.Println("Exit from drop prefix")
```
jarifibrahim, line 430: This would never be printed. Move it a line above or remove it.

badger/cmd/write_bench.go
```diff
@@ -415,3 +419,38 @@ func dropAll(c *y.Closer, db *badger.DB) {
...
+
+	t := time.NewTicker(dropPeriod)
+	defer t.Stop()
+	for {
+		select {
+		case <-c.HasBeenClosed():
+			return
+		case <-t.C:
+			fmt.Println("[DropPrefix] Started")
+			err := db.DropAll()
```
jarifibrahim, line 441: You're dropping all the data. DropAll followed by drop prefix wouldn't do anything.

badger/cmd/write_bench.go
```diff
@@ -415,3 +419,38 @@ func dropAll(c *y.Closer, db *badger.DB) {
...
+			for err == badger.ErrBlockedWrites {
+				prefix := make([]byte, 1+int(float64(keySz)*0.1))
+				rand.Read(prefix)
+				err = db.DropPrefix(prefix)
```
jarifibrahim, line 445: You should check this error.

badger/cmd/write_bench.go
```diff
@@ -415,3 +419,38 @@ func dropAll(c *y.Closer, db *badger.DB) {
...
+	t := time.NewTicker(dropPeriod)
+	defer t.Stop()
+	for {
+		select {
+		case <-c.HasBeenClosed():
+			return
+		case <-t.C:
+			fmt.Println("[DropPrefix] Started")
+			err := db.DropAll()
+			for err == badger.ErrBlockedWrites {
```
jarifibrahim, line 442: Why would drop prefix see blocked writes? This could only happen if `dropPeriod < time taken to drop the prefix`. I don't think we should do this. Do a simple error check here and stop the program if `err != nil`

badger/cmd/write_bench.go
ahsanbarkati: fixed this

badger/cmd/write_bench.go
ahsanbarkati: Should we just return, or panic?

PR #1396, opened 2020-07-01
Throttle.Do does not panic after Finish

ekalinin:
Fixes #1394

I'm not sure about an error at `Do`. Maybe it will be ok to just return `nil`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1396)

PR #1442, opened 2020-07-22
throttle Do() after Finish() does not panic now

### LINKED ISSUE #1394: WriteBatch.Flush panics if called twice
I understand that flushing twice doesn't make sense, since a batch can only be flushed once, but the second call should error with a helpful message instead of panicking.

```
$ go version
go version devel +152ffca82f Mon Jun 29 08:06:32 2020 +0000 linux/amd64
$ go list -m github.com/dgraph-io/badger/v2
github.com/dgraph-io/badger/v2 v2.0.3
$ cat main.go
package main

import (
	"io/ioutil"
	"os"

	"github.com/dgraph-io/badger/v2"
)

func main() {
	dir, err := ioutil.TempDir("", "badger-test")
	if err != nil {
		panic(err)
	}
	defer os.RemoveAll(dir)
	db, err := badger.Open(badger.DefaultOptions(dir))
	if err != nil {
		panic(err)
	}
	defer db.Close()

	wb := db.NewWriteBatch()
	wb.Flush()
	wb.Flush()
}
$ go run main.go
badger 2020/06/30 18:36:15 INFO: All 0 tables opened in 0s
badger 2020/06/30 18:36:15 INFO: Got compaction priority: {level:0 score:1.73 dropPrefix:[]}
panic: send on closed channel

goroutine 1 [running]:
github.com/dgraph-io/badger/v2/y.(*Throttle).Do(0xc0000200c0, 0x6e7260, 0xc0000200c0)
	/home/mvdan/go/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.3/y/y.go:260 +0xc9
github.com/dgraph-io/badger/v2.(*WriteBatch).commit(0xc000020180, 0x0, 0x0)
	/home/mvdan/go/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.3/batch.go:141 +0x45
github.com/dgraph-io/badger/v2.(*WriteBatch).Flush(0xc000020180, 0x0, 0x0)
	/home/mvdan/go/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.3/batch.go:155 +0x45
main.main()
	/home/mvdan/src/test/main.go:24 +0x1fa
exit status 2
```

NamanJain8:
1. Fixes #1394 (https://discuss.dgraph.io/t/writebatch-flush-panics-if-called-twice/8743 on discuss)
2. Jira: Closes Badger-1394

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1442)

### DIFF
--- a/y/y.go
+++ b/y/y.go
@@ -41,6 +41,10 @@ var (
 	// compression algorithm is being used for compression. ZSTD cannot work
 	// without CGO.
 	ErrZstdCgo = errors.New("zstd compression requires building badger with cgo enabled")
+
+	// ErrDoAfterFinish indicates the invocation of throttle.Do() after
+	// throttle.Finish()
+	ErrDoAfterFinish = errors.New("Do has no point after finish")
 )
 
 const (
@@ -253,6 +257,8 @@ type Throttle struct {
 	ch        chan struct{}
 	errCh     chan error
 	finishErr error
+	finished  uint32
+	finLock   sync.Mutex
 }
 
 // NewThrottle creates a new throttle with a max number of workers.
@@ -267,6 +273,11 @@ func NewThrottle(max int) *Throttle {
 // are already maximum number of workers working. If it detects an error from
 // previously Done workers, it would return it.
 func (t *Throttle) Do() error {
+	t.finLock.Lock()
+	defer t.finLock.Unlock()
+	if t.finished == 1 {
+		return ErrDoAfterFinish
+	}
 	for {
 		select {
 		case t.ch <- struct{}{}:
@@ -299,6 +310,10 @@ func (t *Throttle) Done(err error) {
 // From next calls, it will return same error as found on first call.
 func (t *Throttle) Finish() error {
 	t.once.Do(func() {
+		t.finLock.Lock()
+		t.finished = 1
+		t.finLock.Unlock()
+
 		t.wg.Wait()
 		close(t.ch)
 		close(t.errCh)
--- a/y/y_test.go
+++ b/y/y_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


y/y.go
```diff
@@ -267,6 +273,11 @@ func NewThrottle(max int) *Throttle {
 // are already maximum number of workers working. If it detects an error from
 // previously Done workers, it would return it.
 func (t *Throttle) Do() error {
+	t.finLock.Lock()
+	defer t.finLock.Unlock()
+	if t.finished == 1 {
```
jarifibrahim, line 278: Why do we need a lock here? Using a `atomic.LoadUint32` should be enough

y/y.go
```diff
@@ -299,6 +310,10 @@ func (t *Throttle) Done(err error) {
 // From next calls, it will return same error as found on first call.
 func (t *Throttle) Finish() error {
 	t.once.Do(func() {
+		t.finLock.Lock()
+		t.finished = 1
```
jarifibrahim, line 314: atomic.SetUint32(...)

NamanJain8: Yes, SetUnit32() should be fine here.

y/y.go
NamanJain8: LoadUint32 would just fetch the current value. Consider a scenario where T1 loads the value (false) atomically, then T2 closes the channel, then T1 enters into `for` loop. In that case, we would be sending to the closed channel.

jarifibrahim: Yeah, agreed.

y/y.go
```diff
@@ -41,6 +41,10 @@ var (
 	// compression algorithm is being used for compression. ZSTD cannot work
 	// without CGO.
 	ErrZstdCgo = errors.New("zstd compression requires building badger with cgo enabled")
+
+	// ErrDoAfterFinish indicates the invocation of throttle.Do() after
+	// throttle.Finish()
+	ErrDoAfterFinish = errors.New("Do has no point after finish")
```
jarifibrahim, line 47: I would rephrase this to `Do should not be called after Finish`

jarifibrahim:
Closing this PR in favor of https://github.com/dgraph-io/badger/pull/1396

PR #1445, opened 2020-07-24
Support WAL mode

jarifibrahim:
Fixes DGRAPH-2177

_[This PR contains 300 lines of tests. The actual code is around 250 lines]_

This PR adds support for running badger in `WAL` mode. When `WALMode` option is set, badger will remove the old wal files as soon as new log files are created. This allows for faster disk space reclamation.

The maximum size of value that can be inserted into LSM tree in badger is capped at 1 MB (I don't know why we limit it to 1 MB).
So a user cannot insert a value larger than `1 MB` in the `WALMode`.

Users can switch between the modes without any loss of data. However, they cannot run `Value Log GC` in WAL Mode. This can be supported but this PR doesn't support that. It can be done in a separate PR.

## Problem with WAL Mode and Value Log GC
It could happen that someone inserted a 10 MB value in the normal mode and then opens badger in `WAL mode`. If we now try to run `ValueLog GC`, badger will error out half way while moving the old value since we support only 1 MB (or maxValueThreshold) length of values in WAL mode.
One way to deal with this issue would be to read the entire value log file and GC it only if all values are less than the value threshold. If there is even a single file with threshold greater than the `WAL mode threshold`, we will not GC that `vlog` file.

The approach mentioned above is not implemented in this PR. It will be done in a separate PR. This PR disables value log GC in `WAL mode`.

Todo
- [x] Ensure we run in LSM only mode when `WALMode` is set. (we're setting the max value threshold)
- [x] Add tests [we don't want to accidentally delete value vlog files]
- [x] Disable Value Log GC in WALMode.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1445)


wal_test.go
```diff
@@ -0,0 +1,292 @@
...
+
+func TestWALModeGC(t *testing.T) {
+	opt := DefaultOptions("")
+	opt.WALMode = true
+	runBadgerTest(t, &opt, func(t *testing.T, db *DB) {
+		require.Error(t, db.RunValueLogGC(0.1))
+	})
+}
+
+func TestWALBigValue(t *testing.T) {
```
jarifibrahim, line 254: Add a test with encryption -> switch between modes (wal and vlog)

jarifibrahim:
We don't need this PR anymore. Closing it.

PR #1450, opened 2020-07-29
Add public API to force compactions.

martinmr:
Dgraph is seeing memory issues due to a lot of deleted keys that are not
being cleaned up. Forcing compactions would help with these issues.

This PR makes such an API available. It's based on the existing code for
Flatten except it only performs one compaction at a time instead of
looping until the whole database is in a single level like Flatten does.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1450)

### DIFF
--- a/db.go
+++ b/db.go
@@ -1430,16 +1430,16 @@ func (db *DB) startMemoryFlush() {
 // stopped. Ideally, no writes are going on during Flatten. Otherwise, it would create competition
 // between flattening the tree and new tables being created at level zero.
 func (db *DB) Flatten(workers int) error {
-	return db.flattenInternal(workers, true)
+	return db.flattenInternal(workers, false)
 }
 
 // CompactLSM works like Flatten except it only performs one pass of compactions instead of
 // performing compactions until the whole database is in a single level like Flatten does.
 func (db *DB) CompactLSM(workers int) error {
-	return db.flattenInternal(workers, false)
+	return db.flattenInternal(workers, true)
 }
 
-func (db *DB) flattenInternal(workers int, loop bool) error {
+func (db *DB) flattenInternal(workers int, singlePass bool) error {
 	db.stopCompactions()
 	defer db.startCompactions()
 
@@ -1486,6 +1486,20 @@ func (db *DB) flattenInternal(workers int, loop bool) error {
 				levels = append(levels, i)
 			}
 		}
+
+		// Single pass means we won't try to flatten the entire LSM Tree. We
+		// will try to push the data to the lower levels, once per level.
+		if singlePass {
+			for _, l := range levels {
+				// Create an artificial compaction priority to ensure that we compact the level.
+				cp := compactionPriority{level: l, score: 1.74}
+				if err := compactAway(cp); err != nil {
+					return err
+				}
+			}
+			return nil
+		}
+
 		if len(levels) <= 1 {
 			prios := db.lc.pickCompactLevels()
 			if len(prios) == 0 || prios[0].score <= 1.0 {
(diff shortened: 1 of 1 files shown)


db.go
```diff
@@ -1430,6 +1430,16 @@ func (db *DB) startMemoryFlush() {
 // stopped. Ideally, no writes are going on during Flatten. Otherwise, it would create competition
 // between flattening the tree and new tables being created at level zero.
 func (db *DB) Flatten(workers int) error {
+	return db.flattenInternal(workers, true)
+}
+
+// CompactLSM works like Flatten except it only performs one pass of compactions instead of
+// performing compactions until the whole database is in a single level like Flatten does.
+func (db *DB) CompactLSM(workers int) error {
```
jarifibrahim, line 1438: This call could take a very long time if the LSM tree has a lot of data and since compactions are stopped, we might end up accumulating too much data on level 0.

We should add a timeout to this function. 5 or 10 seconds timeout. @ashish-goswami thoughts?

jarifibrahim:
I think we could make this simpler. I haven't tried running it. We also need some tests for this PR. The `levels_test.go` file has tests for compactions that can be used for testing this change.

```diff
diff --git a/db.go b/db.go
index 4e5a384..267fd81 100644
--- a/db.go
+++ b/db.go
@@ -1430,16 +1430,16 @@ func (db *DB) startMemoryFlush() {
 // stopped. Ideally, no writes are going on during Flatten. Otherwise, it would create competition
 // between flattening the tree and new tables being created at level zero.
 func (db *DB) Flatten(workers int) error {
-	return db.flattenInternal(workers, true)
+	return db.flattenInternal(workers, false)
 }
 
 // CompactLSM works like Flatten except it only performs one pass of compactions instead of
 // performing compactions until the whole database is in a single level like Flatten does.
 func (db *DB) CompactLSM(workers int) error {
-	return db.flattenInternal(workers, false)
+	return db.flattenInternal(workers, true)
 }
 
-func (db *DB) flattenInternal(workers int, loop bool) error {
+func (db *DB) flattenInternal(workers int, singlePass bool) error {
 	db.stopCompactions()
 	defer db.startCompactions()
 
@@ -1486,6 +1486,19 @@ func (db *DB) flattenInternal(workers int, loop bool) error {
 				levels = append(levels, i)
 			}
 		}
+
+		// Single pass means we won't try to flatten the entire LSM Tree. We
+		// will try to push the data to the lower levels, once per level.
+		if singlePass {
+			for _, l := range levels {
+				// Create an artificial compaction priority to ensure that we compact the level.
+				cp := compactionPriority{level: l, score: 1.74}
+				if err := compactAway(cp); err != nil {
+					return err
+				}
+			}
+			return nil
+		}
 		if len(levels) <= 1 {
 			prios := db.lc.pickCompactLevels()
 			if len(prios) == 0 || prios[0].score <= 1.0 {
@@ -1495,11 +1508,6 @@ func (db *DB) flattenInternal(workers int, loop bool) error {
 			if err := compactAway(prios[0]); err != nil {
 				return err
 			}
-
-			// If loop is false, only perform one pass of the outer for loop.
-			if !loop {
-				return nil
-			}
 			continue
 		}
 		// Create an artificial compaction priority, to ensure that we compact the level.
@@ -1507,11 +1515,6 @@ func (db *DB) flattenInternal(workers int, loop bool) error {
 		if err := compactAway(cp); err != nil {
 			return err
 		}
-
-		// If loop is false, only perform one pass of the outer for loop.
-		if !loop {
-			return nil
-		}
 	}
 }
 
```

jarifibrahim:
We don't need this PR anymore. We can close it.

PR #62, opened 2017-06-14
Windows support

Motakjuq:
Add windows support

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/62)

PR #1451, opened 2020-08-04
Fix build on Plan 9

fhs:
Plan 9 doesn't have mmap, so we just return an error.

File locking is implemented using the exclusive mode bit.
From http://man.cat-v.org/plan_9/5/open:

>   If the file is marked for exclusive use (see stat(5)), only
>   one client can have the file open at any time.  That is,
>   after such a file has been opened, further opens will fail
>   until fid has been clunked.

Initial attempt at implementing directory locking using this method was
not successful because badger seems to be opening the directory multiple
times (at least while running the test suite). As a workaround, we lock
the pid lock file instead of the directory itself, same as the windows
implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1451)

### DIFF
--- a/dir_plan9.go
+++ b/dir_plan9.go
@@ -37,7 +37,8 @@ type directoryLockGuard struct {
 // acquireDirectoryLock gets a lock on the directory.
 // It will also write our pid to dirPath/pidFileName for convenience.
 // readOnly is not supported on Plan 9.
-func acquireDirectoryLock(dirPath string, pidFileName string, readOnly bool) (*directoryLockGuard, error) {
+func acquireDirectoryLock(dirPath string, pidFileName string, readOnly bool) (
+	*directoryLockGuard, error) {
 	if readOnly {
 		return nil, ErrPlan9NotSupported
 	}
--- a/dir_plan9.go
+++ b/dir_plan9.go
@@ -52,6 +52,13 @@ func acquireDirectoryLock(dirPath string, pidFileName string, readOnly bool) (*d
 	// If the file was unpacked or created by some other program, it might not
 	// have the ModeExclusive bit set. Set it before we call OpenFile, so that we
 	// can be confident that a successful OpenFile implies exclusive use.
+	//
+	// OpenFile fails if the file ModeExclusive bit set *and* the file is already open.
+	// So, if the file is closed when the DB crashed, we're fine. When the process
+	// that was managing the DB crashes, the OS will close the file for us.
+	//
+	// This bit of code is copied from Go's lockedfile internal package:
+	// https://github.com/golang/go/blob/go1.15rc1/src/cmd/go/internal/lockedfile/lockedfile_plan9.go#L58
 	if fi, err := os.Stat(absPidFilePath); err == nil {
 		if fi.Mode()&os.ModeExclusive == 0 {
 			if err := os.Chmod(absPidFilePath, fi.Mode()|os.ModeExclusive); err != nil {
--- a/dir_plan9.go
+++ b/dir_plan9.go
@@ -28,7 +28,7 @@ import (
 // directoryLockGuard holds a lock on a directory and a pid file inside.  The pid file isn't part
 // of the locking mechanism, it's just advisory.
 type directoryLockGuard struct {
-	// File handle on the directory, which we've flocked.
+	// File handle on the directory, which we've locked.
 	f *os.File
 	// The absolute path to our pid file.
 	path string
@@ -72,7 +72,7 @@ func acquireDirectoryLock(dirPath string, pidFileName string, readOnly bool) (*d
 	}
 
 	_, err = fmt.Fprintf(f, "%d\n", os.Getpid())
-	if err != nil {
+	if _, err = fmt.Fprintf(f, "%d\n", os.Getpid()); err != nil {
 		f.Close()
 		return nil, errors.Wrapf(err, "could not write pid")
 	}
--- a/dir_plan9.go
+++ b/dir_plan9.go
@@ -1,7 +1,5 @@
-// +build !windows,!plan9
-
 /*
- * Copyright 2017 Dgraph Labs, Inc. and Contributors
+ * Copyright 2020 Dgraph Labs, Inc. and Contributors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,12 +18,11 @@ package badger
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/pkg/errors"
-	"golang.org/x/sys/unix"
 )
 
 // directoryLockGuard holds a lock on a directory and a pid file inside.  The pid file isn't part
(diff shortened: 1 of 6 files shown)


dir_plan9.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot get absolute path for pid lock file")
+	}
+
+	// If the file was unpacked or created by some other program, it might not
+	// have the ModeExclusive bit set. Set it before we call OpenFile, so that we
+	// can be confident that a successful OpenFile implies exclusive use.
+	if fi, err := os.Stat(absPidFilePath); err == nil {
+		if fi.Mode()&os.ModeExclusive == 0 {
+			if err := os.Chmod(absPidFilePath, fi.Mode()|os.ModeExclusive); err != nil {
```
jarifibrahim, line 59: Let's say your DB crashed. Now when we reopen it, we will set the `ModeExclusive` bit here and then the `openFile` would fail because the file already has `ModeExclusive` bit set, isn't it?

dir_plan9.go
```diff
@@ -0,0 +1,146 @@
...
+		if isLocked(err) {
+			return nil, errors.Wrapf(err,
+				"Cannot open pid lock file %q.  Another process is using this Badger database",
+				absPidFilePath)
+		}
+		return nil, errors.Wrapf(err, "Cannot open pid lock file %q", absPidFilePath)
+	}
+
+	_, err = fmt.Fprintf(f, "%d\n", os.Getpid())
+	if err != nil {
```
jarifibrahim, line 75: nit: Can be written in a single line.
```suggestion
	if _, err = fmt.Fprintf(f, "%d\n", os.Getpid()); err != nil {
```

dir_plan9.go
```diff
@@ -0,0 +1,146 @@
...
+		f.Close()
+		return nil, errors.Wrapf(err, "could not write pid")
+	}
+	return &directoryLockGuard{f, absPidFilePath, readOnly}, nil
+}
+
+// Release deletes the pid file and releases our lock on the directory.
+func (guard *directoryLockGuard) release() error {
+	var err error
+	if !guard.readOnly {
```
jarifibrahim, line 85: Wouldn't this always be true? We don't allow acquiring lock in a read-only mode.
We can even remove the `readOnly` from the guard itself.

dir_plan9.go
```diff
@@ -0,0 +1,146 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// directoryLockGuard holds a lock on a directory and a pid file inside.  The pid file isn't part
+// of the locking mechanism, it's just advisory.
+type directoryLockGuard struct {
+	// File handle on the directory, which we've flocked.
```
jarifibrahim, line 31:
```suggestion
	// File handle on the directory, which we've locked.
```

dir_plan9.go
fhs: `OpenFile` fails if the file `ModeExclusive` bit set *and* the file is already open. So, if the file is closed when the DB crashed, we're fine. When the process that was managing the DB crashes, the OS will close the file for us.

This bit of code is actually copied from Go's lockedfile internal package: https://github.com/golang/go/blob/go1.15rc1/src/cmd/go/internal/lockedfile/lockedfile_plan9.go#L58

dir_plan9.go
fhs: I was trying to keep the code similar to `dir_unix.go`, but I can remove `readOnly`.

dir_plan9.go
jarifibrahim: @fhs can you please add your comment as the code comment and add the link too for future references?

dir_plan9.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		if isLocked(err) {
+			return nil, errors.Wrapf(err,
+				"Cannot open pid lock file %q.  Another process is using this Badger database",
+				absPidFilePath)
+		}
+		return nil, errors.Wrapf(err, "Cannot open pid lock file %q", absPidFilePath)
+	}
+
+	_, err = fmt.Fprintf(f, "%d\n", os.Getpid())
```
jarifibrahim, line 74: This should be removed now since we added it to the line below.

dir_plan9.go
fhs: Done

dir_plan9.go
fhs: Done

dir_plan9.go
fhs: Oops, done.

fhs:
I've added `GOOS=plan9` build to travis CI.

dir_plan9.go
```diff
@@ -0,0 +1,147 @@
...
+	//
+	// This bit of code is copied from Go's lockedfile internal package:
+	// https://github.com/golang/go/blob/go1.15rc1/src/cmd/go/internal/lockedfile/lockedfile_plan9.go#L58
+	if fi, err := os.Stat(absPidFilePath); err == nil {
+		if fi.Mode()&os.ModeExclusive == 0 {
+			if err := os.Chmod(absPidFilePath, fi.Mode()|os.ModeExclusive); err != nil {
+				return nil, errors.Wrapf(err, "could not set exclusive mode bit")
+			}
+		}
+	}
```
NamanJain8, line 68: If err in line #62 is not nil, shouldn't we return with the error code?

y/file_dsync.go
```diff
@@ -1,4 +1,4 @@
-// +build !dragonfly,!freebsd,!windows
+// +build !dragonfly,!freebsd,!windows,!plan9
```
NamanJain8, line 1: Shouldn't `plan9` be present in `y/file_nodsync.go` file? Otherwise the `datasyncFileFlag` would always be 0.

dir_plan9.go
jarifibrahim: Nice catch @NamanJain8 . We might need something like this 
```suggestion
	} else if !os.IsNotExist(err) {
		return nil, err
	}
```

y/file_dsync.go
fhs: Plan 9's `syscall.O_SYNC` has a made up value of 0 (since Plan 9 doesn't support it) but you're right, `plan9` tag should be added to `file_nodsync.go` for consistency.

dir_plan9.go
fhs: Yep, that's how it's done in the `lockedfile` package, where I copied this from. Don't know how I missed it. Done.

y/file_dsync.go
fhs: Done

jarifibrahim (approved):
LGTM. Thanks for the PR @fhs :tada:  . I made one small line length change on this PR.

I'll merge this PR once CI finishes.

fhs:
When can I expect this to be included in a release? I need this in a v1.6.x release to get ipfs working, but I can also wait for v1.7.0. IPFS tracking issue for Plan 9 support: https://github.com/ipfs/go-ipfs/issues/7575

jarifibrahim:
We will be doing a badger 1.x release this week. This patch would be in that release.

@NamanJain8 This needs to go in 1.x release.

francislavoie:
Looks like this was never backported to badger v2. Could we make that happen?

Building against `v2.2007.3` which is the version Caddy is currently depending on (via smallstep) yields these errors:

```
# github.com/dgraph-io/badger/v2/y
..\..\..\..\go\pkg\mod\github.com\dgraph-io\badger\v2@v2.2007.3\y\file_dsync.go:24:21: undefined: unix.O_DSYNC
..\..\..\..\go\pkg\mod\github.com\dgraph-io\badger\v2@v2.2007.3\y\mmap_unix.go:30:11: undefined: unix.PROT_READ
..\..\..\..\go\pkg\mod\github.com\dgraph-io\badger\v2@v2.2007.3\y\mmap_unix.go:32:12: undefined: unix.PROT_WRITE
..\..\..\..\go\pkg\mod\github.com\dgraph-io\badger\v2@v2.2007.3\y\mmap_unix.go:34:9: undefined: unix.Mmap
..\..\..\..\go\pkg\mod\github.com\dgraph-io\badger\v2@v2.2007.3\y\mmap_unix.go:34:54: undefined: unix.MAP_SHARED
..\..\..\..\go\pkg\mod\github.com\dgraph-io\badger\v2@v2.2007.3\y\mmap_unix.go:39:9: undefined: unix.Munmap
..\..\..\..\go\pkg\mod\github.com\dgraph-io\badger\v2@v2.2007.3\y\mmap_unix.go:46:11: undefined: unix.MADV_NORMAL
..\..\..\..\go\pkg\mod\github.com\dgraph-io\badger\v2@v2.2007.3\y\mmap_unix.go:48:11: undefined: unix.MADV_RANDOM
..\..\..\..\go\pkg\mod\github.com\dgraph-io\badger\v2@v2.2007.3\y\mmap_unix.go:50:9: undefined: unix.Madvise
```

PR #1457, opened 2020-08-14
Fix ever existing iterator and transaction in write benchmark

ahsanbarkati:
This PR adds an option to enable showing keys statistics in the write benchmark. The introduced flag is `--show-keys` and is disabled by default. This PR also fixes the ever existing iterators in the `reportStats()` function.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1457)

### DIFF
--- a/badger/cmd/write_bench.go
+++ b/badger/cmd/write_bench.go
@@ -301,6 +301,32 @@ func writeBench(cmd *cobra.Command, args []string) error {
 	return err
 }
 
+func updateKeysStats(db *badger.DB) {
+	txn := db.NewTransaction(false)
+	defer txn.Discard()
+
+	iopt := badger.DefaultIteratorOptions
+	iopt.AllVersions = true
+	iopt.InternalAccess = true
+	it := txn.NewIterator(iopt)
+	defer it.Close()
+
+	for it.Rewind(); it.Valid(); it.Next() {
+		i := it.Item()
+		if bytes.HasPrefix(i.Key(), []byte("!badger!")) {
+			internalKeyCount++
+		}
+		if bytes.HasPrefix(i.Key(), []byte("!badger!Move")) {
+			moveKeyCount++
+		}
+		if i.IsDeletedOrExpired() {
+			invalidKeyCount++
+		} else {
+			validKeyCount++
+		}
+	}
+}
+
 func reportStats(c *y.Closer, db *badger.DB) {
 	defer c.Done()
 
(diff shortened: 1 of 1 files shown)


badger/cmd/write_bench.go
```diff
@@ -309,28 +312,35 @@ func reportStats(c *y.Closer, db *badger.DB) {
...
-					moveKeyCount++
-				}
-				if i.IsDeletedOrExpired() {
-					invalidKeyCount++
-				} else {
-					validKeyCount++
-				}
+			if showKeysCount {
+				func() {
+					txn := db.NewTransaction(false)
```
jarifibrahim, line 317: Make this a different function and just call it here.

badger/cmd/write_bench.go
```diff
@@ -309,28 +312,35 @@ func reportStats(c *y.Closer, db *badger.DB) {
...
+							moveKeyCount++
+						}
+						if i.IsDeletedOrExpired() {
+							invalidKeyCount++
+						} else {
+							validKeyCount++
+						}
+					}
+				}()
+				fmt.Printf("Valid Keys Count: %d\nInvalid Keys Count: %d\nMove Keys Count:"+
```
jarifibrahim, line 341: Remove the new lines. Make it look like.
```
Valid Keys: X Invalid Keys: Y Move Keys: Z Internal Key: A
```

badger/cmd/write_bench.go
ahsanbarkati: Done

badger/cmd/write_bench.go
ahsanbarkati: Done

badger/cmd/write_bench.go
```diff
@@ -298,6 +301,32 @@ func writeBench(cmd *cobra.Command, args []string) error {
...
+	iopt := badger.DefaultIteratorOptions
+	iopt.AllVersions = true
+	iopt.InternalAccess = true
+	it := txn.NewIterator(iopt)
+	defer it.Close()
+
+	for it.Rewind(); it.Valid(); it.Next() {
+		i := it.Item()
+		if bytes.HasPrefix(i.Key(), []byte("!badger!")) {
+			internalKeyCount++
```
jarifibrahim, line 317: Move these variables inside this function and then print them at the end in this function itself. We don't want to accumulate all the counts.

PR #1460, opened 2020-08-15
fix(writebench): Set gcperiod to zero

jarifibrahim:
Fix flag description and change default gc period.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1460)

### DIFF
--- a/badger/cmd/write_bench.go
+++ b/badger/cmd/write_bench.go
@@ -131,7 +131,7 @@ func init() {
 		"If set, drop random prefixes periodically over given duration.")
 	writeBenchCmd.Flags().StringVar(&ttlDuration, "entry-ttl", "0s",
 		"TTL duration in seconds for the entries, 0 means without TTL")
-	writeBenchCmd.Flags().StringVarP(&gcPeriod, "gc-every", "g", "0", "GC Period.")
+	writeBenchCmd.Flags().StringVarP(&gcPeriod, "gc-every", "g", "0s", "GC Period.")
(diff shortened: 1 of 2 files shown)


badger/cmd/write_bench.go
```diff
@@ -126,12 +126,12 @@ func init() {
...
 	writeBenchCmd.Flags().StringVar(&dropAllPeriod, "dropall", "0s",
-		"Period of dropping all. If 0, doesn't drops all.")
+		"If set, run dropAll periodically over given duration.")
 	writeBenchCmd.Flags().StringVar(&dropPrefixPeriod, "drop-prefix", "0s",
-		"Period of dropping by random prefixes. If 0, doesn't drops by prefix.")
+		"If set, drop random prefixes periodically over given duration.")
 	writeBenchCmd.Flags().StringVar(&ttlDuration, "entry-ttl", "0s",
 		"TTL duration in seconds for the entries, 0 means without TTL")
-	writeBenchCmd.Flags().StringVarP(&gcPeriod, "gc-every", "g", "5m", "GC Period.")
+	writeBenchCmd.Flags().StringVarP(&gcPeriod, "gc-every", "g", "0", "GC Period.")
```
ahsanbarkati, line 134: To be consistent with all other options, can we have `0s` in place of 0. Or the other way around i.e. having 0 instead of 0s in the other options?

jarifibrahim: Fixed.

PR #1465, opened 2020-08-18
Fix(cleanup): Avoid truncating in value.Open on error

jarifibrahim:
The vlog.Open() function can return an error denoting that the open was
unsuccessful but we have `db.cleanup` which will be called to stop all
the running goroutines in case badger couldn't start. The db.cleanup
function calls vlog.Close() which will truncate the maxFid vlog file
based on the vlog.writableLogOffset. The vlog.writableLogOffset was not
being updated in case open failed. As a result of this, we will end up
truncating the vlog file to length 0 if open fails.

This PR fixes this by using `vlog.stopDiscardStatFlush()` instead of `vlog.close`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1465)

### DIFF
--- a/db.go
+++ b/db.go
@@ -447,7 +447,10 @@ func (db *DB) cleanup() {
 	}
 
 	db.orc.Stop()
-	db.vlog.Close()
+
+	// Do not use vlog.Close() here. vlog.Close truncates the files. We don't
+	// want to truncate files unless the user has specified the truncate flag.
+	db.vlog.stopFlushDiscardStats()
 }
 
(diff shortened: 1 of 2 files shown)


ashish-goswami (approved):
:lgtm: 

> Reviewed 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1465#-:-MFjT9wg4_NBxDbAoKhq:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @manishrjain)

PR #1475, opened 2020-08-25
Add IsClosed method

jarifibrahim:
Add an `IsClosed` method which denotes if badger instance is closed or not.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1475)

### DIFF
--- a/db.go
+++ b/db.go
@@ -84,6 +84,7 @@ type DB struct {
 	logRotates int32
 
 	blockWrites int32
+	isClosed    uint32
 
 	orc *oracle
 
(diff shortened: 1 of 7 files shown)


jarifibrahim:
This PR should be cherry picked to older versions.

NamanJain8 (approved):
Does the close check need to be done at `db.NewTransaction()` too? Looks like not necessary, but I am not very sure.

jarifibrahim:
>Does the close check need to be done at db.NewTransaction() too? Looks like not necessary, but I am not very sure.

I could add it over there as well but we don't necessarily need it. Creating a transaction is okay but when you do write you'll get a ErrDBClosed and when you do a read (via Get or an Iterator), we'll return error (or a panic if it's an iterator).

PR #1473, opened 2020-08-21
Rework Block and Index cache

jarifibrahim:
This PR separates the block cache and the index cache. Earlier we had a mix of both the caches. 
Block Cache -> Stores block. Should be used when running badger with compression/encryption
Index Cache -> Used to limit the memory used by table indices and bloom filters. 

This PR also remove the `KeepBlocksInCache` and `KeepBlockIndicesInCache` options.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1473)

PR #1476, opened 2020-08-25
release/v2.2007 - Rework Block and Index cache (#1473)

jarifibrahim:
There were a bunch of conflicts. Please review it thoroughly.

This PR separates the block cache and the index cache. Earlier we had a mix of both the caches.
Block Cache -> Stores block. Should be used when running badger with compression/encryption
Index Cache -> Used to limit the memory used by table indices and bloom filters.

This PR also remove the `KeepBlocksInCache` and `KeepBlockIndicesInCache` options.

(cherry picked from commit c3b15824dc149149616d474e5e150fa8ca337d7f)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1476)

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -76,8 +76,8 @@ type Options struct {
 	// Compression indicates the compression algorithm used for block compression.
 	Compression options.CompressionType
 
-	Cache   *ristretto.Cache
-	BfCache *ristretto.Cache
+	BlockCache *ristretto.Cache
+	IndexCache *ristretto.Cache
 
 	// ZSTDCompressionLevel is the ZSTD compression level used for compressing blocks.
 	ZSTDCompressionLevel int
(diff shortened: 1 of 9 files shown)


table/table.go
```diff
@@ -548,15 +543,10 @@ func (t *Table) blockCacheKey(idx int) []byte {
 	return buf
 }
 
-// blockOffsetsCacheKey returns the cache key for block offsets.
-func (t *Table) blockOffsetsCacheKey() []byte {
-	y.AssertTrue(t.id < math.MaxUint32)
-	buf := make([]byte, 6)
-	buf[0] = 'b'
-	buf[1] = 'o'
```
NamanJain8, line 556: Both ways it will be fine, right? Was there any reasoning behind having this earlier?

NamanJain8 (approved):

table/table.go
ahsanbarkati: I think it was there so that there is no conflict in the cache keys (since we had a mixed cache for blocks and indices before). But now we have separated the caches so conflict shouldn't arise.

ahsanbarkati (approved):

table/table.go
jarifibrahim: Earlier all the items, index and block were in the same cache and that's why there is this comment in BFCacheKey() function.

> 	// Without the "bf" prefix, we will have conflict with the blockCacheKey.

The bloom filter cache was added in https://github.com/dgraph-io/badger/pull/1260 but we didn't fix the old cache keys.

With this PR
Block Cache => Blocks [key: tableID+blockID]
Index Cache => BlockOffets[key: tableID] and BF[Key:"bf"+tableID]

@NamanJain8 @ahsanbarkati

PR #1477, opened 2020-08-25
fix 32 bit build & remove dependency on zstd when cgo is disabled

Stebalien:
1. Fix 32bit builds by explicitly setting the maxVlogFileSize to uint32.
2. Remove dependency on zstd when cgo is disabled.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1477)

### DIFF
--- a/value.go
+++ b/value.go
@@ -47,7 +47,7 @@ import (
 
 // maxVlogFileSize is the maximum size of the vlog file which can be created. Vlog Offset is of
 // uint32, so limiting at max uint32.
-var maxVlogFileSize = math.MaxUint32
+var maxVlogFileSize uint32 = math.MaxUint32
 
 // Values have their first byte being byteData or byteDelete. This helps us distinguish between
 // a key that has never been seen and a key that has been explicitly deleted.


jarifibrahim (approved):
Thanks for the PR @Stebalien :tada:

NamanJain8 (approved):

jarifibrahim:
@Stebalien, the 32-bit build is still failing. Look at https://travis-ci.org/github/dgraph-io/badger/jobs/721170743

Stebalien:
Ah, well, the _build_ succeeds... Let me check.

Stebalien:
Ah, it's probably because the _usable_ addressspace on a 32bit system is less than 32 bits. The default `ValueLogFileSize` may need to be reduced.

jarifibrahim:
Hey @Stebalien, the fix for cgo was merged via https://github.com/dgraph-io/badger/commit/806a325a0462a4970593ba3b95ca31c2cde8c82a . Can you update this PR? I'd like to merge this PR for the uint32 overflow error.

Stebalien:
I've rebased but I'm not sure how you want to fix the ValueLogFileSize issue.

jarifibrahim:
> Ah, it's probably because the _usable_ addressspace on a 32bit system is less than 32 bits. The default `ValueLogFileSize` may need to be reduced.

A 32 bit machine should be able to address up to 4 GB but we're allocating only 2 GB upfront.

jarifibrahim:
I've raised https://github.com/dgraph-io/badger/pull/1496 to mmap only the file size that we need (I don't know why we were mmaping 2x the size). This might fix the `cannot allocate memory issue`.

Stebalien:
> A 32 bit machine should be able to address up to 4 GB but we're allocating only 2 GB upfront.

Not quite. There are usually some hardware restrictions, reserved kernel address space, and flag bits. Most 32bit systems can address up to 2-3GiB. But yeah, 2GiB _should_ be fine on most systems so I'm not sure.

Stebalien:
Closing in favor of #1496.

PR #1496, opened 2020-09-02
Mmap value log file to options.ValueLogFileSize

jarifibrahim:
Currently, we mmap the newly created value log files to
2*options.ValueLogFileSize. This isn't necessary and causes issues on 32
bit systems.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1496)

### DIFF
--- a/value.go
+++ b/value.go
@@ -47,7 +47,7 @@ import (
 
 // maxVlogFileSize is the maximum size of the vlog file which can be created. Vlog Offset is of
 // uint32, so limiting at max uint32.
-var maxVlogFileSize = math.MaxUint32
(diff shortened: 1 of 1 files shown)


Stebalien:
The fix looks fine, but the test failures look problematic.

jarifibrahim:
The 2x file size was done so that we don't cause a segmentation fault if the vlog file size goes beyond its limit at runtime.

I'm closing this PR since we don't need it.

Stebalien:
> I'm closing this PR since we don't need it.

Does this mean 32bit CPUs are not supported?

PR #1505, opened 2020-09-08
(release/v1.6) GC: Consider size of value while rewriting (#1357)

### LINKED ISSUE #1292: Value log rewrites load the entire value log into memory

<pre>
$ go version
go version go1.13.8 linux/amd64
</pre>

1.6.1

### Does this issue reproduce with the latest master?

I haven't tried but the code looks the same.

### Description

When rewriting a value log (e.g., on garbage collection), badger will end up loading the entire value log into memory unless there are enough small values to exceed the batch count (https://github.com/ipfs/go-ds-badger/issues/86).

When estimating the size of an entry:

https://github.com/dgraph-io/badger/blob/8a93a41fef4ba548fc241157164282cb6ace892f/structs.go#L153-L158

Badger only estimates the size of the entry as it exists in the LSM.

When deciding whether or not to flush, badger only looks at this estimated size, not the size of the value itself:

https://github.com/dgraph-io/badger/blob/617ed7c7db9d618b6511adfff5d22dcde2233049/value.go#L566-L576

I _believe_ the fix here is to look at the size of the _value_ plus the LSM overhead, is that correct?

NamanJain8:
The existing code doesn't consider the size of the value while
calculating the size of the entry batch in rewrite method.
This PR fixes it.

Fixes #1292

(cherry picked from commit 14386ac9b764cb6655cb0da7c848e244369a2a27)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1505)

### DIFF
--- a/value.go
+++ b/value.go
@@ -411,6 +411,11 @@ func (vlog *valueLog) rewrite(f *logFile, tr trace.Trace) error {
 
 			ne.Value = append([]byte{}, e.Value...)
 			es := int64(ne.estimateSize(vlog.opt.ValueThreshold))
+			// Consider size of value as well while considering the total size
+			// of the batch. There have been reports of high memory usage in
+			// rewrite because we don't consider the value size. See #1292.
+			es += int64(len(e.Value))
+
 			// Ensure length and size of wb is within transaction limits.
 			if int64(len(wb)+1) >= vlog.opt.maxBatchCount ||
 				size+es >= vlog.opt.maxBatchSize {


jarifibrahim (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1505#-:-MGmZ1BuAGLFggLezi_I:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @ashish-goswami, @jarifibrahim, and @manishrjain)

PR #1507, opened 2020-09-08
(release/v1.6) levels: Compaction incorrectly drops some delete markers (#1422)

NamanJain8:
fd8989493 ("Compaction: Expired keys and delete markers are never
purged") revealed a bug in the compaction logic that leads to delete
markers being incorrectly dropped during compaction.

During compaction, `*levelsController.compactBuildTables` decides to
drop a delete key if there is no overlap with levels lower than the
bottom layer of the compaction definition.

It does that by checking only the `top` table, thinking that proving
that the `top` table doesn't overlap is sufficient to prove that
the `bottom` table doesn't. Unfortunately, this is not the case.
Not in general, and even less in the case of `DropPrefix()` where
we run a same-level compaction and `top` is empty.

The faulty condition was introduced way back when in e597fb721
("Discard key versions during compaction").

(cherry picked from commit 5b90893d2d6014e816047a80207feb614e8ec463)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1507)

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -439,11 +439,8 @@ func (s *levelsController) compactBuildTables(
 	botTables := cd.bot
 
 	// Check overlap of the top level with the levels which are not being
-	// compacted in this compaction. We don't need to check overlap of the bottom
-	// tables with other levels because if the top tables overlap with any of the lower
-	// levels, it implies bottom level also overlaps because top and bottom tables
-	// overlap with each other.
-	hasOverlap := s.checkOverlap(cd.top, cd.nextLevel.level+1)
+	// compacted in this compaction.
+	hasOverlap := s.checkOverlap(cd.allTables(), cd.nextLevel.level+1)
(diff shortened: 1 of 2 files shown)


jarifibrahim (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1507#-:-MGn0fOR9T13qwgrUynY:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @ashish-goswami, @jarifibrahim, and @manishrjain)

jarifibrahim (approved):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1507)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @ashish-goswami and @manishrjain)

PR #1525, opened 2020-09-18
Support force flushing memtable via DB options

jarifibrahim:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1525)


jarifibrahim:
We don't need this PR.

PR #1529, opened 2020-09-20
fix(writeBatch): Avoid deadlock in commit callback

jarifibrahim:
`wb.Flush()` would call `wb.commit()` while holding lock which would call
`wb.callback` in case of an error while committing.

`wb.callback` would try to acquire the lock `wb.Flush` is holding and this
would lead to a deadlock. This PR fixes it.

The test in this PR fails on master.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1529)

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -82,6 +82,7 @@ func (wb *WriteBatch) Cancel() {
 	wb.txn.Discard()
 }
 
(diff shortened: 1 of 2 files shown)


NamanJain8 (approved):
Nice catch!

PR #1530, opened 2020-09-21
fix(getHead): Use forward iteration and add tests

jarifibrahim:
getHead is using reverse iteration and there have been issues with
reverse iteration (see https://dgraph.io/docs/badger/faq/#reverse-iteration-doesn-t-give-me-the-right-results).
This PR uses forward iteration to find the head key.
This PR doesn't change any functionality, it only changes the direction
of iteration.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1530)

### DIFF
--- a/batch_test.go
+++ b/batch_test.go
@@ -18,6 +18,7 @@ package badger
 
 import (
 	"fmt"
+	"math"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


jarifibrahim:
We don't need this PR anymore.

PR #1539, opened 2020-09-23
perf(GC): Remove move keys

jarifibrahim:
Move keys are used to fix the invalid value pointers when vlog GC runs.
This PR removes the move keys. To ensure we return the correct results
for Get calls, we will search all the levels of the LSM tree to find the
correct entry. Earlier, we used to return the first entry whose version
was less than or equal to the required version.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1539)

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -269,21 +269,13 @@ func (s *levelsController) dropTree() (int, error) {
 }
 
 // dropPrefix runs a L0->L1 compaction, and then runs same level compaction on the rest of the
-// levels. For L0->L1 compaction, it runs compactions normally, but skips over all the keys with the
-// provided prefix and also the internal move keys for the same prefix.
+// levels. For L0->L1 compaction, it runs compactions normally, but skips over
+// all the keys with the provided prefix.
 // For Li->Li compactions, it picks up the tables which would have the prefix. The
 // tables who only have keys with this prefix are quickly dropped. The ones which have other keys
 // are run through MergeIterator and compacted to create new tables. All the mechanisms of
 // compactions apply, i.e. level sizes and MANIFEST are updated as in the normal flow.
 func (s *levelsController) dropPrefixes(prefixes [][]byte) error {
-	// Internal move keys related to the given prefix should also be skipped.
-	for _, prefix := range prefixes {
-		key := make([]byte, 0, len(badgerMove)+len(prefix))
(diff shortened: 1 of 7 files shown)


levels.go
```diff
@@ -1102,17 +1096,14 @@ func (s *levelsController) get(key []byte, maxVs *y.ValueStruct, startLevel int)
 		if vs.Value == nil && vs.Meta == 0 {
 			continue
 		}
-		if maxVs == nil || vs.Version == version {
+		if vs.Version == version {
```
NamanJain8, line 1099: Earlier `maxVs == nil` was used to get the first found key? But the invariant that _entry with an older timestamp of the same key should not be written_ is no more valid.

NamanJain8 (approved):

jarifibrahim (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1539#-:-MIKAc8kAiVJ0L6jIZPV:bnfp4nl)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @ashish-goswami, @jarifibrahim, @manishrjain, and @NamanJain8)

___
*[levels.go, line 1099 at r1](https://reviewable.io/reviews/dgraph-io/badger/1539#-MHvZ8T7lDiOT0cELZ5X-r1-1099:-MIKAAP8EZ11UdE6p55z:bysaci1) ([raw file](https://github.com/dgraph-io/badger/blob/c42e04c9072f5e871a8ff11e0fd28aa32347e808/levels.go#L1099)):*
<details><summary><i>Previously, NamanJain8 (Naman Jain) wrote…</i></summary><blockquote>

Earlier `maxVs == nil` was used to get the first found key? But the invariant that "_entry with an older timestamp of the same key should not be written_" is no more valid.
</blockquote></details>

We overwrite the keys. GC will insert the same key version. You can have keys with smaller version above keys with larger version.

manishrjain (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1539#-:-MIKAnfNAGDRVFaomTbo:bnfp4nl)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @ashish-goswami, @jarifibrahim, @manishrjain, and @NamanJain8)

PR #1541, opened 2020-09-24
Fix integer overflow error when building for 386

gammazero:
The untyped const causes conversion to int, for which the value is too large.
Use a typed const instead that is large enough to store value.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1541)

### DIFF
--- a/badger/cmd/backup.go
+++ b/badger/cmd/backup.go
@@ -55,7 +55,7 @@ func doBackup(cmd *cobra.Command, args []string) error {
 	opt := badger.DefaultOptions(sstDir).
 		WithValueDir(vlogDir).
 		WithTruncate(truncate).
-		WithNumVersionsToKeep(int(math.MaxUint32))
+		WithNumVersionsToKeep(int(math.MaxInt32))
 
 	if numVersions > 0 {
 		opt.NumVersionsToKeep = numVersions
--- a/badger/cmd/backup.go
+++ b/badger/cmd/backup.go
@@ -55,7 +55,7 @@ func doBackup(cmd *cobra.Command, args []string) error {
 	opt := badger.DefaultOptions(sstDir).
 		WithValueDir(vlogDir).
 		WithTruncate(truncate).
-		WithNumVersionsToKeep(math.MaxUint32)
+		WithNumVersionsToKeep(int(math.MaxUint32))
 
 	if numVersions > 0 {
 		opt.NumVersionsToKeep = numVersions
--- a/badger/cmd/backup.go
+++ b/badger/cmd/backup.go
@@ -55,7 +55,7 @@ func doBackup(cmd *cobra.Command, args []string) error {
 	opt := badger.DefaultOptions(sstDir).
 		WithValueDir(vlogDir).
 		WithTruncate(truncate).
-		WithNumVersionsToKeep(int(math.MaxInt32))
+		WithNumVersionsToKeep(math.MaxInt32)
 
 	if numVersions > 0 {
 		opt.NumVersionsToKeep = numVersions
--- a/badger/cmd/restore.go
+++ b/badger/cmd/restore.go
@@ -68,7 +68,7 @@ func doRestore(cmd *cobra.Command, args []string) error {
 	// Open DB
 	db, err := badger.Open(badger.DefaultOptions(sstDir).
 		WithValueDir(vlogDir).
-		WithNumVersionsToKeep(int(math.MaxUint32)))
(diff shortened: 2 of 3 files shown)


badger/cmd/backup.go
```diff
@@ -55,7 +55,7 @@ func doBackup(cmd *cobra.Command, args []string) error {
 	opt := badger.DefaultOptions(sstDir).
 		WithValueDir(vlogDir).
 		WithTruncate(truncate).
-		WithNumVersionsToKeep(math.MaxUint32)
+		WithNumVersionsToKeep(int(math.MaxUint32))
```
damz, line 58: Unless I'm missing something, this doesn't seem necessary.

badger/cmd/restore.go
```diff
@@ -68,7 +68,7 @@ func doRestore(cmd *cobra.Command, args []string) error {
 	// Open DB
 	db, err := badger.Open(badger.DefaultOptions(sstDir).
 		WithValueDir(vlogDir).
-		WithNumVersionsToKeep(math.MaxUint32))
+		WithNumVersionsToKeep(int(math.MaxUint32)))
```
damz, line 71: This either.

value.go
```diff
@@ -48,7 +48,7 @@ import (
 
 // maxVlogFileSize is the maximum size of the vlog file which can be created. Vlog Offset is of
 // uint32, so limiting at max uint32.
-var maxVlogFileSize = math.MaxUint32
+var maxVlogFileSize = uint32(math.MaxUint32)
```
damz, line 51: That does make sense, untyped numeric variables are `int`, which is not big enough to hold `math.MaxUint32` on 32 bit architectures.

damz (changes requested):

badger/cmd/backup.go
gammazero: Right, this should be `WithNumVersionsToKeep(int(math.MaxInt32))`

damz: In any case, the parameter is an `int` already, so you don't need to manually cast.

value.go
```diff
@@ -48,7 +48,7 @@ import (
 
 // maxVlogFileSize is the maximum size of the vlog file which can be created. Vlog Offset is of
 // uint32, so limiting at max uint32.
-var maxVlogFileSize = math.MaxUint32
+var maxVlogFileSize = int64(math.MaxUint32)
```
jarifibrahim, line 51:
```suggestion
var maxVlogFileSize uint32 = math.MaxUint32
```

Setting the variable type here would also fix the issue.

jarifibrahim (approved):
Thanks @gammazero :tada: . This might fix the build but we've seen other `cannot allocate memory` errors on 386. We can handle those in a separate PR.

@NamanJain8 review please.

jarifibrahim:
Thanks for helping with the review @damz :)

NamanJain8 (approved):
This looks good. Thanks, @gammazero.

NamanJain8:
Not directly related to the PR itself, @jarifibrahim I was wondering about the use of `int64` for `ValueLogFileSize` and `uint32` for `maxVlogFileSize`. Maybe we need to revisit the types assigned to these.

value.go
gammazero: @jarifibrahim This PR is part of fixing the 386 build issue, but the complete fix for that requires some additional changes.

I have a second PR that will provide that fix.  I would rather provide that separately, as in involves changing the amount of memory requested from mmap in multiple places, when running tests with `GOARCH=386`.  It demonstrates a fix to the problem, but may not be exactly in the form you would prefer.

jarifibrahim: @gammazero that makes sense.
Do you think making the change I requested make sense?

gammazero: @jarifibrahim Committed suggestion - makes sense.

gammazero:
The [DeepSource](https://github.com/dgraph-io/badger/pull/1541/checks?check_run_id=1178029678) and the [travis-ci](https://travis-ci.org/github/dgraph-io/badger/builds/731042033) build failures are not related to this PR, and can/should be fixed in a separate PR.

gammazero:
@jarifibrahim Once this PR is merged, I can offer a PR that fixes the 32bit build issues.

jarifibrahim:
Thanks for fixing this @gammazero

jsoriano:
Hi, will this fix be backported to v2.2007?

jsoriano:
> Hi, will this fix be backported to v2.2007?

I have create #1558 to fix this issue in v2.2007.

PR #1548, opened 2020-10-01
Make 32bit tests pass

gammazero:
This is an attempt to show a way to get the 32bit tests to pass, by asking for a smaller amount of memory from mmap.  There is likely a more ellegant solution, but this PR is to demonstrate a possibility. This only changes the size asked for when running tests with GOARCH=386, and does not affect any functionality of badger.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1548)


gammazero:
Retracting due to additional problems.

PR #1549, opened 2020-10-01
Fix(builder): Too many small tables when compression is enabled

ahsanbarkati:
This fixes the issue of too many sst files of very small sizes when compression is enabled. We now account for the actual sizes of blocks after compression and we assume that the table capacity is reached if the sum of actual sizes of block buffers is more than 90% of the table capacity.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1549)

### DIFF
--- a/table/builder.go
+++ b/table/builder.go
@@ -278,6 +278,7 @@ func (b *Builder) finishBlock() {
 	// If compression/encryption is disabled, no need to send the block to the blockChan.
 	// There's nothing to be done.
 	if b.blockChan == nil {
+		atomic.StoreUint32(&b.actualSize, b.sz)
 		b.addBlockToIndex()
 		return
 	}
--- a/table/builder.go
+++ b/table/builder.go
@@ -21,6 +21,7 @@ import (
 	"math"
 	"runtime"
 	"sync"
+	"sync/atomic"
 	"unsafe"
 
 	"github.com/dgryski/go-farm"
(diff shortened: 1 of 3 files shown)


manishrjain (approved):
:lgtm: One comment.

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1549#-:-MIZpl585Mco3wgxfL44:bpjwta0)* status: all files reviewed, 1 unresolved discussion (waiting on @ahsanbarkati, @ashish-goswami, and @jarifibrahim)

___
*[table/builder.go, line 364 at r1](https://reviewable.io/reviews/dgraph-io/badger/1549#-MIZpYBe3i8UmO42Xyx2:-MIZpYBe3i8UmO42Xyx3:b-yr7zmu) ([raw file](https://github.com/dgraph-io/badger/blob/41d19b52e79c7e6a9d2c670b4f823c726c7e3dbc/table/builder.go#L364)):*
> ```Go
> 		5*(uint32(len(b.tableIndex.Offsets))) // approximate index size
> 
> 	return estimateSz >= uint32(float64(cap)*0.90)
> ```

Do the calculation upfront, and set cap to that calculation.

table/builder.go
```diff
@@ -152,6 +154,9 @@ func (b *Builder) handleBlock() {
 		copy(b.buf[item.start:], blockBuf)
 		b.bufLock.Unlock()
 
+		// Add the actual size of current block.
+		atomic.AddUint32(&b.actualSize, uint32(len(blockBuf)))
```
jarifibrahim, line 158: This code will be executed only when we have encryption or compression enabled. What if it's not enabled? We don't start handle block goroutine

PR #1555, opened 2020-10-04
[BREAKING]: Change how Badger handles WAL

manishrjain:
This PR significantly improves Badger's disk usage behavior.

Breaking: This PR increases the magic version from 7 to 8. So, no older Badger directories would work with this change.

With this PR, we no longer use value log as write-ahead log. Instead, each MemTable has its own WAL. Value logs now only write values which are greater than ValueThreshold, while MemTable WAL only writes smaller values and value pointers.

On a crash and restart, the MemTable WALs are replayed to apply updates to Skiplist. When MemTables are flushed to L0, the corresponding WALs are deleted.

This PR makes big changes to how value log GC works:
- Discard stats are now stored in a separate file, instead of within the LSM tree.
- GC only picks up value logs based off discard stats.
- GC no longer does sampling, it uses discard stats to inform when a value log needs to be GCed.
- Value log would now no longer grow indefinitely, because of the shift to MemTable WAL.
- This PR also removes the `badger gc` tool.
- Value Log Head pointer tracking is removed.
- Only the last value log file is replayed on every start, and truncated as necessary.

This PR also makes a bunch of other changes:
- Removes ValueLogLoadingMode (always uses mmap now).
- Removes TableLoadingMode (always uses mmap now).
- Removes Truncate option.
- Removes KeepL0InMemory option.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1555)


kode54:
This renders Badger completely unusable on iOS, as it attempts to mmap 2GB at once, and fails.

baryluk:
FYI. This is not mentioned in the MR or commit message, but this also removed `LoadBloomsOnOpen` option.

PR #1560, opened 2020-10-08
feat(cmd/flatten): Add encryption-key-file flag.

danielmai:
Add an `--encryption-key-file` flag to the `badger flatten` tool to support encrypted databases.

Changes:
* Add --encryption-key-file flag
* Set block and index cache sizes to fix panic when running flatten.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1560)

### DIFF
--- a/badger/cmd/flatten.go
+++ b/badger/cmd/flatten.go
@@ -61,6 +61,8 @@ func flatten(cmd *cobra.Command, args []string) error {
 		WithValueDir(vlogDir).
 		WithNumVersionsToKeep(numVersions).
 		WithNumCompactors(0).
+		WithBlockCacheSize(100 << 20).
+		WithIndexCacheSize(200 << 20).
 		WithEncryptionKey(encKey)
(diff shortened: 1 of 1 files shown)


badger/cmd/flatten.go
```diff
@@ -43,17 +44,26 @@ func init() {
...
+	if err != nil {
+		return err
+	}
 	opt := badger.DefaultOptions(sstDir).
 		WithValueDir(vlogDir).
 		WithNumVersionsToKeep(numVersions).
-		WithNumCompactors(0)
+		WithNumCompactors(0).
+		WithBlockCacheSize(100 << 20).
+		WithIndexCacheSize(200 << 20).
```
jarifibrahim, line 65: Do we need a flat for this?

jarifibrahim (approved):

badger/cmd/flatten.go
danielmai: I don't think so. We can set some static value here as it is now.

PR #1565, opened 2020-10-14
add checks when db is opened in read-only mode

NamanJain8:
When db is opened in read-only mode, we should not create/truncate memtables or wal files. This PR addresses that issue.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1565)

### DIFF
--- a/db.go
+++ b/db.go
@@ -304,9 +304,11 @@ func Open(opt Options) (*DB, error) {
 		return nil, y.Wrapf(err, "while opening memtables")
 	}
 
-	db.mt, err = db.newMemTable()
-	if err != nil {
-		return nil, y.Wrapf(err, "cannot create memtable")
+	if !db.opt.ReadOnly {
+		db.mt, err = db.newMemTable()
+		if err != nil {
+			return nil, y.Wrapf(err, "cannot create memtable")
+		}
 	}
 
 	// newLevelsController potentially loads files in directory.
--- a/db.go
+++ b/db.go
@@ -506,7 +506,7 @@ func (db *DB) close() (err error) {
 	// and remove them completely, while the block / memtable writer is still
 	// trying to push stuff into the memtable. This will also resolve the value
 	// offset problem: as we push into memtable, we update value offsets there.
-	if !db.opt.ReadOnly && !db.mt.sl.Empty() {
+	if !db.opt.ReadOnly {
 		db.opt.Debugf("Flushing memtable")
 		for {
 			pushedFlushTask := func() bool {
@@ -534,11 +534,6 @@ func (db *DB) close() (err error) {
 	}
 	db.stopMemoryFlush()
 	db.stopCompactions()
-
-	// Delete the mem table if not already deleted.
-	if db.mt != nil {
-		db.mt.DecrRef()
-	}
 
 	// Force Compact L0
 	// We don't need to care about cstatus since no parallel compaction is running.
@@ -622,17 +617,19 @@ func (db *DB) getMemTables() ([]*memTable, func()) {
 	db.RLock()
 	defer db.RUnlock()
 
-	tables := make([]*memTable, len(db.imm)+1)
+	var tables []*memTable
 
-	// Get mutable memtable.
-	tables[0] = db.mt
-	tables[0].IncrRef()
+	if !db.opt.ReadOnly {
+		// Get mutable memtable.
+		tables = append(tables, db.mt)
+		db.mt.IncrRef()
+	}
 
 	// Get immutable memtables.
 	last := len(db.imm) - 1
 	for i := range db.imm {
-		tables[i+1] = db.imm[last-i]
-		tables[i+1].IncrRef()
+		tables = append(tables, db.imm[last-i])
+		db.imm[last-i].IncrRef()
 	}
 	return tables, func() {
 		for _, tbl := range tables {
--- a/db.go
+++ b/db.go
@@ -619,6 +619,7 @@ func (db *DB) getMemTables() ([]*memTable, func()) {
 
 	var tables []*memTable
 
+	// Mutable memtable does not exist in read-only mode.
 	if !db.opt.ReadOnly {
 		// Get mutable memtable.
 		tables = append(tables, db.mt)
--- a/db.go
+++ b/db.go
@@ -371,6 +371,7 @@ func Open(opt Options) (*DB, error) {
 }
 
 func (db *DB) MaxVersion() uint64 {
+	return 0
 	var maxVersion uint64
 	update := func(a uint64) {
 		if a > maxVersion {
@@ -541,6 +542,10 @@ func (db *DB) close() (err error) {
 	}
 	db.stopMemoryFlush()
 	db.stopCompactions()
+
+	if db.mt != nil {
+		db.mt.DecrRef()
+	}
 
 	// Force Compact L0
 	// We don't need to care about cstatus since no parallel compaction is running.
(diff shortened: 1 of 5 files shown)


manishrjain (approved):
:lgtm: Really test thoroughly for any side effects.

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1565#-:-MKjO7ABEyz4Y37oumfl:b-kisyir)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jarifibrahim)

db.go
```diff
@@ -530,6 +535,11 @@ func (db *DB) close() (err error) {
 	db.stopMemoryFlush()
 	db.stopCompactions()
 
+	// Delete the mem table if not already deleted.
+	if db.mt != nil {
```
jarifibrahim, line 539: why do you need this?

db_test.go
```diff
@@ -2188,3 +2188,24 @@ func TestUpdateMaxCost(t *testing.T) {
...
 }
+
+func TestOpenDB(t *testing.T) {
+	dir, err := ioutil.TempDir("", "badger-test")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	ops := getTestOptions(dir)
+	ops.ReadOnly = false
+	db, err := Open(ops)
```
jarifibrahim, line 2199: Write some data which goes to sst and vlog file.

value_test.go
```diff
@@ -663,7 +663,7 @@ func TestReadOnlyOpenWithPartialAppendToWAL(t *testing.T) {
 		{Key: k2, Value: v2},
 	})
 	buf = buf[:offset-6]
-	require.NoError(t, ioutil.WriteFile(vlogFilePath(dir, 1), buf, 0777))
```
jarifibrahim, line 666: This test is returning error for memtable even when we have not written any data. Please fix this @NamanJain8

jarifibrahim (approved):
Test the code by Inserting 1 million records and reopening the db multiple times (without closing the db always). 

Feel free to merge once the CI and your manual tests work fine.

db.go
NamanJain8: I used this for deleting the created memtable in ReadOnly mode. This is not needed as we are not creating the memtable in upfront now in read-only mode.

db_test.go
NamanJain8: Done.

value_test.go
NamanJain8: Fixed this.

value.go
```diff
@@ -894,7 +897,7 @@ func (vlog *valueLog) getFileRLocked(vp valuePointer) (*logFile, error) {
 
 	// Check for valid offset if we are reading from writable log.
 	maxFid := vlog.maxFid
-	if vp.Fid == maxFid {
+	if vp.Fid == maxFid && !vlog.opt.ReadOnly {
```
NamanJain8, line 900: **Explanation**: In read-only mode, we need not check for the `vlog.writableOffset`.

db.go
```diff
@@ -612,17 +617,20 @@ func (db *DB) getMemTables() ([]*memTable, func()) {
 	db.RLock()
 	defer db.RUnlock()
 
-	tables := make([]*memTable, len(db.imm)+1)
```
jarifibrahim, line 615: Why do we need this change?

db.go
```diff
@@ -501,7 +506,7 @@ func (db *DB) close() (err error) {
 	// and remove them completely, while the block / memtable writer is still
 	// trying to push stuff into the memtable. This will also resolve the value
 	// offset problem: as we push into memtable, we update value offsets there.
-	if !db.mt.sl.Empty() {
+	if !db.opt.ReadOnly {
```
jarifibrahim, line 509: We might end up pushing an empty memtable. Why push it?

db.go
```diff
@@ -1351,7 +1359,9 @@ func (db *DB) KeySplits(prefix []byte) []string {
 		for _, mt := range memTables {
 			mtSplits(mt)
 		}
-		mtSplits(db.mt)
+		if !db.opt.ReadOnly {
+			mtSplits(db.mt)
```
jarifibrahim, line 1363: Add a `if nil return` check in `mtsplit`. It looks much cleaner.

db_test.go
```diff
@@ -2188,3 +2188,81 @@ func TestUpdateMaxCost(t *testing.T) {
...
+func TestOpenDBReadOnly(t *testing.T) {
+	dir, err := ioutil.TempDir("", "badger-test")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	mp := make(map[string][]byte)
+
+	ops := getTestOptions(dir)
+	ops.ReadOnly = false
+	db, err := Open(ops)
```
jarifibrahim, line 2201: Check err

db_test.go
```diff
@@ -2188,3 +2188,81 @@ func TestUpdateMaxCost(t *testing.T) {
...
+		val := make([]byte, db.opt.ValueThreshold+10)
+		rand.Read(val)
+		for i := 0; i < 10; i++ {
+			key := fmt.Sprintf("KEY-%05d", i)
+			require.NoError(t, txn.Set([]byte(key), val))
+			mp[key] = val
+		}
+		return nil
+	}))
+	require.NoError(t, err)
```
jarifibrahim, line 2214: This should be one line above.

value.go
jarifibrahim: Why is this the case? the writable offset should be greater than the read offset always.

db.go
```diff
@@ -298,9 +298,11 @@ func Open(opt Options) (*DB, error) {
 		return nil, y.Wrapf(err, "while opening memtables")
 	}
 
-	db.mt, err = db.newMemTable()
-	if err != nil {
-		return nil, y.Wrapf(err, "cannot create memtable")
+	if !db.opt.ReadOnly {
+		db.mt, err = db.newMemTable()
+		if err != nil {
```
jarifibrahim, line 303: Single line please. `if db.mt, err = db.newMemTable(); err != nil `

jarifibrahim (commented):
Got some comments

db.go
NamanJain8: `handleFlushTask` do nothing on empty memtable. Also, DecrRef() is called hence deleting the empty memtable. So, I am avoiding special handling here.

db.go
NamanJain8: In readOnly mode, we should not append the mutable memtable in the tables list. The `+1` is for the `db.mt`.
These `tables` get iterated upon and will cause panic on nil tables. `nil` check at various places can be cumbersome as well as untidy. 
So, either we make this change, or create a slice of size `len(db.imm)` in readOnly mode.

db.go
NamanJain8: thanks. done.

db.go
NamanJain8: done.

value.go
NamanJain8: Yes, it should be. We would not need this change if we update the writable offset in read-only mode as well by iterating over last vlog file.
I was thinking along the lines that this check is for the vlog file that is currently in use(i.e., they may be written to). As we are not checking the validity of offset for the files other than that with the `maxFid`.

db.go
jarifibrahim: okay

value.go
```diff
@@ -561,7 +561,7 @@ func (vlog *valueLog) open(db *DB) error {
 		return err
 	}
 	// If no files are found, then create a new file.
-	if len(vlog.filesMap) == 0 {
+	if len(vlog.filesMap) == 0 && !vlog.opt.ReadOnly {
```
jarifibrahim, line 564: This wouldn't work if the db is opened for the first time in read only mode.

jarifibrahim (approved):

db.go
NamanJain8: done

PR #1572, opened 2020-10-20
feat(tests): Run tests in parallel

rahulgurnani:
Run tests in parallel. Break up the tests into 3 functions: manual, packages and root. Run those via GNU parallel. Moved some expensive tests to manual mode.

With these changes, the badger tests can now run in under 6 minutes on a fast machine.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1572)

### DIFF
--- a/backup_test.go
+++ b/backup_test.go
@@ -310,14 +310,14 @@ func TestBackup(t *testing.T) {
 
 		defer removeDir(tmpdir)
 		opt := DefaultOptions(filepath.Join(tmpdir, "backup0"))
-		runBadgerTestParallel(t, &opt, func(t *testing.T, db *DB) {
+		runBadgerTest(t, &opt, func(t *testing.T, db *DB) {
 			test(t, db)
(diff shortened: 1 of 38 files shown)


backup_test.go
```diff
@@ -310,14 +310,14 @@ func TestBackup(t *testing.T) {
 
 		defer removeDir(tmpdir)
 		opt := DefaultOptions(filepath.Join(tmpdir, "backup0"))
-		runBadgerTest(t, &opt, func(t *testing.T, db *DB) {
+		runBadgerTestParallel(t, &opt, func(t *testing.T, db *DB) {
```
jarifibrahim, line 313: Why change `runBadgerTest` to `runBadgerTestParallel`? We could run all tests in parallel, always.

rahulgurnani: Some tests need to be run sequentially TestConflict, TestGoroutineLeak

PR #1582, opened 2020-11-05
fix(subcompaction): Close builder before throttle.Done

jarifibrahim:
There was a race condition in the subcompaction code which was caused
because we were doing builder.Close after doing throttle.Done in
subcomapction. This PR fixes that race condition by calling builder.Done
before calling throttle.Done .

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1582)


NamanJain8 (approved):
Great. That's why `throttle.Finish()` was not blocking and `z.Done()` being called caused the race.

ahsanbarkati (approved):

PR #1587, opened 2020-11-06
fix(compaction): Don't drop data when split overlaps with top tables

jarifibrahim:
Badger compactions are split into sub compactions if there are more than 3 tables
bottom tables are involved. The current implementation would drop keys if the top
table would overlap with a bottom table. Consider the following scenario
```
Top table => [C (deleted)]
bot table => [ A ] [ B ] [ C (valid) ] [ D ] 
```
When this range is compacted, the result was `[ A B C(valid) D ] ` 
(only `C(deleted)` was dropped). The expected result was `[A B D]` .

This PR fixes the above-mentioned issue. The test in this PR fails on master
but works with this change.

 
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1587)

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -780,8 +780,8 @@ func (s *levelsController) subcompact(it y.Iterator, kr keyRange, cd compactDef,
 		}
 		go func(builder *table.Builder) {
 			var err error
-			defer builder.Close()
 			defer inflightBuilders.Done(err)
+			defer builder.Close()
 
 			build := func(fileID uint64) (*table.Table, error) {
 				fname := table.NewFilename(fileID, s.kv.opt.Dir)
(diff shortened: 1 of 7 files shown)


ahsanbarkati (approved):
LGTM

levels.go
```diff
@@ -1001,7 +1001,15 @@ func (s *levelsController) addSplits(cd *compactDef) {
...
 		if i%N == N-1 {
-			addRange(t.Biggest())
+			// Right should always have ts=maxUint64 otherwise we'll lose keys
+			// in subcompaction. Consider the following.
+			// Top table is [A1...C3(deleted)]
+			// bot table is [B1....C2]
+			// This will generate splits like [A1 ... C2] . Notice that we
+			// dropped the C3 which is the last key of the top table.
+			// See TestCompaction/with_split test.
+			right := y.KeyWithTs(y.ParseKey(t.Biggest()), math.MaxUint64)
```
manishrjain, line 1011: getKeyRange takes care of these things. I thought I was using it -- that's the reason for the bug.

In fact, don't we want to consider the right most key? In that case, the right should be using 0 as the version.

jarifibrahim: We're generating key ranges using the `bot` tables so they don't have the correct key version.
>In fact, don't we want to consider the right most key? In that case, the right should be using 0 as the version.

No. We skip the right keys of the key range. 
https://github.com/dgraph-io/badger/blob/6c074551f8f5a5b470357236cd90e860d0deb517/levels.go#L666-L668

PR #1590, opened 2020-11-10
fix(skiplist): Remove z.buffer from skiplist

ahsanbarkati:
This PR removes z.buffer from the skiplist. Now, we use mmap file for the skiplist in `dgraph/xidmap` and we use a buffer from calloc for skiplist in badger. The mmap file can grow, but if we grow it then the skiplist must be run in serial. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1590)


skl/arena.go
```diff
@@ -35,11 +36,27 @@ const (
...
-	return int64(s.LenNoPadding())
+	return int64(atomic.LoadUint32(&s.offset))
+}
+
+func (s *Arena) allocate(sz uint32) uint32 {
+	// limit is hit, need to grow.
+	if atomic.LoadUint32(&s.offset)+sz > uint32(cap(s.data)-1) {
+		if s.Grow == nil {
+			log.Fatalf("Arena too small, toWrite:%d newTotal:%d limit:%d",
+				sz, len(s.data)+int(sz), cap(s.data))
```
jarifibrahim, line 54: this should be `s.offset`

skl/arena.go
```diff
@@ -35,11 +36,27 @@ const (
...
+		}
+		s.data = s.Grow(sz)
+		y.AssertTrue(uint32(cap(s.data))-1 > atomic.LoadUint32(&s.offset)+sz)
```
jarifibrahim, line 57: len(s.data) > offset+sz

skl/skl.go
```diff
@@ -100,23 +100,20 @@ func (s *Skiplist) DecrRef() {
...
 }
 
-func newNode(arena *Arena, key []byte, u uint64, height int) *node {
+func newNode(arena *Arena, key []byte, val uint64, height int) *node {
 	// The base level is already allocated in the node struct.
-	offset := arena.putNode(height)
-	node := arena.getNode(offset)
-	node.keyOffset = arena.putKey(key)
+	nodeOffset := arena.putNode(height)
+	keyOffset := arena.putKey(key)
```
jarifibrahim, line 111: Add a comment about why `putKey` should happen before getNode.

skl/skl.go
```diff
@@ -132,21 +129,25 @@ func decodeValue(value uint64) (valOffset uint32, valSize uint32) {
...
+	s.OnClose = func() {
+		z.Free(s.arena.data)
+	}
+	return s
 }
 
-// NewSkiplistWithBuffer makes a new skiplist, with a given buffer.
-func NewSkiplistWithBuffer(buf *z.Buffer, hasVersions bool) *Skiplist {
+// NewSkiplistWith makes a new skiplist, with a given byte slice.
+func NewSkiplistWith(buf []byte, hasVersions bool, grow func(uint32) []byte) *Skiplist {
```
jarifibrahim, line 141: Don't pass the grow function. Set it explicitely.

skl/skl_test.go
```diff
@@ -128,6 +130,45 @@ func TestBasic(t *testing.T) {
...
+func TestGrow(t *testing.T) {
+	fd, err := ioutil.TempFile("", "skiplist")
+	y.Check(err)
+	mf, err := z.OpenMmapFileUsing(fd, 1<<5, true)
+	if err != z.NewFile {
+		y.Check(err)
+	}
+
+	grow := func(sz uint32) []byte {
+		var newSz int
```
jarifibrahim, line 143: Add a bool `grown` and verify later if this was actually executed.

jarifibrahim (approved):

skl/skl.go
ahsanbarkati: We cannot do this because in that case our code would look like this:
```
s := NewSkiplistWith(buf, true)
s.SetGrowFunc(grow)
```
but the function `NewSkiplistWith()` initializes the head node and if we define the grow function after calling `NewSkiplistWith()`, then initializing the head function can fail if the initial buffer is not big enough.

ahsanbarkati (commented):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1590)* status: 0 of 3 files reviewed, 8 unresolved discussions (waiting on @ahsanbarkati, @jarifibrahim, and @manishrjain)

___
*[skl/arena.go, line 54 at r3](https://reviewable.io/reviews/dgraph-io/badger/1590#-MLmWd3bnOVk2P9OBeOX-r3-54:-MLwXu3J03QbjUrd4PD0:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/b9be3762ee6625c365fe6011a5fd6fd730f8f6cc/skl/arena.go#L54)):*
<details><summary><i>Previously, jarifibrahim (Ibrahim Jarif) wrote…</i></summary><blockquote>

this should be `s.offset`
</blockquote></details>

Done.
___
*[skl/arena.go, line 57 at r3](https://reviewable.io/reviews/dgraph-io/badger/1590#-MLmWd3bnOVk2P9OBeOX-r3-57:-MLwXvRm8XB5tO2sY5IS:b-896fix) ([raw file](https://github.com/dgraph-io/badger/blob/b9be3762ee6625c365fe6011a5fd6fd730f8f6cc/skl/arena.go#L57)):*
<details><summary><i>Previously, jarifibrahim (Ibrahim Jarif) wrote…</i></summary><blockquote>

len(s.data) > offset+sz
</blockquote></details>

Done.

manishrjain (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1590)* status: 0 of 3 files reviewed, 9 unresolved discussions (waiting on @ahsanbarkati, @jarifibrahim, and @manishrjain)

___
*[skl/arena.go, line 51 at r4](https://reviewable.io/reviews/dgraph-io/badger/1590#-MLxfird7sICOWxy1OJ_:-MLxfird7sICOWxy1OJa:b71mws8) ([raw file](https://github.com/dgraph-io/badger/blob/3e5c770aea7f8507d6b027030a0a4f68bde2586e/skl/arena.go#L51)):*
> ```Go
> func (s *Arena) allocate(sz uint32) uint32 {
> 	for {
> 		offset := atomic.LoadUint32(&s.offset)
> ```

offset := atomic.AddUint32(&s.offset, sz)

if offset < len(s.data) {
  return s.data[offset-sz:offset]
}

if s.grow == nil {
  fatal
}

s.grow()

return s.data[offset-sz:offset]

ahsanbarkati:
Closing this PR because we will be using b+ tree instead of skiplist in dgraph's xidmap.

PR #1591, opened 2020-11-11
fix(stream): Fix memory leak from allocators

jarifibrahim:
There were two memory leaks in the stream framework and this PR fixes both of them.
The reuse of allocators was causing the memory leak.

Fixes DGRAPH-2650

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1591)

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -309,6 +309,13 @@ func (st *Stream) streamKVs(ctx context.Context) error {
 	defer t.Stop()
 	now := time.Now()
 
+	var allocs []*z.Allocator
+	defer func() {
+		for _, a := range allocs {
+			a.Release()
+		}
+	}()
+
 	sendBatch := func(batch *pb.KVList) error {
 		sz := uint64(proto.Size(batch))
 		bytesSent += sz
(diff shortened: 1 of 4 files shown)


stream.go
```diff
@@ -206,6 +201,8 @@ func (st *Stream) produceKVs(ctx context.Context, threadId int) error {
 	}
 	defer txn.Discard()
 
+	outList := new(pb.KVList)
+	outList.AllocRef = st.newAllocator(threadId).Ref
```
NamanJain8, line 205: Why do we want to move it outside iterate? I see no difference with case when if it is inside `iterate`.

NamanJain8 (approved):
Looks good. Just a minor comment.

stream.go
NamanJain8: Also, do we want a test where `KeyToList()` does not use the allocator?

jarifibrahim: The outlist was inside `iterate` function and `sendit` creates a new outlist.
So if the following sequence of calls happen
```
iterate <- New outlist created (say o1)
sendIt <- Old outlist (o1) pushed to kvchan and new outlist (say o2) created 
iterate <- new outlist (o3 created)
```
Notice that we did not push 02 to the `kvChan` and so the allocator in `o2` will leak.
Moving the outlist outside prevents this problem.

jarifibrahim: I'll push a test for this.

manishrjain (commented):
Is this PR still relevant?

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1591#-:-MMqYjRv6kr253fDKtxE:b-5bnhf)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @manishrjain and @NamanJain8)

PR #1600, opened 2020-11-19
fix(skiplist): Remove z.Buffer from skiplist

ahsanbarkati:
Skiplist uses a static sized buffer so z.buffer is not needed for it.


go.sum
```diff
@@ -36,72 +39,102 @@ github.com/google/flatbuffers v1.12.0 h1:/PtAHvnBY4Kqnx/xCQ3OIV9uYcSFGScBsWI3Oog
...
 github.com/stretchr/testify v1.4.0 h1:2E4SXV/wtOkTonXsotYi4li6zVWxYlZuYNCXe9XRJyk=
 github.com/stretchr/testify v1.4.0/go.mod h1:j7eGeouHqKxXV5pUuKE4zz7dFj8WfuZ+81PSLYec5m4=
+github.com/twitchyliquid64/golang-asm v0.15.0/go.mod h1:a1lVb/DtPvCB8fslRZhAngC2+aY1QWCk3Cedj/Gdt08=
 github.com/ugorji/go/codec v0.0.0-20181204163529-d75b2dcb6bc8/go.mod h1:VFNgLljTbGfSG7qAOspJ7OScBnGdDN/yBr0sguwnwf0=
 github.com/xordataexchange/crypt v0.0.3-0.20170626215501-b2862e3d0a77/go.mod h1:aYKd//L2LvnjZzWKhF00oedf4jCCReLcmhLdhm1A27Q=
+github.com/yuin/goldmark v1.2.1/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 go.opencensus.io v0.22.5 h1:dntmOdLpSpHlVqbW5Eay97DelsZHe+55D+xC6i0dDS0=
 go.opencensus.io v0.22.5/go.mod h1:5pWMHQbX5EPX2/62yrJeAkowc+lfs/XD7Uxpq3pI6kk=
+go.starlark.net v0.0.0-20190702223751-32f345186213/go.mod h1:c1/X6cHgvdXj6pUlmWKMkuqRnW4K8x2vwt6JAaaircg=
+golang.org/x/arch v0.0.0-20190927153633-4e8777c89be4/go.mod h1:flIaEI6LNU6xOCD5PaJvn9wGP0agmIOqjrtsKGRguv4=
```
jarifibrahim, line 91: Run `go mod tidy`.

skl/arena.go
```diff
@@ -35,11 +36,22 @@ const (
...
 func (s *Arena) size() int64 {
-	return int64(s.LenNoPadding())
+	return int64(atomic.LoadUint32(&s.offset))
+}
+
+func (s *Arena) allocate(sz uint32) uint32 {
+	offset := atomic.AddUint32(&s.offset, sz)
+	if offset >= uint32(len(s.data)) {
+		log.Fatalf("Arena too small, toWrite:%d newTotal:%d limit:%d",
+			sz, offset+sz, len(s.data))
```
jarifibrahim, line 52: new total should be just `offset` since you've already added it on line 49.

skl/skl.go
```diff
@@ -132,16 +134,18 @@ func decodeValue(value uint64) (valOffset uint32, valSize uint32) {
 
 // NewSkiplist makes a new empty skiplist, with a given arena size.
 func NewSkiplist(arenaSize int64) *Skiplist {
-	buf, err := z.NewBufferWith(int(arenaSize), int(arenaSize), z.UseCalloc)
-	y.Check(err)
-	skl := NewSkiplistWithBuffer(buf, true)
-	return skl
+	buf := z.Calloc(int(arenaSize))
+	s := NewSkiplistWith(buf, true)
+	s.arena.Release = func() {
```
jarifibrahim, line 139: How would someone set this `s.arena.Release()`? The `arena` is not exposed. You might want to move the `Release` method to the skiplist.

jarifibrahim (changes requested):
Got some comments.

skl/arena.go
ahsanbarkati: Done

skl/skl.go
ahsanbarkati: Thanks. Fixed.

skl/skl.go
```diff
@@ -84,6 +84,7 @@ type Skiplist struct {
 	hasVersions bool
 	comparator  comparatorFunc
 	OnClose     func()
+	Release     func()
```
NamanJain8, line 87: I think Release can be a part of OnClose itself, as we would be releasing the memory only on close.

jarifibrahim (approved):

skl/skl.go
jarifibrahim: In case of skiplist in badger, that would mean you have two functions to run.
First one will close and delete the `.mem` file and the second one will free up the memory.

We can either change the onClose to be `[]func()` and then append both the functions or keep the `onRelease` as it is.

jarifibrahim: @ahsanbarkati This should be called `onRelease`.

ahsanbarkati: OnClose is a general purpose function. Memtable uses it as a callback to delete WAL when skiplist is closed. I think we should have a release function.

ahsanbarkati: @jarifibrahim I saw the comments after merging the PR, may be because we commented at the same time. 

By `onRelease` do you mean that we should not expose it? or do you mean `OnRelease`?

PR #1618, opened 2020-12-10
[Do Not Merge] fix(compaction): Don't drop keys if there is an overlap

jarifibrahim:
This commit fixes the `Unable to read Key: xxx error: file with ID yyy not
found` issue.

The issue occurs because we drop intermediate keys (the moved ones)
during compaction. If key X is moved because of GC and there is a
version X' with a higher version and bitDiscardEarlierVersions set, we
might end up dropping the moved key X and the original key X (which
points to an invalid file) will show up again.

This change also has a side effect. The LSM Tree would be cleaned up
only when all the compaction doesn't have any overlap with the lower
levels. This could also mean that keys are dropped from the LSM tree
only when they reach the bottommost level.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1618)

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -660,7 +660,12 @@ func (s *levelsController) subcompact(it y.Iterator, kr keyRange, cd compactDef,
 
 			// See if we need to skip this key.
 			if len(skipKey) > 0 {
-				if y.SameKey(it.Key(), skipKey) {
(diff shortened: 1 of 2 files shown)


jarifibrahim:
Closing this PR since we no longer need it.

PR #1619, opened 2020-12-10
fix(GC): Set bits correctly for moved keys

jarifibrahim:
When keys are moved because of the GC, we were removing the
bitDiscardEarlierVersions and other bits. Only the transaction markers
should be removed and all the other bits should be kept.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1619)


NamanJain8 (approved):

jarifibrahim:
Approved by @manishrjain

PR #1621, opened 2020-12-17
opt(DropPrefix): check if there exist some data to drop before dropping prefixes

NamanJain8:
Fixes DGRAPH-2853

We currently do not check if there exist some data to drop for prefixes in [`DropPrefix()`](https://github.com/dgraph-io/badger/blob/bcfae6104545e824b40814e4bc3bc4b1430c3367/db.go#L1631-L1677). This can make drop prefix to be expensive because it has to do a bunch of compactions, etc.

This PR optimizes that behavior by not doing any of that in case no data exist for prefixes.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1621)

### DIFF
--- a/db.go
+++ b/db.go
@@ -1638,6 +1638,16 @@ func (db *DB) DropPrefix(prefixes ...[]byte) error {
 		return err
 	}
 	defer f()
+
+	var filtered [][]byte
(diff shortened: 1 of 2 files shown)


ahsanbarkati (approved):
LGTM

jarifibrahim (approved):
Add a test please.

PR #1622, opened 2020-12-17
(cherry-pick) (release/v2.2007): fix(dropPrefix): check properly if the key is present in a table

NamanJain8:
There may be a case when the table contains `[0x0000,...., 0xffff]`. If we are searching for `k=0x0011`, we should not directly infer that `k` is present. It may not be present. This leads to the triggering of compaction at the time of `DropPrefix` which is quite expensive.

This PR fixes the issue by confirming that case by searching inside the table for the prefix.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1622)

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -329,7 +329,7 @@ func (s *levelsController) dropPrefixes(prefixes [][]byte) error {
 		}
 
 		for _, table := range l.tables {
-			if containsAnyPrefixes(table.Smallest(), table.Biggest(), prefixes) {
+			if containsAnyPrefixes(table, prefixes) {
 				tableGroup = append(tableGroup, table)
 			} else {
 				finishGroup()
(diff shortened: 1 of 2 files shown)


levels.go
```diff
@@ -753,24 +753,46 @@ func hasAnyPrefixes(s []byte, listOfPrefixes [][]byte) bool {
...
 		return true
 	}
+	isPresent := func() bool {
+		ti := table.NewIterator(false)
+		defer ti.Close()
+		// In table iterator's Seek, we assume that key has timestamp in last 8 bytes. We set ts=0,
+		// so that we don't skip the key prefixed with prefix.
+		prefixWithTs := make([]byte, len(prefix)+8)
+		copy(prefixWithTs[:len(prefix)], prefix)
+		prefixWithTs = prefixWithTs[:cap(prefixWithTs)]
```
jarifibrahim, line 772: Maybe you don't need this. Also, the cap could be much bigger than the actual length.

levels.go
```diff
@@ -753,24 +753,46 @@ func hasAnyPrefixes(s []byte, listOfPrefixes [][]byte) bool {
 	return false
 }
 
-func containsPrefix(smallValue, largeValue, prefix []byte) bool {
+func containsPrefix(table *table.Table, prefix []byte) bool {
```
jarifibrahim, line 756: Add a couple of tests.
Build a table with some prefixes and then run contains prefix on it.

Make sure there is a test for prefix (with versions) which lies at the end of the table.

jarifibrahim (commented):
Please add some tests

levels.go
NamanJain8: Done

levels.go
NamanJain8: Done.

NamanJain8:
Closing this as this is not required in v2.2007. If needed we can cherry-pick it back.

PR #1623, opened 2020-12-21
fix(dropPrefix): check properly if the key is present in a table 

NamanJain8:
There may be a case when the table contains [0x0000,...., 0xffff]. If we are searching for `k=0x0011`, we should not directly infer that k is present. It may not be present. This leads to the triggering of compaction at the time of DropPrefix which is quite expensive.

This PR fixes the issue by confirming that case by searching inside the table for the prefix.
Fix for v2.2007 in #1622 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1623)

### DIFF
--- a/levels_test.go
+++ b/levels_test.go
@@ -1088,5 +1088,6 @@ func TestTableContainsPrefix(t *testing.T) {
 
 	require.False(t, containsPrefix(tbl, []byte("key0")))
 	require.False(t, containsPrefix(tbl, []byte("key2")))
+	require.False(t, containsPrefix(tbl, []byte("key323")))
 	require.False(t, containsPrefix(tbl, []byte("key5")))
 }
--- a/levels_test.go
+++ b/levels_test.go
@@ -1055,15 +1055,12 @@ func TestKeyVersions(t *testing.T) {
 
 func TestTableContainsPrefix(t *testing.T) {
 	opts := table.Options{
-		LoadingMode:        options.LoadToRAM,
 		BlockSize:          4 * 1024,
-		BloomFalsePositive: 0.001,
+		BloomFalsePositive: 0.01,
 	}
 
-	buildTable := func(keys []string) *os.File {
+	buildTable := func(keys []string) *table.Table {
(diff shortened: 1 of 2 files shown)


levels_test.go
```diff
@@ -1049,3 +1052,41 @@ func TestKeyVersions(t *testing.T) {
...
 }
+
+func TestTableContainsPrefix(t *testing.T) {
+	opts := table.Options{
+		BlockSize:          4 * 1024,
+		BloomFalsePositive: 0.01,
+	}
+
+	buildTable := func(keys []string) *table.Table {
+		filename := fmt.Sprintf("%s%s%d.sst", os.TempDir(), string(os.PathSeparator), rand.Uint32())
```
ahsanbarkati, line 1063: Can we use `ioutil.TempFile()`?. It makes sure that it creates a non-conflicting file.

ahsanbarkati (approved):

levels_test.go
```diff
@@ -1049,3 +1052,41 @@ func TestKeyVersions(t *testing.T) {
...
+	}
+
+	tbl := buildTable([]string{"key1", "key3", "key31", "key32", "key4"})
+	defer tbl.DecrRef()
+
+	require.True(t, containsPrefix(tbl, []byte("key")))
+	require.True(t, containsPrefix(tbl, []byte("key1")))
+	require.True(t, containsPrefix(tbl, []byte("key3")))
+	require.True(t, containsPrefix(tbl, []byte("key32")))
+	require.True(t, containsPrefix(tbl, []byte("key4")))
```
jarifibrahim, line 1087: Add
```go
	require.False(t, containsPrefix(tbl, []byte("key323")))

```

jarifibrahim (approved):

levels_test.go
NamanJain8: That can be done. But that would require more handling of the rest of the logic like `CreateTable`, etc. For simplicity, keeping it the same.

levels_test.go
NamanJain8: Done.

PR #1624, opened 2020-12-21
fix(badger): Do not reuse variable across badger commands

jarifibrahim:
The same variables were being across multiple badger commands. The
default value of a flag that uses the same variable across two commands
could be from one of the commands and it isn't clear which default value
would be used.

The `numVersions` command was being used in flatten (default value 1)
and stream (default value 0). Running `badger stream --help` would show
the default value to be `0` but if you print the value of the
`numVersions` variable, it would turn out to be `1`. This is misleading.

This PR separates the variable so that they don't overlap.

This PR also adds the `compression` flag to the flatten tool.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1624)

### DIFF
--- a/badger/cmd/flatten.go
+++ b/badger/cmd/flatten.go
@@ -21,6 +21,8 @@ import (
 	"math"
 
 	"github.com/dgraph-io/badger/v2"
+	"github.com/dgraph-io/badger/v2/options"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
@@ -34,9 +36,10 @@ This command would compact all the LSM tables into one level.
 }
 
 var fo = struct {
-	keyPath     string
-	numWorkers  int
-	numVersions int
+	keyPath         string
+	numWorkers      int
+	numVersions     int
+	compressionType uint32
 }{}
 
 func init() {
@@ -49,6 +52,9 @@ func init() {
 			"Values <= 0 will be considered to have the max number of versions.")
 	flattenCmd.Flags().StringVar(&fo.keyPath, "encryption-key-file", "",
 		"Path of the encryption key file.")
+	streamCmd.Flags().Uint32VarP(&fo.compressionType, "compression", "", 1,
+		"Option to configure the compression type in output DB. "+
+			"0 to disable, 1 for Snappy, and 2 for ZSTD.")
 }
 
 func flatten(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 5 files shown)


badger/cmd/read_bench.go
```diff
@@ -64,18 +65,15 @@ func init() {
...
-	readBenchCmd.Flags().StringVar(
-		&loadingMode, "loading-mode", "mmap", "Mode for accessing SSTables and value log files. "+
-			"Valid loading modes are fileio and mmap.")
+		&ro.readOnly, "read-only", true, "If true, DB will be opened in read only mode.")
 	readBenchCmd.Flags().BoolVar(
-		&fullScan, "full-scan", false, "If true, full db will be scanned using iterators.")
-	readBenchCmd.Flags().Int64Var(&blockCacheSize, "block-cache", 0, "Max size of block cache in MB")
-	readBenchCmd.Flags().Int64Var(&indexCacheSize, "index-cache", 0, "Max size of index cache in MB")
+		&ro.fullScan, "full-scan", false, "If true, full db will be scanned using iterators.")
+	readBenchCmd.Flags().Int64Var(&ro.blockCacheSize, "block-cache", 1024, "Max size of block cache in MB")
```
NamanJain8, line 75: Shouldn't this be set to default value `256MB`?

badger/cmd/flatten.go
```diff
@@ -33,36 +35,48 @@ This command would compact all the LSM tables into one level.
...
 		"Number of concurrent compactors to run. More compactors would use more"+
 			" server resources to potentially achieve faster compactions.")
-	flattenCmd.Flags().IntVarP(&numVersions, "num_versions", "", 1,
+	flattenCmd.Flags().IntVarP(&fo.numVersions, "num_versions", "", 0,
 		"Option to configure the maximum number of versions per key. "+
 			"Values <= 0 will be considered to have the max number of versions.")
-	flattenCmd.Flags().StringVar(&keyPath, "encryption-key-file", "",
+	flattenCmd.Flags().StringVar(&fo.keyPath, "encryption-key-file", "",
 		"Path of the encryption key file.")
+	streamCmd.Flags().Uint32VarP(&fo.compressionType, "compression", "", 1,
```
NamanJain8, line 55: typo `flattenCmd`

NamanJain8 (changes requested):

badger/cmd/flatten.go
```diff
@@ -33,36 +35,48 @@ This command would compact all the LSM tables into one level.
...
+}{}
 
 func init() {
 	RootCmd.AddCommand(flattenCmd)
-	flattenCmd.Flags().IntVarP(&numWorkers, "num-workers", "w", 1,
+	flattenCmd.Flags().IntVarP(&fo.numWorkers, "num-workers", "w", 1,
 		"Number of concurrent compactors to run. More compactors would use more"+
 			" server resources to potentially achieve faster compactions.")
-	flattenCmd.Flags().IntVarP(&numVersions, "num_versions", "", 1,
+	flattenCmd.Flags().IntVarP(&fo.numVersions, "num_versions", "", 0,
```
ahsanbarkati, line 50: For some options we are using snake case and for some we are using kebab case. Can we make it consistent?

jarifibrahim: We cannot change them now. That will be a breaking change.

badger/cmd/flatten.go
jarifibrahim: Nice catch. Fixed it.

badger/cmd/read_bench.go
jarifibrahim: Sure. Fixed it.

NamanJain8: There is one more in the write bench. I missed publishing that.

NamanJain8 (approved):

ahsanbarkati (approved):

badger/cmd/read_bench.go
jarifibrahim: fixed

PR #1638, opened 2021-01-12
feat: add functionality to ban a prefix

NamanJain8:
Fixes DGRAPH-2912

This PR adds functionality to ban a prefix so that no read/write is allowed to a key prefixed with any of the banned prefixes:

- Add BanPrefix() and GetBannedPrefix() API.
- Set(), Get() operations on the keys prefixed with any of these banned prefixes is denied.
- The iterator will skip these prefixes.
- These banned prefixes will be dropped at the time of compaction as the iterator will not even pick them.

**Issues:**
This will make iterator and transaction Set/Get slower as we need to check through the banned prefixes for these operations.

**Future Optimisations:** (can be a part of separate PR) [Some of this is not valid now]
- We can do some sort of prioritized compaction where the tables with banned prefixes are compacted first. Some sort of asynchronous dropPrefix will also help.
- If the above is done, we can keep 2 lists, one is wholesome(containing all the banned prefixes) which will be used at the time of Get/Set. Another shorter one contains the list of active banned prefixes (which have been banned and we are not sure if they have completely been removed from the tree). This list can be used for the iterator.
- Another optimization is related to how we pick tables and modifications to block iterator.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1638)


db.go
```diff
@@ -795,11 +832,24 @@ func (db *DB) writeRequests(reqs []*request) error {
...
+		}
+	}
+	return nil
+}
 
 func (db *DB) sendToWriteCh(entries []*Entry) (*request, error) {
 	if atomic.LoadInt32(&db.blockWrites) == 1 {
 		return nil, ErrBlockedWrites
 	}
+	if err := db.validateEntries(entries); err != nil {
```
jarifibrahim, line 850: You should validate the entries before accepting them. If you return error from here, the user wouldn't know which prefix is banned. Instead, return the error from `txn.Modify` function.

db_test.go
```diff
@@ -2287,3 +2287,77 @@ func TestOpenDBReadOnly(t *testing.T) {
...
+	ops.ValueThreshold = 0
+	ops.ValueLogMaxEntries = 2
+	db, err := Open(ops)
+	require.NoError(t, err)
+	require.Equal(t, 1, len(db.vlog.filesMap))
+
+	var keys [10][]byte
+	var bannedPrefixes [][]byte
+	isBanned := func(key []byte) bool {
+		for _, prefix := range bannedPrefixes {
```
jarifibrahim, line 2308: Don't store the bannedPrefixes in the test. Fetch it using `db.GetBannedPrefixes`

jarifibrahim (commented):
Got some comments

stream.go
```diff
@@ -237,6 +237,11 @@ func (st *Stream) produceKVs(ctx context.Context, threadId int) error {
 				continue
 			}
 
+			// Do not send the keys which are banned.
+			if st.db.isBanned(item.Key()) {
```
jarifibrahim, line 241: Maybe move this above choose key.

jarifibrahim (changes requested):
This PR is missing the changes for iterators

db.go
```diff
@@ -360,6 +372,31 @@ func Open(opt Options) (*DB, error) {
...
+	vs, err := db.get(prefixKey)
+	if err != nil {
+		return err
+	}
+	encoded := vs.Value
+	if vs.Meta&bitValuePointer > 0 {
+		var vptr valuePointer
+		vptr.Decode(vs.Value)
+		buf, cb, err := db.vlog.Read(vptr, nil)
+		cb()
```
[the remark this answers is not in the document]
NamanJain8: Done

db.go
NamanJain8: Done. Thanks.

db_test.go
NamanJain8: Added a separate check to comparision with `db.GetBannedPrefixes()`. The local list acts as the source of truth.

stream.go
NamanJain8: This is no longer needed with the iterator changes.

iterator.go
```diff
@@ -602,6 +613,26 @@ func (it *Iterator) parseItem() bool {
...
+			// Avoid calling isBanned() by checking with the found prefix.
+			mi.Next()
+			for mi.Valid() {
+				key = mi.Key()
+				if !bytes.HasPrefix(y.ParseKey(key), prefix) {
+					break
+				}
+				mi.Next()
+			}
+			continue
```
jarifibrahim, line 629: You can remove this

iterator.go
```diff
@@ -602,6 +613,26 @@ func (it *Iterator) parseItem() bool {
...
+		key = mi.Key()
+		if prefix := isBanned(y.ParseKey(key)); prefix != nil {
+			// Avoid calling isBanned() by checking with the found prefix.
+			mi.Next()
+			for mi.Valid() {
+				key = mi.Key()
+				if !bytes.HasPrefix(y.ParseKey(key), prefix) {
+					break
+				}
+				mi.Next()
```
jarifibrahim, line 627: Move this into for loop.

iterator.go
```diff
@@ -602,6 +613,26 @@ func (it *Iterator) parseItem() bool {
 		}
 	}
 
+	// Skip keys with banned prefixes.
+	for mi.Valid() {
```
jarifibrahim, line 617: You should check this before the all versions check below.

jarifibrahim (commented):
Got some comments.

db.go
```diff
@@ -360,6 +372,31 @@ func Open(opt Options) (*DB, error) {
...
+	prefixKey := y.KeyWithTs(bannedPrefixKey, math.MaxUint64)
+	vs, err := db.get(prefixKey)
+	if err != nil {
+		return err
+	}
+	encoded := vs.Value
+	if vs.Meta&bitValuePointer > 0 {
+		var vptr valuePointer
+		vptr.Decode(vs.Value)
+		buf, cb, err := db.vlog.Read(vptr, nil)
```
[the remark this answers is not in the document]
NamanJain8: Done by https://github.com/dgraph-io/badger/pull/1640

iterator.go
NamanJain8: This is the correct place for skipping the banned prefixes because after skipping the banned keys, we still need to check if the keys are internal or the key's version is under is above readTs.

manishrjain (changes requested):
Perhaps make uint64 as the prefix key. Perhaps an option in Badger, which says all keys have uint64 namespace prefix -- so expectation from Dgraph is to pass that always.

The key format in Dgraph would be `<byte><len-4-bytes><namespace-8-bytes>`.  Or perhaps, `<namespace-8-bytes><byte><len-4-bytes>...`.

> Reviewed 2 of 6 files at r1, 2 of 5 files at r2, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1638#-:-MR5u8-O5F3CzUC6p33k:bt9czr2)* status: all files reviewed, 11 unresolved discussions (waiting on @jarifibrahim and @NamanJain8)

___
*[db.go, line 69 at r4](https://reviewable.io/reviews/dgraph-io/badger/1638#-MR5r1D89WUz_0flWH5j:-MR5r1D89WUz_0flWH5k:b-5h7k1y) ([raw file](https://github.com/dgraph-io/badger/blob/d60067e3b00b876f59afea68ccec856f959c747f/db.go#L69)):*
> ```Go
> 	keys [][]byte
> }
> ```

add a func to get keys, which would do internal locking. So, the caller doesn't have to do locking.
___
*[iterator.go, line 609 at r4](https://reviewable.io/reviews/dgraph-io/badger/1638#-MR5tNowE7QhhAhWIXLh:-MR5tNowE7QhhAhWIXLi:bz44a22) ([raw file](https://github.com/dgraph-io/badger/blob/d60067e3b00b876f59afea68ccec856f959c747f/iterator.go#L609)):*
> ```Go
> 
> 	isBanned := func(key []byte) []byte {
> 		for _, prefix := range it.banned {
> ```

MUST be a lookup.
___
*[iterator.go, line 627 at r4](https://reviewable.io/reviews/dgraph-io/badger/1638#-MR5tx9a9adMEkjHeS8O:-MR5tx9a9adMEkjHeS8P:b-msrg9y) ([raw file](https://github.com/dgraph-io/badger/blob/d60067e3b00b876f59afea68ccec856f959c747f/iterator.go#L627)):*
> ```Go
> 		mi.Next()
> 		// Avoid calling isBanned(). Skip all the keys prefixed with the found banned prefix.
> 		for ; mi.Valid(); mi.Next() {
> ```

let's not loop here. Just return false.

db.go
```diff
@@ -1721,6 +1760,75 @@ func (db *DB) filterPrefixesToDrop(prefixes [][]byte) ([][]byte, error) {
...
+	db.bannedPrefixes.view(func(keys map[uint64]struct{}) {
+		if _, ok := keys[key]; ok {
+			banned = true
+		}
+	})
+	return banned
+}
+
+// Returns true, if the key is prefixed with any banned prefix.
+func (db *DB) isBanned(key []byte) bool {
```
jarifibrahim, line 1786: The `isBannedHelper` can be part of this function. you don't need two functions.

jarifibrahim (commented):
Got some comments

db.go
```diff
@@ -1721,6 +1760,75 @@ func (db *DB) filterPrefixesToDrop(prefixes [][]byte) ([][]byte, error) {
...
+	lk.RLock()
+	defer lk.RUnlock()
+	fn(lk.keys)
+}
+
+func (db *DB) isBannedHelper(key uint64) bool {
+	banned := false
+	db.bannedPrefixes.view(func(keys map[uint64]struct{}) {
+		if _, ok := keys[key]; ok {
+			banned = true
```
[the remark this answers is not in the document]
NamanJain8: We are not iterating. The view function was created to take care of locking and unlocking only.

options.go
```diff
@@ -98,6 +98,9 @@ type Options struct {
 	// conflict detection is disabled.
 	DetectConflicts bool
 
+	// DgraphMode determines whether db should expect namespace in the first 8 bytes of the keys.
+	DgraphMode bool
```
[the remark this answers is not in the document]
NamanJain8: Done.

db.go
NamanJain8: Done. Thanks.

manishrjain (changes requested):
> Reviewed 2 of 8 files at r5, 2 of 4 files at r6, 4 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1638)* status: all files reviewed, 19 unresolved discussions (waiting on @jarifibrahim and @NamanJain8)

___
*[db.go, line 53 at r7](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRVSDLCDLapYulTTfm2:-MRVSDLCDLapYulTTfm3:b-py5tx) ([raw file](https://github.com/dgraph-io/badger/blob/22d9c9507247a506387007e8a69aaa1b33b20b78/db.go#L53)):*
> ```Go
> const (
> 	maxNumSplits    = 128
> 	u64PrefixOffset = 3
> ```

seems like this should be an option.
___
*[db.go, line 1764 at r7](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRVS_04E9huSRIHUZhD:-MRVS_04E9huSRIHUZhE:b-79u7co) ([raw file](https://github.com/dgraph-io/badger/blob/22d9c9507247a506387007e8a69aaa1b33b20b78/db.go#L1764)):*
> ```Go
> }
> 
> func (lk *lockedKeys) update(fn func(keys map[uint64]struct{})) {
> ```

move the funcs next to the struct.
___
*[db.go, line 1801 at r7](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRVTVXC7mbajimpuT5s:-MRVTVXC7mbajimpuT5t:b-awnevr) ([raw file](https://github.com/dgraph-io/badger/blob/22d9c9507247a506387007e8a69aaa1b33b20b78/db.go#L1801)):*
> ```Go
> 
> // BanPrefix bans a prefix. Read/write to keys prefixed with any of such prefix is denied.
> func (db *DB) BanPrefix(prefix uint64) error {
> ```

`BanNamespace(ns uint64)`
___
*[errors.go, line 82 at r7](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRVV-Y5DHW0y2Hu4gtN:-MRVV-Y5DHW0y2Hu4gtO:b-1ex4r) ([raw file](https://github.com/dgraph-io/badger/blob/22d9c9507247a506387007e8a69aaa1b33b20b78/errors.go#L82)):*
> ```Go
> 		"Invalid API request. Not allowed to perform this action using ManagedDB")
> 
> 	// ErrUint64PrefixMode is returned if the user tries to use an API which isn't allowed in non-Dgraph
> ```

100 chars.
___
*[txn.go, line 384 at r7](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRVVCUFCHpm9Tchj9cA:-MRVVCUFCHpm9Tchj9cB:b-kepui6) ([raw file](https://github.com/dgraph-io/badger/blob/22d9c9507247a506387007e8a69aaa1b33b20b78/txn.go#L384)):*
> ```Go
> 	}
> 
> 	banned, err := txn.db.isBanned(e.Key)
> ```

`if err := isBanned(e.Key); err != nil { return err }`

Don't have a bool and an error.

manishrjain (approved):
:lgtm: Got a few comments. Do address before merging. Defer to @jarifibrahim for final approval.

> Reviewed 6 of 7 files at r8.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1638#-:-MR_eMZXCSBcs_FoUXbL:b-6cshoq)* status: all files reviewed, 25 unresolved discussions (waiting on @jarifibrahim and @NamanJain8)

___
*[db.go, line 406 at r8](https://reviewable.io/reviews/dgraph-io/badger/1638#-MR_cCHt-mrGxSElBwRo:-MR_cCHt-mrGxSElBwRp:b7ngs07) ([raw file](https://github.com/dgraph-io/badger/blob/55a45c9484b8d6c47ee290ab82b21167695cf6f4/db.go#L406)):*
> ```Go
> 		return nil
> 	})
> 	db.bannedNamespaces.update(func(keys map[uint64]struct{}) {
> ```

This update system isn't needed here.

`db.bannedNamespaces.Add(ns)` in the iterator above.
___
*[db.go, line 1782 at r8](https://reviewable.io/reviews/dgraph-io/badger/1638#-MR_dvs31VNJnzWQ-yw-:-MR_dvs31VNJnzWQ-yw0:b-kuq2tu) ([raw file](https://github.com/dgraph-io/badger/blob/55a45c9484b8d6c47ee290ab82b21167695cf6f4/db.go#L1782)):*
> ```Go
> 	}
> 	if len(key) <= db.opt.NamespaceOffset+8 {
> 		return ErrInvalidKey
> ```

Just return nil here.
___
*[db.go, line 1785 at r8](https://reviewable.io/reviews/dgraph-io/badger/1638#-MR_cOr3CiTCEhtAKrt_:-MR_cOr3CiTCEhtAKrta:b-4rlli8) ([raw file](https://github.com/dgraph-io/badger/blob/55a45c9484b8d6c47ee290ab82b21167695cf6f4/db.go#L1785)):*
> ```Go
> 	}
> 	var err error
> 	db.bannedNamespaces.view(func(keys map[uint64]struct{}) {
> ```

This view is also not needed. Can just say `db.bannedNamespaces.Has(ns)`.
___
*[db.go, line 1812 at r8](https://reviewable.io/reviews/dgraph-io/badger/1638#-MR_ccOi8NxggPZ98DJK:-MR_ccOi8NxggPZ98DJL:bb6fmzc) ([raw file](https://github.com/dgraph-io/badger/blob/55a45c9484b8d6c47ee290ab82b21167695cf6f4/db.go#L1812)):*
> ```Go
> 		return err
> 	}
> 	db.bannedNamespaces.update(func(keys map[uint64]struct{}) {
> ```

.Add
___
*[db.go, line 1819 at r8](https://reviewable.io/reviews/dgraph-io/badger/1638#-MR_cfGD6X58advVB-GU:-MR_cfGD6X58advVB-GV:b-3mx8rp) ([raw file](https://github.com/dgraph-io/badger/blob/55a45c9484b8d6c47ee290ab82b21167695cf6f4/db.go#L1819)):*
> ```Go
> 
> // GetBannedPrefixes returns the list of prefixes banned for DB.
> func (db *DB) GetBannedPrefixes() []uint64 {
> ```

BannedNamespaces
___
*[iterator.go, line 607 at r8](https://reviewable.io/reviews/dgraph-io/badger/1638#-MR_czF-9je_OhXrUDKr:-MR_czF06gUZYm4xpuLD:bu7qvwc) ([raw file](https://github.com/dgraph-io/badger/blob/55a45c9484b8d6c47ee290ab82b21167695cf6f4/iterator.go#L607)):*
> ```Go
> 
> 	// Skip badger keys.
> 	if !it.opt.InternalAccess && bytes.HasPrefix(key, badgerPrefix) {
> ```

In critical path of iteration (not compactions and other background jobs, but triggered by a user), bytes.HasPrefix has already been checked. So, just assign it and use it below.
___
*[options.go, line 106 at r8](https://reviewable.io/reviews/dgraph-io/badger/1638#-MR_dQ7F5uaqmx7bhz1v:-MR_dQ7G48RLFj1Uk4fZ:b-puki4w) ([raw file](https://github.com/dgraph-io/badger/blob/55a45c9484b8d6c47ee290ab82b21167695cf6f4/options.go#L106)):*
> ```Go
> 	// NamespaceOffset specifies the offset from where the next 8 bytes contains the namespace.
> 	// Valid only when NamespaceMode is set.
> 	NamespaceOffset int
> ```

Just add only one option. NamespaceOffset can be -1 by default, indicating it's not being used.
___
*[options.go, line 597 at r8](https://reviewable.io/reviews/dgraph-io/badger/1638#-MR_eGRcCegVXJoYjunr:-MR_eGRcCegVXJoYjuns:b-p9qysn) ([raw file](https://github.com/dgraph-io/badger/blob/55a45c9484b8d6c47ee290ab82b21167695cf6f4/options.go#L597)):*
> ```Go
> // will expect the namespace in each key at the 8 bytes starting from NamespaceOffset.
> //
> // The default value for NamespaceMode is 0. This option is only valid when NamespaceMode is set.
> ```

is -1.

jarifibrahim (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1638#-:-MRsU4bY9tdTHR3JseJa:bnfp4nl)* status: 0 of 7 files reviewed, 28 unresolved discussions (waiting on @jarifibrahim, @manishrjain, and @NamanJain8)

___
*[db.go, line 86 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsTzlDCrDXwI1V5CFH:-MRsTzlDCrDXwI1V5CFI:b-wxw16x) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/db.go#L86)):*
> ```Go
> 	lk.RLock()
> 	defer lk.RUnlock()
> 	var keys []uint64
> ```

allocate it upfront.
___
*[db.go, line 377 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsP36uCucMbRxAgTFe:-MRsP36uCucMbRxAgTFf:b7lpqng) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/db.go#L377)):*
> ```Go
> 	db.orc.incrementNextTs()
> 
> 	if err := db.setBannedNamespaces(); err != nil {
> ```

initBannedNamespaces()
___
*[db.go, line 403 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsQFMM3Jw037LRf_6m:-MRsQFMM3Jw037LRf_6n:bpsqw6h) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/db.go#L403)):*
> ```Go
> 		return nil
> 	}
> 	db.View(func(txn *Txn) error {
> ```

return this
___
*[db.go, line 1780 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsQXkI75FrSstQaZLY:-MRsQXkI75FrSstQaZLZ:b-swu9x2) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/db.go#L1780)):*
> ```Go
> }
> 
> // Checks if the key is banned. Returns the respective error is the key is prefixed by any of banned
> ```

typo: `error if the key`
___
*[db_test.go, line 2296 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsROLvEgvysbUdVcsV:-MRsROLvEgvysbUdVcsW:b-tjv98c) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/db_test.go#L2296)):*
> ```Go
> 	defer os.RemoveAll(dir)
> 
> 	ops := getTestOptions(dir).WithNamespaceOffset(3)
> ```

Check setting/reading without the offset.
___
*[db_test.go, line 2311 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsRpJm3_YMfE4wkKXt:-MRsRpJm3_YMfE4wkKXu:b-iifqfj) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/db_test.go#L2311)):*
> ```Go
> 		for i := 0; i < 10; i++ {
> 			// We store the uint64 namespace at idx=3, so first 3 bytes are insignificant to us.
> 			key := append(make([]byte, db.opt.NamespaceOffset),
> ```

[]byte(fmt.Sprintf("%03d%s-key%02d", ....))
___
*[db_test.go, line 2368 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsSRF-AdIKgGznUV2S:-MRsSRF-AdIKgGznUV2T:b-alkdr1) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/db_test.go#L2368)):*
> ```Go
> }
> 
> // Tests that the iterator skips the banned prefixes.
> ```

Add a description of the test
___
*[db_test.go, line 2375 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsSKLc3kuJ3TBre5q3:-MRsSKLc3kuJ3TBre5q4:b-r06k9l) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/db_test.go#L2375)):*
> ```Go
> 		bkey := func(prefix uint64, i int) []byte {
> 			// We store the uint64 namespace at idx=3, so first 3 bytes are insignificant to us.
> 			return append(make([]byte, db.opt.NamespaceOffset),
> ```

fmt.Sprintf(....)
___
*[db_test.go, line 2457 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsQTq71N5Wop_oOKrp:-MRsQTq71N5Wop_oOKrq:b-dqyxyv) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/db_test.go#L2457)):*
> ```Go
> 		// Iterate over the banned prefix. Passing -1 as we don't expect to access keys.
> 		validate(IteratorOptions{Prefix: getIterablePrefix('g')}, -1, 0)
> 		validate(IteratorOptions{Prefix: getIterablePrefix('g'), Reverse: true, AllVersions: true}, -1, 0)
> ```

100 chars
___
*[errors.go, line 59 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsSpOzE54CojZFvGpS:-MRsSpOzE54CojZFvGpT:bvttuun) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/errors.go#L59)):*
> ```Go
> 
> 	// ErrBannedKey is returned if the read/write key is prefixed by any banned prefix.
> 	ErrBannedKey = errors.New("Key is using the banned prefix")
> ```

belongs
___
*[errors.go, line 82 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsSunI6vbfis3ugBn0:-MRsSunI6vbfis3ugBn1:b-culoxl) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/errors.go#L82)):*
> ```Go
> 		"Invalid API request. Not allowed to perform this action using ManagedDB")
> 
> 	// ErrNamespaceMode is returned if the user tries to use an API which isn't allowed when
> ```

is allowed only in the namespace mode.
___
*[iterator.go, line 621 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsTC0P2o2_7gdsL8xG:-MRsTC0P2o2_7gdsL8xH:b-csmwfs) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/iterator.go#L621)):*
> ```Go
> 
> 	// Skip banned keys only if it does not have badger internal prefix.
> 	if it.txn.db.isBanned(key) != nil && !isInternalKey {
> ```

move the cheaper check to the left.
___
*[options.go, line 102 at r9](https://reviewable.io/reviews/dgraph-io/badger/1638#-MRsTJyz-pgonHJZdRty:-MRsTJyz-pgonHJZdRtz:bj5hbfv) ([raw file](https://github.com/dgraph-io/badger/blob/33dae7ad3fc6d271a7e04fa0e8f9d89abfb4b612/options.go#L102)):*
> ```Go
> 
> 	// NamespaceOffset specifies the offset from where the next 8 bytes contains the namespace.
> 	// The namespace is used by Dgraph for multi-tenency.
> ```

You can remove this comment.

PR #1639, opened 2021-01-13
chore: use gogofaster instead of gofast

NamanJain8:
Switch to gogofaster as it does not produce the extra XXX_ fields that increase up the space of protobuf.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1639)


pb/gen.sh
```diff
@@ -5,4 +5,4 @@
 
 # You might need to go get -v github.com/gogo/protobuf/...
 
-protoc --gofast_out=plugins=grpc:. --gofast_opt=paths=source_relative -I=. badgerpb3.proto
+protoc --gogofaster_out=plugins=grpc:. --gogofaster_opt=paths=source_relative -I=. badgerpb3.proto
```
jarifibrahim, line 8: Do we need this grpc plugin? Can we remove it?

jarifibrahim (approved):
Looks good to me. @NamanJain8 double-check this doesn't break anything. Also, check what these `XXX_...` fields were storing.

manishrjain (commented):
Yeah. Just ensure it doesn't break anything. :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1639#-:-MR_ebPA9c8-yjij4oG_:bnul538)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @NamanJain8)

pb/gen.sh
NamanJain8: No, we don't need. Yes, we can remove it. Removed.

jarifibrahim (approved):
Verify this with the older version of badger. This change shouldn't break anything.

NamanJain8:
> Verify this with the older version of badger. This change shouldn't break anything.

Verified, the backup and restore work fine. Also, read-write can be performed seamlessly across this change.

PR #1658, opened 2021-02-04
[BREAKING] fix(Subscribe): Add option to subscribe with holes in prefixes.

ahsanbarkati:
Add an argument in `(*DB) Subscribe()` for the ignore string so that the subscriber can subscribe to a prefix with holes (or ignore bytes).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1658)

### DIFF
--- a/db.go
+++ b/db.go
@@ -1826,18 +1826,17 @@ type KVList = pb.KVList
 // At least one prefix should be passed, or an error will be returned.
 // You can use an empty prefix to monitor all changes to the DB.
 // Ignore string is the byte ranges for which prefix matching will be ignored.
-// For example: ignore = "2-3", and prefix = "abcde" will match for keys "abxxe", "abdfe" etc.
+// For example: ignore = "2-3", and prefix = "abc" will match for keys "abxxc", "abdfc" etc.
 // This function blocks until the given context is done or an error occurs.
 // The given function will be called with a new KVList containing the modified keys and the
 // corresponding values.
-func (db *DB) Subscribe(ctx context.Context, cb func(kv *KVList) error, ignore string,
-	prefixes ...[]byte) error {
+func (db *DB) Subscribe(ctx context.Context, cb func(kv *KVList) error, matches []pb.Match) error {
 	if cb == nil {
 		return ErrNilCallback
 	}
 
 	c := z.NewCloser(1)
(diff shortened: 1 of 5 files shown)


db.go
```diff
@@ -1828,13 +1828,14 @@ type KVList = pb.KVList
 // This function blocks until the given context is done or an error occurs.
 // The given function will be called with a new KVList containing the modified keys and the
 // corresponding values.
-func (db *DB) Subscribe(ctx context.Context, cb func(kv *KVList) error, prefixes ...[]byte) error {
+func (db *DB) Subscribe(ctx context.Context, cb func(kv *KVList) error, ignore string,
```
NamanJain8, line 1831: Update the function description. Rest looks good. 🎉

NamanJain8 (approved):
Just a small comment. Rest looks good.

db.go
ahsanbarkati: Done. :)

db.go
```diff
@@ -1822,19 +1822,22 @@ func (db *DB) BannedNamespaces() []uint64 {
...
+// Subscribe can be used to watch key changes for the given key prefixes and the ignore string.
 // At least one prefix should be passed, or an error will be returned.
 // You can use an empty prefix to monitor all changes to the DB.
+// Ignore string is the byte ranges for which prefix matching will be ignored.
+// For example: ignore = "2-3", and prefix = "abcde" will match for keys "abxxe", "abdfe" etc.
 // This function blocks until the given context is done or an error occurs.
 // The given function will be called with a new KVList containing the modified keys and the
 // corresponding values.
-func (db *DB) Subscribe(ctx context.Context, cb func(kv *KVList) error, prefixes ...[]byte) error {
+func (db *DB) Subscribe(ctx context.Context, cb func(kv *KVList) error, ignore string,
```
jarifibrahim, line 1833: Can't we use the pb.Match here directly? There is a list of prefixes. What if I want to specify multiple ignore strings? Right now you're using the same one for all prefixes.

Also, this is a breaking change (Public API change)

ahsanbarkati: Yeah. That would be better to have. Thanks

jarifibrahim (approved):

PR #1664, opened 2021-02-09
Increase value threshold from 1 KB to 1 MB

jarifibrahim:
This PR increases the default value of ValueThreshold from 1 KB to 1 MB.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1664)

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -68,9 +68,6 @@ func (s *DB) validate() error { return s.lc.validate() }
 
 func getTestOptions(dir string) Options {
 	opt := DefaultOptions(dir).
-		WithMemTableSize(1 << 15).
-		WithBaseTableSize(1 << 15). // Force more compaction.
-		WithBaseLevelSize(4 << 15). // Force more compaction.
 		WithSyncWrites(false).
 		WithLoggingLevel(WARNING)
 	return opt
@@ -1521,7 +1518,7 @@ func TestWriteDeadlock(t *testing.T) {
 	var count int
 	val := make([]byte, 10000)
 	require.NoError(t, db.Update(func(txn *Txn) error {
-		for i := 0; i < 1500; i++ {
+		for i := 0; i < 1000; i++ {
 			key := fmt.Sprintf("%d", i)
 			rand.Read(val)
 			require.NoError(t, txn.SetEntry(NewEntry([]byte(key), val)))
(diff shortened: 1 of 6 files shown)


manishrjain (approved):
:lgtm:

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1664#-:-MT7PYRRBHZMfF4YvRub:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

funny-falcon:
But why???????
No description in PR, no description in commit message.
Why the hell you off stick main Badger purpose: store large values in a log?

strokovok:
Hi @manishrjain, I find this change a bit confusing :)
As @funny-falcon has already mentioned, it kinda contradicts to one of the key ideas behind Badger.
Also as far as I could understood, Badger's LSM tree is intended to be fully stored in RAM (correct me if I'm wrong) - which unfortunately breaks my use-case (50M entries * 200Kb per value = 9.3Tb).

Could you please comment on that? Essentially, I would like to know the reason of this change. Also, would I experience any negative consequences if setting this threshold to zero?

manishrjain:
Yeah. The original idea for Badger was relying upon the WiscKey paper -- that values can be stored separately. In practice what we found was that the value garbage collection is a real storage problem, in fast-updating datasets. The paper didn't elaborate on that much -- in production, that's a major issue -- in other words, disk amplification. If X amount of data takes 3X amount of disk space, that's a problem.

So, having much larger value log thresholds allowed us to only store the most expensive values in the value log, while most other things would stay within the LSM tree. LSM tree has frequent compactions, so things can stay bounded (there's a separate concern here about write amp).

Overall, Outserv, the fork of Dgraph that I run, no longer uses a value log. In fact, Dgraph 21.12 basically stopped using value log, or even WAL from Badger entirely.

If you happen to deal with large values and don't overwrite keys much, then a value log makes a lot of sense (say long-term file storage, where each key corresponds to a file).

P.S. Note that I'm no longer with Dgraph Labs. I maintain a fork of Badger here: https://github.com/outcaste-io/badger

strokovok:
Got it, thank you very much for your reply!
My use case doesn't imply entries overwriting at all, so perhaps I'll just lower down this setting :)

PR #1672, opened 2021-02-24
fix(discard): close the discard stats file on db close

NamanJain8:
In windows, an unclosed file cannot be removed. Which causes the failure of `os.RemoveAll()`.
See https://discuss.dgraph.io/t/dgraph-takes-all-disk-space-in-windows/12494
Fixes DGRAPH-3002

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1672)

### DIFF
--- a/value.go
+++ b/value.go
@@ -640,6 +640,11 @@ func (vlog *valueLog) Close() error {
 			err = terr
 		}
 	}
+	if vlog.discardStats != nil {
+		if terr := vlog.discardStats.Close(-1); terr != nil && err == nil {
+			err = terr
+		}
+	}
 	return err
 }
 


value.go
```diff
@@ -640,6 +640,11 @@ func (vlog *valueLog) Close() error {
 			err = terr
 		}
 	}
+	if vlog.discardStats != nil {
+		if terr := vlog.discardStats.Close(-1); terr != nil && err == nil {
```
jarifibrahim, line 644: what does -1 mean?

NamanJain8: -1 means that we don't want to truncate the file. Just Sync and unmap it. 
https://github.com/dgraph-io/ristretto/blob/8ec1dc18f880d5e12cda4f6da65141f049dc4e02/z/file.go#L185-L203

jarifibrahim: Got it.

jarifibrahim (approved):

PR #1678, opened 2021-03-04
Add a name to mutexes to make them unexported

corlinp:
By embedding `sync.Mutex` into an exported struct, it also exports the `Lock()` and `Unlock()` functions. This can cause trouble/confusion for users of Badger because `Lock()` is used internally and calling it can have unintended consequences. 

By naming the mutex, and not embedding it, the internally-used Lock and Unlock are not exposed to the user.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1678)

### DIFF
--- a/db.go
+++ b/db.go
@@ -689,8 +689,8 @@ func (db *DB) Sync() error {
 
 // getMemtables returns the current memtables and get references.
 func (db *DB) getMemTables() ([]*memTable, func()) {
-	db.RLock()
-	defer db.RUnlock()
+	db.lock.RLock()
+	defer db.lock.RUnlock()
 
 	var tables []*memTable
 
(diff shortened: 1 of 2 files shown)


jarifibrahim (approved):
This looks like a useful change.
I think a better way to do this would be to do
```go
type DB struct{
   internal struct{
       sync.Mutex
       // Some internal fields
   }
}
```
This way the API would look like
```
db.internal.lock()
db.internal.somefield = x
db.internal.unlock()
```
instead of
```
db.lock.lock()
db.somefield = x
db.lock.unlock()
```

The current placement of the lock makes it very difficult to figure out which fields it protects.
This PR can be merged. We can refactor the code later.


Thanks @corlinp :tada:

corlinp:
Thanks for the review @jarifibrahim - just FYI only those with write access to this repository can merge pull requests. Just don't want this to get too stale.

jarifibrahim:
@corlinp apologies for the delay. I didn't realize this PR was open all this time. Thank you for your contribution :+1:

PR #1692, opened 2021-04-15
fix comment about default value in WithCompression

atetubou:
zstd is not set by default even when cgo is enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1692)


NamanJain8 (approved):
Thanks, @atetubou for fixing this.

PR #1716, opened 2021-06-25
Fix conflict detection for managed DB

roysc:
I propose this simple fix for detecting conflicts in managed mode. Addresses https://discuss.dgraph.io/t/fatal-error-when-writing-conflicting-keys-in-managed-mode/14784.

When a write conflict exists for a managed DB, an internal assert can fail: https://github.com/dgraph-io/badger/blob/aaab25305b94b57e461b09d5dc9c399b1bd6635f/txn.go#L616

This occurs because a detected conflict is indicated with `commitTs` of 0, but handling the error is skipped for managed DB instances: https://github.com/dgraph-io/badger/blob/aaab25305b94b57e461b09d5dc9c399b1bd6635f/txn.go#L563.

Rather than conflate conflict detection with a timestamp of 0, it can be indicated with another return value from `hasConflict`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1716)

### DIFF
--- a/txn.go
+++ b/txn.go
@@ -160,12 +160,12 @@ func (o *oracle) hasConflict(txn *Txn) bool {
 	return false
 }
 
-func (o *oracle) newCommitTs(txn *Txn) uint64 {
+func (o *oracle) newCommitTs(txn *Txn) (uint64, bool) {
 	o.Lock()
 	defer o.Unlock()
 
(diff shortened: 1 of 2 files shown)


NamanJain8 (approved):
Thanks @roysc, for the fix. This looks good to me.

PR #1724, opened 2021-07-07
Fix the memory usage part of doc

qichengzx:
According to CHANGELOG.md#removed-apis. `TableLoadingMode` option is removed.
fixup the memory usage part of doc.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1724)


NamanJain8 (approved):
Hey @qichengzx , thanks for the fix. It would be great if you could remove `ValueLoadingMode` option doc a few lines above. That is also removed.

PR #1745, opened 2021-09-16
feat(externalMagic): Introduce external magic number

ahsanbarkati:
Currently, badger writes a magic number in the manifest. It helps badger to decide if its current version is compatible with the data on the disk or not. But now the internal data storage of badger is not changed but the data format for dgraph has changed. This causes a problem if someone starts dgraph-2109 on an older dgraph directory. This change adds an external magic number to badger which will help avoid data corruption by causing panic if opening Dgraph on the wrong directory.

There are 8 magic bytes in the manifest.
Prior to this change:
```
0-4: magic text
4-8: badgerMagicNumber
```

After this change:
```
0-4: magic text
4-6: externalMagicNumber
6-8: badgerMagicNumber
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1745)

### DIFF
--- a/badger/cmd/info.go
+++ b/badger/cmd/info.go
@@ -51,6 +51,7 @@ type flagOptions struct {
 	encryptionKey            string
 	checksumVerificationMode string
 	discard                  bool
+	externalMagicVersion     uint16
 }
 
 var (
@@ -82,6 +83,8 @@ func init() {
 		"[none, table, block, tableAndBlock] Specifies when the db should verify checksum for SST.")
 	infoCmd.Flags().BoolVar(&opt.discard, "discard", false,
 		"Parse and print DISCARD file from value logs.")
+	infoCmd.Flags().Uint16Var(&opt.externalMagicVersion, "external-magic", 0,
+		"External magic number")
 }
 
 var infoCmd = &cobra.Command{
@@ -104,7 +107,8 @@ func handleInfo(cmd *cobra.Command, args []string) error {
 		WithBlockCacheSize(100 << 20).
 		WithIndexCacheSize(200 << 20).
 		WithEncryptionKey([]byte(opt.encryptionKey)).
-		WithChecksumVerificationMode(cvMode)
+		WithChecksumVerificationMode(cvMode).
+		WithExternalMagic(opt.externalMagicVersion)
 
 	if opt.discard {
 		ds, err := badger.InitDiscardStats(bopt)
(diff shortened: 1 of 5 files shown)


manifest.go
```diff
@@ -237,20 +244,26 @@ func (mf *manifestFile) addChanges(changesParam []*pb.ManifestChange) error {
 // Has to be 4 bytes.  The value can never change, ever, anyway.
 var magicText = [4]byte{'B', 'd', 'g', 'r'}
 
-// The magic version number.
-const magicVersion = 8
+// The magic version number. It is allocated 2 bytes, so it's value must be <= math.MaxUint16
+const badgerMagicVersion = 8
```
NamanJain8, line 248: Add an assert somewhere that the `badgerMagicVersion <= math.MaxUint64`.

options.go
```diff
@@ -796,6 +800,13 @@ func (opt Options) WithNamespaceOffset(offset int) Options {
 	return opt
 }
 
+// WithExternalMagic returns a new Options value with WithExternalMagic set to the given value.
```
NamanJain8, line 803:
```suggestion
// WithExternalMagic returns a new Options value with ExternalMagic set to the given value.
```

options.go
```diff
@@ -796,6 +800,13 @@ func (opt Options) WithNamespaceOffset(offset int) Options {
...
+// The DB would fail to start if both the internal and the external magic number are not validated.
```
NamanJain8, line 804:
```suggestion
// The DB would fail to start if either the internal or the external magic number fails validation.
```

y/y.go
```diff
@@ -274,6 +274,18 @@ func BytesToU32(b []byte) uint32 {
 	return binary.BigEndian.Uint32(b)
 }
 
+// U16ToBytes converts the given Uint16 to bytes
```
NamanJain8, line 277: Move this slightly up before U32 functions.

badger/cmd/info.go
```diff
@@ -322,7 +322,7 @@ func printInfo(dir, valueDir string) error {
 			fp.Close()
 		}
 	}()
-	manifest, truncOffset, err := badger.ReplayManifestFile(fp)
+	manifest, truncOffset, err := badger.ReplayManifestFile(fp, 0)
```
NamanJain8, line 325: This should not be `zero`. We won't be able to run `badger info` on the directory with an external magic version set.

NamanJain8 (approved):
Some nit picks. Please address the comments before merging.

PR #1747, opened 2021-09-21
fix: avoid deadlock by separating subscribers and publisher

aman-bansal:
There is one scenario where badger could end up in a deadlock state. Complete goroutine dump can be found here https://drive.google.com/file/d/1nIrYlrbwlGtvk4WGDCzM0z996eEydaNI/view?usp=sharing

The issue is publisher sends out the messages over the subscriber channel and subscriber will process those message one at a time. Now subsriber channel size is 1000, so if the channel is completely filled publisher will wait indefinitly to send the message. And this is what happened. while processing the message, subscriber receive the error after 15 min and in the meantime publisher was waiting indefinitely for the channel to clear. On receiving the error subscriber asked the publisher to delete the subscriber and this turns into a deadlock state where one lock is acquired by the publisher to send the message and message processor of subscriber is waiting on that lock to delete the subscriber.

This PR tries to solve this issue by separating out the publisher and subscriber from each other and thus making them to work independently of each other. 

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1747)

### DIFF
--- a/db.go
+++ b/db.go
@@ -2127,6 +2127,12 @@ func (db *DB) Subscribe(ctx context.Context, cb func(kv *KVList) error, matches 
 			err := slurp(batch)
 			if err != nil {
 				c.Done()
+				// we ran into the deadlock state with subscription with following
+				// inactivate the subscription so we dont have any more updates on this channel
+				db.pub.inactivateSubscription(id)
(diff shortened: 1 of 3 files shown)


NamanJain8:
@aman-bansal this is not needed. Lets close this PR.

PR #1749, opened 2021-09-23
fix: avoid deadlock in publisher and subscriber

aman-bansal:
There is one scenario where badger could end up in a deadlock state. Complete goroutine dump can be found here https://drive.google.com/file/d/1nIrYlrbwlGtvk4WGDCzM0z996eEydaNI/view?usp=sharing

The issue is publisher sends out the messages over the subscriber channel and subscriber will process those message one at a time. Now subsriber channel size is 1000, so if the channel is completely filled publisher will wait indefinitly to send the message. And this is what happened. while processing the message, subscriber receive the error after 15 min and in the meantime publisher was waiting indefinitely for the channel to clear. On receiving the error subscriber asked the publisher to delete the subscriber and this turns into a deadlock state where one lock is acquired by the publisher to send the message and message processor of subscriber is waiting on that lock to delete the subscriber.

This PR tries to solve this issue by adding a atomic variable in subscriber so that publisher can stop sending out the new updates and thus making the deadlock to be released eventually.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1749)

### DIFF
--- a/publisher_test.go
+++ b/publisher_test.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"fmt"
 	"github.com/pkg/errors"
-	"log"
 	"sync"
 	"sync/atomic"
 	"testing"
--- a/publisher_test.go
+++ b/publisher_test.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/pkg/errors"
+	"log"
 	"sync"
 	"sync/atomic"
 	"testing"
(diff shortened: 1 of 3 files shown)


publisher_test.go
```diff
@@ -18,14 +18,69 @@ package badger
...
+		var firstUpdate sync.WaitGroup
+		firstUpdate.Add(1)
+
+
+		var subDone sync.WaitGroup
+		subDone.Add(1)
+		go func() {
+			subWg.Done()
+			match := pb.Match{Prefix: []byte("ke"), IgnoreBytes: ""}
+			db.Subscribe(context.Background(), func(kvs *pb.KVList) error {
```
NamanJain8, line 49: Lets catch the errors from db.Subscribe, db.Update, etc as well.

publisher_test.go
```diff
@@ -18,14 +18,69 @@ package badger
...
+				firstUpdate.Done()
+				time.Sleep(time.Second * 20)
```
NamanJain8, line 51: Can this sleep time be reduced? It adds 20s to each of the CI build.

NamanJain8 (approved):
Looks good to me.

publisher_test.go
aman-bansal: it takes 13 seconds on my laptop to push all the updates. that is necessary to create the deadlock situation thats why i added 20 sec delay to be on safe side

publisher_test.go
aman-bansal: yup sounds good

PR #1757, opened 2021-11-05
BTree-based pending writes storage in transcation

dbolotin:
https://discuss.dgraph.io/t/iteration-over-keys-added-in-transaction/15811

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1757)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/protobuf v1.3.1
 	github.com/golang/snappy v0.0.3
+	github.com/google/btree v1.0.1 // indirect
 	github.com/google/flatbuffers v1.12.1
 	github.com/google/go-cmp v0.5.4 // indirect
 	github.com/klauspost/compress v1.12.3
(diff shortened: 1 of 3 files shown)


iluminae:
it is worth noting that dgraph maintains a btree implementation in https://github.com/dgraph-io/ristretto/blob/master/z/btree.go - not to say the google one is not good but they may prefer to stay consistent.

dbolotin:
@iluminae that implementation seems to be a "`map[uint64]uint64`", while this case requires a map with `[]byte` keys, sorted accordingly.

Small update: since I opened this PR, I used it in a project, where the code seems to be working ok. I also went away form huge transactions, an was able to split them into smaller pieces (mostly for different reasons), but still, current implementation of the pending writes iterator, and approach when resorting of pending records happens on each iterator creation, for me, seems to be a sub-optimal solution. In my opinion the PR stands even for the "small transaction" use-case, recommended by dgraph developers. 

In case a small performance penalty for write only cases is of a concern, I thought that it is possible to perform switch to a BTree structure for pending writes, only after user first time creates an iterator. If it is required, I will be happy to implement it.

PR #1772, opened 2022-02-07
fix(stream): improve error handling

SaveTheRbtz:
This fixes error handling during streaming.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1772)

### DIFF
--- a/badger/cmd/stream.go
+++ b/badger/cmd/stream.go
@@ -123,7 +123,8 @@ func stream(cmd *cobra.Command, args []string) error {
 
 	} else if len(so.outFile) > 0 {
 		stream.LogPrefix = "DB.Backup"
-		f, err := os.OpenFile(so.outFile, os.O_RDWR|os.O_CREATE, 0666)
+		var f *os.File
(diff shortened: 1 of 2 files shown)


akon-dey:
@SaveTheRbtz thank you for cleaning this up. Please add a unit test to cover these fixes.

SaveTheRbtz:
@akon-dey any suggestions how to best test it?  It looks like I can only do that by changing the `StreamDB` API to dependency inject the writer.

akon-dey:
@SaveTheRbtz I guess it's more work and the code looks good (also from offline feedback from former maintainers). I will merge it. Thank you for working on this.

SaveTheRbtz:
I did a stab at trying to figure out how to best test it.

Converting `StreamWriter` to an interface may work, e.g.: https://github.com/dgraph-io/badger/compare/master...SaveTheRbtz:streamTests?expand=1
I'm not sure how much of that worth the added complexity, but for the ease of testing all public structures in this package should likely be interfaces, so that they can easily be mocked in all kinds of tests (incl. tests by the users of the library, when they want to mock badger database.)

PR #1785, opened 2022-04-29
fix typo

TensShinet:
fix typo: from `contigous` to `contiguous`


TensShinet (commented):
It's a spelling mistake. It should be changed to contiguous

PR #1788, opened 2022-06-06
feat(txn): add public method for getting the value of discarded field

matino:
I have a use case, where there's a network service on top of the BadgerDB. The client can start transaction, do their stuff, and finally commit, all via separate network calls. This makes it impossible to `defer` transaction `Discard` call, as the commit/rollback is executed in a different context. With the current design, I need to send a "ping" `Get` call, to check if the transaction is discarded, with the proposed change things get much simpler.

### DIFF
--- a/txn.go
+++ b/txn.go
@@ -760,6 +760,11 @@ func (txn *Txn) ReadTs() uint64 {
 	return txn.readTs
 }
 
+// Discarded returns the discarded value of the transaction.
+func (txn *Txn) Discarded() bool {
+	return txn.discarded
+}
+
 // NewTransaction creates a new transaction. Badger supports concurrent execution of transactions,
 // providing serializable snapshot isolation, avoiding write skews. Badger achieves this by tracking
 // the keys read and at Commit time, ensuring that these read keys weren't concurrently modified by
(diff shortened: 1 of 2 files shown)


akon-dey:
@matino adding the method seems to be fine since all it does is return whether the transaction has been discarded or not. However, I have concerns about your use case. I wonder whether you will still need to do reference counting of the transaction in each client thread before Discard can be called.

matino:
@akon-dey Can you please elaborate?

akon-dey:
My point was that you still have a problem with the transaction being
discarded/committed  when another client may expect their to be active.

matino:
> My point was that you still have a problem with the transaction being discarded/committed when another client may expect their to be active.

That's true :) In fact the real problem I face, is to detect the`ErrDiscardedTxn` and return my own error instead (in order to maintain clear separation of the storage layer). Currently, the only way to detect it, is via `Get` method, which is not really optimal.

akon-dey (commented):
@matino please add unit tests to txn_test.go. The changes look good otherwise.

matino:
> @matino please add unit tests to txn_test.go. The changes look good otherwise.

@akon-dey done

pandeyshubham25 (commented):
@matino Thanks for the changes
The added test case also looks good to me.

akon-dey (approved):

akon-dey:
Thank you for your contribution @matino!

PR #1790, opened 2022-06-22
Bump xxhash to v2 to fix https://github.com/cespare/xxhash/issues/54

mjungsbluth:
Fixes build issues when using badger in a Golang plugin with 1.17+

See https://github.com/cespare/xxhash/issues/54 for the upstream fix


pandeyshubham25 (approved):
@mjungsbluth Thanks for the changes! Version upgrade looks good to me

joshua-goldstein (approved):

PR #1784, opened 2022-04-22
GC discardStat fallback mechanism

requilence:
This PR aims to fix the GC issue described here
https://discuss.dgraph.io/t/gc-may-not-work-in-some-cases/17197

The idea is to calculate discardStat directly on vlog in case discard data is not available. Because current discardStat calculation only done on compaction it may be not available until memtables got full. Check the issue on forum for the reproduction steps.

Also during the tests I've found out that value log rewrite doesn't work well when the key was removed so I've tried to fix this at the line 238-240

PR #1792, opened 2022-07-06
initialize PR for fix of race condition due to calculateDiscardStat

joshua-goldstein:
Opening branch and PR for fix to race condition introduced by previous merge.  See discussion [here](https://github.com/dgraph-io/badger/pull/1784).


joshua-goldstein:
@lithp We have created a PR based on the deadlock issue you brought up.  We would be happy to add you as a reviewer, or if you would like to create your own PR we can take a look at that as well.  Thanks again.

PR #1800, opened 2022-08-14
fix: update maxHeaderSize

microyahoo:
Based on [pkg-constants](https://pkg.go.dev/encoding/binary#pkg-constants), the `maxHeaderSize` should be 22=5+5+10+1+1, which is discussed on [Why is the `maxHeaderSize` 21 and not 22?](https://discuss.dgraph.io/t/why-is-the-maxheadersize-21-and-not-22/17644)

Signed-off-by: Liang Zheng <zhengliang0901@gmail.com>

### DIFF
--- a/structs.go
+++ b/structs.go
@@ -72,8 +72,8 @@ type header struct {
 
 const (
 	// Maximum possible size of the header. The maximum size of header struct will be 18 but the
-	// maximum size of varint encoded header will be 21.
-	maxHeaderSize = 21
+	// maximum size of varint encoded header will be 22.
+	maxHeaderSize = 22
 )
 
 // Encode encodes the header into []byte. The provided []byte should be atleast 5 bytes. The


microyahoo:
@jarifibrahim PTAL, thanks.

jarifibrahim (commented):
I think this might break existing badger DBs because we use this value to determine where to write the data. 

We'll need to bump the `magicVersion` over here to prevent existing databases from breaking https://github.com/dgraph-io/badger/blob/2de177900e3ced14fdab91aa7c6a7e3e441f0192/manifest.go#L248

Also, once this PR is merged, we'll need a major version update.

joshua-goldstein (approved):
Thanks for the contribution.  This does indeed look like a bug which could cause a panic in 
```
func (h header) Encode(out []byte) int {
	out[0], out[1] = h.meta, h.userMeta
	index := 2
	index += binary.PutUvarint(out[index:], uint64(h.klen))
	index += binary.PutUvarint(out[index:], uint64(h.vlen))
	index += binary.PutUvarint(out[index:], h.expiresAt)
	return index
}
```
since the size of the `out` buffer is allocated based on this parameter.  We haven't seen this issue before since it would require each varInt encoded field to be its maximum size, but it certainly could happen.  This doesn't require a magic number upgrade since this is fixing an edge case.

mangalaman93:
@joshua-goldstein could you write a test for it so that in case something changes, the test fails. We also should figure out what are the cases when the header size could become 22.

mangalaman93:
I am closing this PR in lieu of #1877. Once again, thank you for your contribution

PR #1807, opened 2022-10-11
chore: add github templates + cleanup

joshua-goldstein:
## Problem

Links are broken, and issue / PR templates don't exist.  Codeowners is out of date.

## Solution

We also add github templates to bring the repo to parity with Dgraph and Ristretto.  We update the codeowners file and fix some links.


skrdgraph (approved):
looks good 👍 merge away

PR #1813, opened 2022-10-13
chore: add coveralls

joshua-goldstein:
## Problem

We are not reporting test coverage.

## Solution

We show test coverage with coveralls.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -4,6 +4,15 @@ set -eo pipefail
 
 go version
 
+# Check if Github Actions is running
+if [ $CI = "true" ]; then
+  # Enable code coverage
+  # export because tests run in a subprocess
+  export covermode="-covermode=atomic"
+  export coverprofile="-coverprofile=cover_tmp.out"
+  echo "mode: atomic" >> cover.out
+fi
+
 # Run `go list` BEFORE setting GOFLAGS so that the output is in the right
 # format for grep.
 # export packages because the test will run in a sub process.
@@ -11,9 +20,9 @@ export packages=$(go list ./... | grep "github.com/dgraph-io/badger/v3/")
 
 tags="-tags=jemalloc"
 
-# Ensure that we can compile the binary.
+# Compile the Badger binary
 pushd badger
-go build -v $tags .
+  go build -v $tags .
 popd
 
 # Run the memory intensive tests first.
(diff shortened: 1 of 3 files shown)


test.sh
```diff
@@ -4,16 +4,25 @@ set -eo pipefail
...
 # export packages because the test will run in a sub process.
 export packages=$(go list ./... | grep "github.com/dgraph-io/badger/v3/")
 
 tags="-tags=jemalloc"
 
-# Ensure that we can compile the binary.
+# Compile the Badger binary
 pushd badger
-go build -v $tags .
+  go build -v $tags .
```
skrdgraph, line 25: fix the extra space* (I think it was added accidentally)

test.sh
```diff
@@ -61,32 +82,44 @@ root() {
 
   echo "==> Running root level tests."
   set -e
-  go test $tags -timeout=25m . -v -race -parallel 16
+  go test $tags -v -race -parallel=16 -timeout=25m $covermode $coverprofile .
+  write_coverage
```
skrdgraph, line 86: you could do this in 1line, with an `&&`, I think that would look cleaner

test.sh
```diff
@@ -61,32 +82,44 @@ root() {
...
+    ./badger benchmark read --dir=$baseDir/test --full-scan | tee --append $baseDir/log.txt
+    ./badger benchmark read --dir=$baseDir/test -d=30s | tee --append $baseDir/log.txt
+    ./badger stream --dir=$baseDir/test -o "$baseDir/test2" | tee --append $baseDir/log.txt
+    count=$(cat "$baseDir/log.txt" | grep "at program end: 0 B" | wc -l)
+    rm -rf $baseDir
+    if [ $count -ne 4 ]; then
+      echo "LEAK detected in Badger stream."
+      return 1
+    fi
+    echo "==> DONE stream test"
```
skrdgraph, line 104: just curious, why the extra space?

test.sh
```diff
@@ -61,32 +82,44 @@ root() {
...
+      cat cover_tmp.out >> cover.out && rm cover_tmp.out
+    fi
+  fi
+  
+}
+
 export -f stream
 export -f manual
 export -f root
+export -f write_coverage
```
skrdgraph, line 122: I think this is redundant, you may not need this line

test.sh
joshua-goldstein: Done 👍

test.sh
joshua-goldstein: I think that got pushed by accident, I reverted it

test.sh
joshua-goldstein: In that case do we need any of these export statements?  I'm not sure why we needed them to the environment

joshua-goldstein: @skrdgraph  I removed all export statements, couldn't understand why they were there

skrdgraph (approved):

test.sh
skrdgraph: 👍

PR #1838, opened 2022-12-14
chore(ci): update cd workflows for arm64

joshua-goldstein:
## Problem

Currently we only deploy amd64 badger CLI tool builds.  We would like arm64 builds too.

## Solution

Use an arm64 self-hosted runner to build arm64 badger CLI tool.


skrdgraph (approved):
Looks good!

meghalims (approved):
Looks good

PR #1842, opened 2022-12-15
chore(iterator): `yieldItemValue` no error return

qdm12:
## Problem

I was trying to trigger an error from `func (item *Item) ValueCopy(dst []byte) ([]byte, error)` in a unit test, but it turns out errors cannot happen. This just makes internal code clearer no error can happen.

## Solution

1. Change `func (item *Item) yieldItemValue() ([]byte, func(), err error)` -> `func (item *Item) yieldItemValue() ([]byte, func())`
2. Keep exported methods with an error returned to keep compatibility, although some such as `ValueCopy` cannot produce a non-nil error (as it was before)

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -101,11 +101,8 @@ func (item *Item) Value(fn func(val []byte) error) error {
 		}
 		return item.err
 	}
-	buf, cb, err := item.yieldItemValue()
+	buf, cb := item.yieldItemValue()
 	defer runCallback(cb)
-	if err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


mangalaman93:
@qdm12 Could you cherry pick this change on `release/v4.0` branch instead? Thanks

mangalaman93:
@qdm12 will it be possible for you to cheery pick the change on latest main?

qdm12:
Done, sorry for the delay!

PR #1849, opened 2023-01-24
chore(ci): resolve send coverage errors

joshua-goldstein:
## Problem

CI is unable to send coverage results to coveralls.io.


skrdgraph (dismissed):

skrdgraph:
seems to have failed ... because of the same issue. @joshua-goldstein - I am not sure what's causing this: 
```
422 Unprocessable Entity response status code indicates that the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.
```

I am thinking of regenerating this token.

skrdgraph:
Looks like we really need to debug this failure now. We have tried changing the coveralls token as potential fix.

PR #1862, opened 2023-02-11
docs: add project that use badger in README.md

rfyiamcool:


joshua-goldstein:
Many thanks for the contribution.  Could you please also add the project to https://github.com/dgraph-io/badger/blob/main/docs/content/projects-using-badger/index.md?

PR #1837, opened 2022-12-11
chore(tests): unbreak tests on Windows (#1775)

akon-dey:
Currently [appveyor tests](https://ci.appveyor.com/project/manishrjain/badger/builds/42502297) are failing in multiple places on Windows.

Original PR #1775

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1775)

PR #1775, opened 2022-02-09
chore(tests): unbreak tests on Windows

SaveTheRbtz:
Currently [appveyor tests](https://ci.appveyor.com/project/manishrjain/badger/builds/42502297) are failing in multiple places on Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1775)

PR #1865, opened 2023-02-13
chore(tests): unbreak tests on Windows (#1775) #1837

joshua-goldstein:
Cherry pick of #1775 and #1837 .

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -2234,11 +2234,13 @@ func TestOpenDBReadOnly(t *testing.T) {
 
 	ops.ReadOnly = true
 	db, err = Open(ops)
-	require.NoError(t, err)
-	require.NoError(t, db.Close())
+	if runtime.GOOS == "windows" {
+		require.Equal(t, err, ErrWindowsNotSupported)
+	} else {
+		require.NoError(t, err)
+		require.NoError(t, db.Close())
+	}
 
-	db, err = Open(ops)
-	require.NoError(t, err)
 	var count int
 	read := func() {
 		count = 0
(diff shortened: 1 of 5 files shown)


value_test.go
```diff
@@ -724,8 +725,12 @@ func TestReadOnlyOpenWithPartialAppendToWAL(t *testing.T) {
 	opts.ReadOnly = true
 	// Badger should fail a read-only open with values to replay
 	_, err = Open(opts)
-	require.Error(t, err)
-	require.Regexp(t, "Log truncate required", err.Error())
+	if runtime.GOOS == "windows" {
+		require.Equal(t, err, ErrWindowsNotSupported)
```
mangalaman93, line 729: could you just return after this line? and not have the else block

db_test.go
```diff
@@ -2279,10 +2285,14 @@ func TestOpenDBReadOnly(t *testing.T) {
 
 	ops.ReadOnly = true
 	db, err = Open(ops)
-	require.NoError(t, err)
-	read()
-	require.Equal(t, 20, count)
-	require.NoError(t, db.Close())
+	if runtime.GOOS == "windows" {
+		require.Equal(t, err, ErrWindowsNotSupported)
```
mangalaman93, line 2289: same here, return after this line and we don't need the else line

db_test.go
```diff
@@ -2234,11 +2234,13 @@ func TestOpenDBReadOnly(t *testing.T) {
 
 	ops.ReadOnly = true
 	db, err = Open(ops)
-	require.NoError(t, err)
-	require.NoError(t, db.Close())
+	if runtime.GOOS == "windows" {
+		require.Equal(t, err, ErrWindowsNotSupported)
```
mangalaman93, line 2238: same here

mangalaman93 (changes requested):

value_test.go
joshua-goldstein: done

mangalaman93:
LGTM. Just one more thing, can we add a comment explaining why things break on windows? And how do we know that the tests now pass on windows?

mangalaman93:
One more question, do we run CI on windows?

PR #1866, opened 2023-02-14
chore(memtable): refactor code for memtable flush

mangalaman93:
Removes the `flushTask` struct that was used in two different context before. Once place when we were sending the memtable for flushing it onto the disk after it becomes full, and other context is when we were sending the memtable to be written onto disk during flush.

### DIFF
--- a/db.go
+++ b/db.go
@@ -109,7 +109,7 @@ type DB struct {
 	lc        *levelsController
 	vlog      valueLog
 	writeCh   chan *request
-	flushChan chan flushTask // For flushing memtables.
+	flushChan chan *memTable // For flushing memtables.
 	closeOnce sync.Once      // For closing DB only once.
 
 	blockWrites atomic.Int32
@@ -240,7 +240,7 @@ func Open(opt Options) (*DB, error) {
 
 	db := &DB{
 		imm:              make([]*memTable, 0, opt.NumMemtables),
-		flushChan:        make(chan flushTask, opt.NumMemtables),
+		flushChan:        make(chan *memTable, opt.NumMemtables),
 		writeCh:          make(chan *request, kvWriteChCapacity),
 		opt:              opt,
 		manifest:         manifestFile,
@@ -351,11 +351,11 @@ func Open(opt Options) (*DB, error) {
 
 		db.closers.memtable = z.NewCloser(1)
 		go func() {
-			_ = db.flushMemtable(db.closers.memtable) // Need levels controller to be up.
+			db.flushMemtable(db.closers.memtable) // Need levels controller to be up.
 		}()
 		// Flush them to disk asap.
 		for _, mt := range db.imm {
-			db.flushChan <- flushTask{mt: mt}
+			db.flushChan <- mt
 		}
 	}
 	// We do increment nextTxnTs below. So, no need to do it here.
@@ -568,12 +568,12 @@ func (db *DB) close() (err error) {
 		} else {
 			db.opt.Debugf("Flushing memtable")
 			for {
-				pushedFlushTask := func() bool {
+				pushedMemTable := func() bool {
 					db.lock.Lock()
 					defer db.lock.Unlock()
 					y.AssertTrue(db.mt != nil)
 					select {
-					case db.flushChan <- flushTask{mt: db.mt}:
+					case db.flushChan <- db.mt:
 						db.imm = append(db.imm, db.mt) // Flusher will attempt to remove this from s.imm.
 						db.mt = nil                    // Will segfault if we try writing!
 						db.opt.Debugf("pushed to flush chan\n")
@@ -586,7 +586,7 @@ func (db *DB) close() (err error) {
 					}
 					return false
 				}()
-				if pushedFlushTask {
+				if pushedMemTable {
 					break
 				}
 				time.Sleep(10 * time.Millisecond)
@@ -826,6 +826,7 @@ func (db *DB) writeRequests(reqs []*request) error {
 		}
 		count += len(b.Entries)
 		var i uint64
+		var err error
 		for err = db.ensureRoomForWrite(); err == errNoRoom; err = db.ensureRoomForWrite() {
 			i++
 			if i%100 == 0 {
@@ -987,7 +988,7 @@ func (db *DB) ensureRoomForWrite() error {
 	}
 
 	select {
-	case db.flushChan <- flushTask{mt: db.mt}:
+	case db.flushChan <- db.mt:
 		db.opt.Debugf("Flushing memtable, mt.size=%d size of flushChan: %d\n",
 			db.mt.sl.MemSize(), len(db.flushChan))
 		// We manage to push this task. Let's modify imm.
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -1074,11 +1071,48 @@ func (db *DB) handleFlushTask(ft flushTask) error {
 func (db *DB) flushMemtable(lc *z.Closer) error {
 	defer lc.Done()
 
+	var sz int64
+	var itrs []y.Iterator
+	var mts []*memTable
+	slurp := func() {
```
joshua-goldstein, line 1077: can we check the size before we wait on the channel?

db.go
```diff
@@ -755,13 +755,6 @@ var requestPool = sync.Pool{
 }
 
 func (db *DB) writeToLSM(b *request) error {
-	// We should check the length of b.Prts and b.Entries only when badger is not
-	// running in InMemory mode. In InMemory mode, we don't write anything to the
-	// value log and that's why the length of b.Ptrs will always be zero.
-	if !db.opt.InMemory && len(b.Ptrs) != len(b.Entries) {
-		return errors.Errorf("Ptrs and Entries don't match: %+v", b)
-	}
-
```
joshua-goldstein, line 764: should we keep this check?

db.go
```diff
@@ -1074,11 +1071,48 @@ func (db *DB) handleFlushTask(ft flushTask) error {
...
 
+	var sz int64
+	var itrs []y.Iterator
+	var mts []*memTable
+	slurp := func() {
+		for {
+			select {
+			case more := <-db.flushChan:
+				if more.mt == nil {
+					return
```
joshua-goldstein, line 1082: continue or return?

options.go
```diff
@@ -148,7 +148,7 @@ func DefaultOptions(path string) Options {
 		NumCompactors:           4, // Run at least 2 compactors. Zero-th compactor prioritizes L0.
 		NumLevelZeroTables:      5,
 		NumLevelZeroTablesStall: 15,
-		NumMemtables:            5,
+		NumMemtables:            15,
```
joshua-goldstein, line 151: performance/memory implications?

joshua-goldstein (changes requested):

db.go
mangalaman93: done

mangalaman93:
These are the places where we are sending the memtable to the `flushchan` channel:
1. when we open the db and find existing memtables https://github.com/dgraph-io/badger/blob/release/v4.0/db.go#L358
2. when we close the db https://github.com/dgraph-io/badger/blob/origin/v4.0/db.go#L576
3. when it is full https://github.com/dgraph-io/badger/blob/release/v4.0/db.go#L986

The only place this code will be useful when we open and close the DB. If the DB is closed and opened a few times in short duration, it may end up collecting a few small sized memtables and will require compaction in L0 -> L1 (or higher) unnecessarily. This will increase write amplification unnecessarily. I think this code only helps us, doesn't harm us in any way.

options.go
mangalaman93: switched it back to 5 keeping the read speed in mind.

db.go
mangalaman93: I think that is what we are doing, I think I got confused or didn't understand what @harshil-goel meant.

db.go
mangalaman93: done

db.go
```diff
@@ -1009,12 +1010,12 @@ func arenaSize(opt Options) int64 {
...
 // buildL0Table builds a new table from the memtable.
-func buildL0Table(ft flushTask, bopts table.Options) *table.Builder {
-	iter := ft.mt.sl.NewIterator()
+func buildL0Table(iter y.Iterator, dropPrefixes [][]byte, bopts table.Options) *table.Builder {
 	defer iter.Close()
+
 	b := table.NewTableBuilder(bopts)
-	for iter.SeekToFirst(); iter.Valid(); iter.Next() {
-		if len(ft.dropPrefixes) > 0 && hasAnyPrefixes(iter.Key(), ft.dropPrefixes) {
+	for iter.Rewind(); iter.Valid(); iter.Next() {
```
harshil-goel, line 1017: Whats the difference  between revind and seektofirst?

db.go
```diff
@@ -1024,23 +1025,14 @@ func buildL0Table(ft flushTask, bopts table.Options) *table.Builder {
...
-type flushTask struct {
-	mt           *memTable
-	dropPrefixes [][]byte
+	return b
 }
 
-// handleFlushTask must be run serially.
-func (db *DB) handleFlushTask(ft flushTask) error {
-	// There can be a scenario, when empty memtable is flushed.
-	if ft.mt.sl.Empty() {
```
harshil-goel, line 1038: lets keep this check?

harshil-goel (dismissed):

db_test.go
```diff
@@ -1886,7 +1886,11 @@ func ExampleOpen() {
 	if err != nil {
 		panic(err)
 	}
-	defer db.Close()
+	defer func() {
+		if err := db.Close(); err != nil {
+			panic(err)
```
joshua-goldstein, line 1891: can we use `y.Check(db.Close())` here?

db_test.go
```diff
@@ -1942,7 +1946,11 @@ func ExampleTxn_NewIterator() {
...
+		}
```
joshua-goldstein, line 1952: can we use `y.Check(db.Close())` here?

db.go
```diff
@@ -1758,13 +1751,9 @@ func (db *DB) DropPrefix(prefixes ...[]byte) error {
 			memtable.DecrRef()
 			continue
 		}
-		task := flushTask{
-			mt: memtable,
-			// Ensure that the head of value log gets persisted to disk.
-			dropPrefixes: filtered,
```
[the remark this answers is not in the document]
mangalaman93: this goes as an argument to the function `handleMemTableFlush` (prev called `handleFlushTask`)

db_test.go
mangalaman93: Done

db_test.go
mangalaman93: Done

db.go
mangalaman93: SeekToFirst is only available on a memtable iterator whereas Rewind function is available on both (uni and merge iterators)

joshua-goldstein (changes requested):
- CI is failing on TestPenultimateMemCorruption ([link)](https://github.com/dgraph-io/badger/actions/runs/4301023562/jobs/7497795293)
- Please add a description to the PR explaining why this change is useful / necessary (I see your comment makes some of this clear, but it should also be in the description)
- This change was based on changes that were in another PR in Badger, this should be referenced
- We should benchmark the performance benefits of this change. Currently our assumption is this won't do any harm and may have some benefits when opening/closing the DB, but we should be able to quantify this benefit.  Adding a benchmark would also be useful for future changes.

joshua-goldstein:
Looks like it is based off https://github.com/dgraph-io/badger/pull/1696 - is this change necessary to bring in the handover skiplist feature?

mangalaman93:
> * CI is failing on TestPenultimateMemCorruption ([link)](https://github.com/dgraph-io/badger/actions/runs/4301023562/jobs/7497795293)
=>  looking at the failure
> 
>     * Please add a description to the PR explaining why this change is useful / necessary (I see your comment makes some of this clear, but it should also be in the description)
=> updated the description
> 
>     * This change was based on changes that were in another PR in Badger, this should be referenced
=> It is not anymore. Earlier it was using some changes from another commit but I realized that we don't want those changes either because it was doubling the size of L0 sstables to twice of the expected size.
> 
>     * We should benchmark the performance benefits of this change. Currently our assumption is this won't do any harm and may have some benefits when opening/closing the DB, but we should be able to quantify this benefit.  Adding a benchmark would also be useful for future changes.
=> We are only making structural changes to the code, the logic is the same as it was before. I don't expect the performance to change at all.

mangalaman93:
> 

This change is not based off of #1696 any more. That is how the change started but I have removed the code to club together memtables.

joshua-goldstein (approved):

PR #1698, opened 2021-04-29
feat(dropPrefix): add DropPrefixNonBlocking API

NamanJain8:
Related to DGRAPH-3319

This PR adds `DropPrefixNonBlocking`  API that can be used to logically delete the data for specified prefixes at a given timestamp. This also adds an equivalent API of older `DropPrefix` == `DropPrefixBlocking`. 
`DropPrefix` now makes decision based on badger option `AllowStopTheWorld` whose default is to use `DropPrefixBlocking`.
The data would not be cleared from the LSM tree immediately. It would be deleted eventually through compactions.
This operation is useful when we don't want to block writes while we delete the prefixes (`DropPrefix` blocks writes).
It does this in the following way:
- Stream the given prefixes.
- Write them to skiplist and handover that skiplist to DB.

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1698)

PR #1867, opened 2023-02-14
compute block length accounting for compression

mangalaman93:
Taken from PR #1698 and commit da5f789


db.go
```diff
@@ -755,6 +755,8 @@ var requestPool = sync.Pool{
 }
 
 func (db *DB) writeToLSM(b *request) error {
+	db.lock.RLock()
+	defer db.lock.RUnlock()
```
joshua-goldstein, line 759: this lock was introduced because it was needed when HandoverSkiplist was introduced - it might not be necessary now

mangalaman93: removed the lock.

joshua-goldstein (approved):

PR #1722, opened 2021-07-05
feat(stream): add support for incremental stream writer

NamanJain8:
This PR adds support for stream writing incrementally to the DB.
Adds an API: `StreamWriter.PrepareIncremental`

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/badger/1722)

PR #1874, opened 2023-02-14
feat(stream): add support for incremental stream writer (#1722)

mangalaman93:
This PR adds support for stream writing incrementally to the DB.
Adds an API: StreamWriter.PrepareIncremental

### DIFF
--- a/stream_writer.go
+++ b/stream_writer.go
@@ -48,6 +48,7 @@ type StreamWriter struct {
 	throttle   *y.Throttle
 	maxVersion uint64
 	writers    map[uint32]*sortedWriter
+	prevLevel  int
 }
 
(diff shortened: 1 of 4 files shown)


stream_writer.go
```diff
@@ -75,18 +75,58 @@ func (db *DB) NewStreamWriter() *StreamWriter {
...
+			sw.prevLevel = level.Level
+			isEmptyDB = false
+			break
+		}
+	}
+	if isEmptyDB {
+		// If DB is empty, we should allow doing incremental stream write.
+		return nil
+	}
+	if sw.prevLevel == 0 {
```
joshua-goldstein, line 126: can we write a test to exercise this case?

joshua-goldstein (approved):
Looks good!

PR #1880, opened 2023-02-20
perf: Upgrade compress package

klauspost:
## Solution

Upgrade for faster/better compression: https://github.com/klauspost/compress/compare/v1.12.3...v1.15.15

Use 's2' in Snappy compatible mode for faster/better Snappy (de)compression:

https://github.com/klauspost/compress/tree/master/s2#blocks

Should improve compression speed, and `MaxEncodedLen` is considerably less. From `n = 32 + n + n/6` to `n + ((log2(n)+7)/7)` (max 10 bytes)

Use zstd header to determine decompressed size.

This change is both forward and backwards compatible, meaning output will be decompressable by older versions as well.

I cannot run tests or benchmarks as both are completely broken on Windows.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,9 +8,8 @@ require (
 	github.com/dustin/go-humanize v1.0.0
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/protobuf v1.3.1
-	github.com/golang/snappy v0.0.3
 	github.com/google/flatbuffers v1.12.1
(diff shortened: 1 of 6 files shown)


mangalaman93:
Thanks for the PR, let me do some testing and run some benchmarks and get back to you. Please gimme a few days.

mangalaman93:
This is done as part of #2009. Thanks for the PR.

PR #1886, opened 2023-02-24
chore(ci): update changelog + import paths for v4

joshua-goldstein:
Changelog should reflect that we are switching to SemVer.  This requires moving to a new major version.  We also update all import paths in Badger.


skrdgraph (dismissed):

skrdgraph (approved):

skrdgraph (approved):
👍

PR #1887, opened 2023-02-27
chore(cd): tag based deployments

skrdgraph:
## Problem
issue example ->
Screen Shot 2023-02-27 at 19 36 26

Needs a better standardization

## Solution
solution example ->
Screen Shot 2023-02-27 at 19 36 19

Migrating to this model

### DIFF
--- a/.github/workflows/cd-badger.yml
+++ b/.github/workflows/cd-badger.yml
@@ -1,10 +1,18 @@
 name: cd-badger
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+      releasetag:
+        description: 'releasetag'
+        required: true
+        type: string
 jobs:
   badger-build-amd64:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: '${{ github.event.inputs.releasetag }}'
       - name: Get Go Version
         run: |
           #!/bin/bash
@@ -17,12 +25,12 @@ jobs:
       - name: Set Badger Release Version
         run: |
           #!/bin/bash
-          GIT_BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD)
-          if [[ "$GIT_BRANCH_NAME" == "release/v"* ]]; 
+          GIT_BRANCH_NAME='${{ github.event.inputs.releasetag }}'
+          if [[ "$GIT_BRANCH_NAME" == "v"* ]]; 
           then
-            echo "this is a release branch"
+            echo "this is a release tag"
           else
-            echo "this is NOT a release branch"
+            echo "this is NOT a release tag"
             exit 1
           fi
           BADGER_RELEASE_VERSION=$(git rev-parse --abbrev-ref HEAD | sed  's/release\///')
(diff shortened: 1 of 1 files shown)


.github/workflows/cd-badger.yml
```diff
@@ -1,10 +1,18 @@
 name: cd-badger
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+      releasetag:
+        description: 'releasetag'
+        required: true
+        type: string
```
skrdgraph, line 8: user entered

joshua-goldstein: this workflow is triggered manually via the github UI; and the tag is applied once the release is cut on the github UI

joshua-goldstein: essentially release has the steps:
- run cd-dgraph workflow (pipeline); this will build dgraph and create artifacts and push images to dockerhub
- make release on github ui; as part of this step github will automatically apply a tag at the specific commit we are releasing
- upload artifacts to the release on github

.github/workflows/cd-badger.yml
```diff
@@ -17,12 +25,12 @@ jobs:
...
+          if [[ "$GIT_TAG_NAME" == "v"* ]]; 
           then
-            echo "this is a release branch"
+            echo "this is a release tag"
           else
-            echo "this is NOT a release branch"
+            echo "this is NOT a release tag"
             exit 1
           fi
           BADGER_RELEASE_VERSION=$(git rev-parse --abbrev-ref HEAD | sed  's/release\///')
```
joshua-goldstein, line 36: this can be updated

.github/workflows/cd-badger.yml
```diff
@@ -58,12 +68,12 @@ jobs:
...
```
joshua-goldstein, line 79: same:
this can be updated
```BADGER_RELEASE_VERSION='${{ github.event.inputs.releasetag }}'```

.github/workflows/cd-badger.yml
sanjayk-github-dev: Where is the tag getting applied? I don't see that step in the workflow.

.github/workflows/cd-badger.yml
skrdgraph: +1 ^

.github/workflows/cd-badger.yml
skrdgraph: @sanjayk-github-dev - 

the process is as follows:
- cut a tag
- enter the tag from step(1) to trigger the cd pipeline

If I have understood your question correctly - I have not automated the auto-trigger on tag because we have to figure out a process to limit tag creators. If someone accidentally adds a tag, that could trigger the pipeline. The manual step ensures we are safe today. This can evolve.

sanjayk-github-dev: I was wondering if the tag can be created as a step in the workflow once the workflow is successfully completed.

joshua-goldstein (approved):

.github/workflows/cd-badger.yml
skrdgraph: Sticking to tag based model, discussed over a call.

PR #1888, opened 2023-02-27
fix(pb): fix generated protos

joshua-goldstein:
We update protobuf files and generation script to version v4.


skrdgraph (approved):
👍

meghalims (approved):
LGTM.

PR #1890, opened 2023-02-28
chore(ci): fail fast when testing

joshua-goldstein:
Add failfast flag to `go test` commands in test.sh


skrdgraph (approved):
This looks good - because the CI was failing silently before. 

With the fixes we get an actual failure (which is what we want).

PR #1891, opened 2023-02-28
chore(go.mod): add retraction

joshua-goldstein:
Since the v4.0.0 tag cannot be removed from the goproxy service, we added a retraction directive so users do not accidentally use v4.0.0 tag.


skrdgraph (approved):
Looks good 👍

skrdgraph (approved):

PR #1894, opened 2023-03-01
chore(ci): use pull request target

joshua-goldstein:
Secrets are not accessible to community PR's.  This is a problem when reporting code coverage, since our goveralls token is stored as a secret, resulting in CI failure.  See discussion https://github.com/dependabot/dependabot-core/issues/3253 for workaround.  More details about permissions here: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

### DIFF
--- a/.github/workflows/ci-badger-tests.yml
+++ b/.github/workflows/ci-badger-tests.yml
@@ -4,7 +4,7 @@ on:
     branches:
       - main
       - 'release/v*'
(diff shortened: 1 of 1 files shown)


joshua-goldstein:
Issue we filed https://github.com/coverallsapp/github-action/issues/150 here is resolved now

PR #1896, opened 2023-03-01
chore(ci): remove unnecessary token

joshua-goldstein:
We added a token because we thought there was a permission issue, however the issue was the action secrets were not available to community PR's.  See discussion https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544


skrdgraph:
👍 looks good

PR #1898, opened 2023-03-02
feat(txn): Add PreCommit and PreCommitAt methods

protosam:
## Problem
There is no way to perform a 2 phase commit with badger directly. This PR adds pre-commit methods for 2 phase commits.

## Solution
The new `PreCommit` and `PreCommitAt` methods added to `Txn` performs the validation portion of a commit for observability without needing to write the commit.

### DIFF
--- a/managed_db.go
+++ b/managed_db.go
@@ -88,12 +88,17 @@ func (db *DB) SetDiscardTs(ts uint64) {
 	db.orc.setDiscardTs(ts)
 }
 
-// CanCommitAt will return true if commit will succeed at specified commit timestamp
-func (txn *Txn) CanCommitAt(commitTs uint64) bool {
+// PreCommitAt will perform prevalidation for CommitAt and return true if commit the transaction can be committed.
+func (txn *Txn) PreCommitAt(commitTs uint64) bool {
 	if !txn.db.opt.managedTxns {
 		panic("Cannot use CanCommitAt with managedDB=false. Use CanCommit instead.")
 	}
+
 	txn.commitTs = commitTs
 
-	return !txn.db.orc.hasConflict(txn)
+	txn.db.orc.writeChLock.Lock()
+	defer txn.db.orc.writeChLock.Unlock()
+	txn.precommitTs, txn.precommitConflict = txn.db.orc.newCommitTs(txn)
+
+	return !txn.precommitConflict
 }
--- a/managed_db.go
+++ b/managed_db.go
@@ -87,3 +87,13 @@ func (db *DB) SetDiscardTs(ts uint64) {
 	}
 	db.orc.setDiscardTs(ts)
 }
+
+// CanCommitAt will return true if commit will succeed at specified commit timestamp
+func (txn *Txn) CanCommitAt(commitTs uint64) bool {
+	if !txn.db.opt.managedTxns {
+		panic("Cannot use CommitAt with managedDB=false. Use Commit instead.")
+	}
+	txn.commitTs = commitTs
+
+	return !txn.db.orc.hasConflict(txn)
+}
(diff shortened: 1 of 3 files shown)


mangalaman93:
Hi there, Thanks for the PR. Could you provide us more details about your use case? Maybe point to any code which is using these functions.

mangalaman93:
@protosam I was wondering if you are looking to use badger where you'd need these functions. I would prefer to add these functions for a use case out there instead of a made up (even when it seems logical) scenario. This will ensure that the functions are useful in at least one scenario.

protosam:
Yes, I'm actively working on an unreleased project that will be using these functions.

protosam:
As an update, this is being tested in a privately funded project. To my surprise, it seems to be working so far. It would great if it could be iterated on and contributed back to badger, but I can just maintain a fork if this doesn't seem worth-while to add upstream. Just let me know if there's a hard rejection on it.

Using Lamport clocks with `PreCommitAt` I was able to make a transaction coordinator with very little effort. The only alternative to this I could devise was maintaining what equates to being a distributed WAL (for partitioned ranges). I expect the 2PC process to require less up front cost and paralelizes better than the WAL would.

mangalaman93:
I see, thanks for more details. I do see the value of these functions for 2 PC, though, I want to be careful adding new APIs in the library. It is easy to add more APIs and painful and change or modify them. Give me some time to think about and let me figure out what we need in order to get this change in.

protosam:
Thanks for your feedback. The use case for 2PC is pretty specific to coordinating transactions between nodes. Perhaps an assumption can be made that having this validator is only useful for managed transactions?

I think this might limit the scope to adding `PreCommitAt` and might not need all the changes that I had to add for `PreCommit`.

managed_db.go
```diff
@@ -87,3 +87,18 @@ func (db *DB) SetDiscardTs(ts uint64) {
...
+func (txn *Txn) PreCommitAt(commitTs uint64) bool {
+	if !txn.db.opt.managedTxns {
+		panic("Cannot use CanCommitAt with managedDB=false. Use CanCommit instead.")
+	}
+
+	txn.commitTs = commitTs
+
+	txn.db.orc.writeChLock.Lock()
+	defer txn.db.orc.writeChLock.Unlock()
+	txn.precommitTs, txn.precommitConflict = txn.db.orc.newCommitTs(txn)
```
mangalaman93, line 101: calling `newCommitTs` here is pointless because we have already set `precommitTs` and it will mostly be non zero

mangalaman93 (changes requested):
I think what you are trying to do is decide on a commit timestamp and whether transaction can be committed or not. But txn is not written to disk until Commit or CommitAt is called. That seems reasonable I think.

I think we should test whether everything is fine in non-managed mode when PreCommit is called but Commit is not called. I want to make sure that badger is not unhappy with not finding that txn for which it has assigned a timestamp later.

mangalaman93:
I am fine with the API. I will ask other engineers to take a look at this PR too. These are the two things we need now before we can merge this change:
1. a test that does lots of writes and uses PreCommit/PreCommitAt. Discards some of the writes and Commits rest of them. Now, when we do the read, we should only find the writes that were committed and not find the writes that were only PreCommitted.
2. Cleanup the PR, and make sure you are happy with the change.

Thanks again for all the work.

protosam:
These functions don't work as intended and I'm not sure why. I ended up having to abandon this implementation and work around the issue with a group based paxos implementation.

Going to close this. The code is licenses as-is from to PR if anyone wants to use and advance it in the future for some reason.

mangalaman93:
@protosam could you elaborate what didn't work? An example would be very useful.

PR #1901, opened 2023-03-07
fix(stream): improve incremental stream writer

mangalaman93:
This commit fails incremental stream writer if memtable has data. It also allows calling incremental stream store more times than the number of levels we have. This is done by calling Flatten when we have reached L0 and incremental stream writer can be used again.

### DIFF
--- a/stream_writer.go
+++ b/stream_writer.go
@@ -104,6 +104,14 @@ func (sw *StreamWriter) PrepareIncremental() error {
 	}
 	sw.done = func() { once.Do(done) }
 
+	mts, decr := sw.db.getMemTables()
+	defer decr()
+	for _, m := range mts {
+		if !m.sl.Empty() {
+			return fmt.Errorf("Unable to do incremental writes because MemTable has data")
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


joshua-goldstein (dismissed):
Looks good!  Let's get https://github.com/dgraph-io/badger/pull/1908 in first so Dgraph tests run against this (don't think there is an issue but it will be good to see the workflow run).

stream_writer.go
```diff
@@ -117,7 +125,10 @@ func (sw *StreamWriter) PrepareIncremental() error {
 		return nil
 	}
 	if sw.prevLevel == 0 {
-		return fmt.Errorf("Unable to do incremental writes because L0 has data")
+		if err := sw.db.Flatten(3); err != nil {
```
harshil-goel, line 128: What does this do? Can you write a comment about how we used to do it before, and why its okay now to flatten instead of cancelling here?

mangalaman93: Done.

harshil-goel (approved):

PR #1908, opened 2023-03-15
chore(ci): add dgraph regression tests

joshua-goldstein:
Run dgraph tests when pushes are made to Badger as sanity check.

### DIFF
--- a/.github/workflows/ci-dgraph-tests.yml
+++ b/.github/workflows/ci-dgraph-tests.yml
@@ -1,0 +1,67 @@
+name: ci-dgraph-tests
+on:
+  push:
+    branches:
+      - main
+      - 'release/**'
+jobs:
+  dgraph-tests:
+    runs-on: [self-hosted, x64]
+    steps:
+      - name: Checkout Dgraph repo
+        uses: actions/checkout@v3
+        with:
+          repository: dgraph-io/dgraph
+          ref: main
+      - name: Get Go Version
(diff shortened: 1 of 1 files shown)


joshua-goldstein:
@skrdgraph We will have to attach a runner to the Badger repo for this to work, what are your thoughts?  My thinking here is that this should never fail and this is a safeguard against releasing changes that break Dgraph tests.  Ideally this should never happen (which is why it is only a sanity check that runs on pushes to main), but we saw it once before, so it would be nice to have CI catch this problem first.

.github/workflows/ci-dgraph-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+          #!/bin/bash
+          GOVERSION=$({ [ -f .go-version ] && cat .go-version; })
+          echo "GOVERSION=$GOVERSION" >> $GITHUB_ENV
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GOVERSION }}
+      - name: Fetch latest Badger version
+        run: |
+          go get github.com/dgraph-io/badger/v4@main
```
joshua-goldstein, line 26: Trick is here: we are updating dep in dgraph to use latest version of badger on main branch

mangalaman93 (approved):

karbonnanroid (approved):

PR #1914, opened 2023-03-27
fix(docs): Update getting started docs for v4 release

evanlurvey:
I took out the chunk about versioning since going to standard SEMVER and I removed the thing about CGO because its not longer relevant. 

I read the change log before going to the docs page and noticed the discrepancies.


mangalaman93 (approved):

evanlurvey:
I need to update the badger cli to use version 4. I am trying to make this change but it seems github broke pushes according to down detector. https://downdetector.com/status/github/

evanlurvey:
Was able to push and update this. Should be good to go

PR #1924, opened 2023-04-12
fix(test): add missing unlock in TestPersistLFDiscardStats

joshua-goldstein:
## Description

`TestPersistLFDiscardStats` is currently flaky.  E.g., [1](https://github.com/dgraphio/badger/actions/runs/4662647309/jobs/8253226059), [2](https://github.com/dgraph-io/badger/actions/runs/4660157619/jobs/8247868522), [3](https://github.com/dgraph-io/badger/actions/runs/4659474072/jobs/8246402042) (but this goes back further).  To reproduce the problem:
```
go test -run=TestPersistLFDiscardStats -count=80 -timeout=1500s
```

From the stack traces in the logs, this looks like a deadlock.  The test currently sleeps for 2 seconds to allow for compaction to complete.  Assuming compaction does not complete in this time, and we make the call to `db.vlog.discardStats.Lock()`, seems to be enough to put us into this deadlock.  This is supported experimentally; running the above test command with this change no longer produces the deadlock.

There is another sleep in the test to give discardStats a chance to be populated by the newly generated discardStats.  We have not seen this issue in CI, but when running this test at very high frequencies (i.e. with `-count=160`), we start to see this failure as well.  This is mitigated by increasing the sleep.

### DIFF
--- a/value_test.go
+++ b/value_test.go
@@ -522,7 +522,7 @@ func TestPersistLFDiscardStats(t *testing.T) {
 		require.NoError(t, err)
 	}
 
-	time.Sleep(4 * time.Second) // wait for compaction to complete
+	time.Sleep(2 * time.Second) // wait for compaction to complete
 
 	persistedMap := make(map[uint64]uint64)
 	db.vlog.discardStats.Lock()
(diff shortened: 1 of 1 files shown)


value_test.go
```diff
@@ -531,14 +531,15 @@ func TestPersistLFDiscardStats(t *testing.T) {
 		persistedMap[fid] = val
 	})
 
+	db.vlog.discardStats.Unlock()
```
mangalaman93, line 534: I find it odd that we were acquiring the lock before and not calling unlock. Could you elaborate more on this?

joshua-goldstein: I also found that very odd.  I have no insight into why it was like this.  Probably an oversight.

mangalaman93: Seems like `db` variable is re-assigned and hence calling `Lock` again without `Unlock` is not a problem. But this is better.

mangalaman93:
@joshua-goldstein could you explain why would there be a deadlock if we don't put sleep altogether?

mangalaman93 (approved):

PR #1928, opened 2023-04-19
Close memtable wal once we are done writing

harshil-goel:


kevinmingtarja:
Should/can we also close the memtable wal right after the current memtable `db.mt` is sent to `db.flushChan` and `db.mt` is appended to `db.imm`?

Because at this point, there will be no more writes happening to that memtable wal right? We'll need to make sure though that there are no more reads happening to the wal else we will get a segfault. But i imagine the only reads happening to an immutable memtable's wal would be on restart?

PR #1932, opened 2023-04-27
fix(level): change split key range right key to use ts=0

hugy718:
## Problem
This old implementation is not incorrect, but it does not seem to follow the logic in its description. When a compaction is split into subcompaction jobs, each of them builds an iterator on all tables touched by the compaction and iterates over a certain `keyRange` defined in `addSplits`. In combination, they covers the key range of all the tables in the compaction. The right key of a key range is intended to be the right key of a bottom level table (`cd,bot`), It is intended to include all different versions of that key, as described in the comments. However, using `math.MaxUint64` will exclude those keys from this `keyRange` and include them in the next split. The reason is that the timestamp is encoded as `math.MaUint64-ts` in `y.KeyWithTs`, so a key with larger ts is actually smaller in `y.CompareKeys`. It can be corrected by using ts=0. Note that even using ts=math.MaxUint64 is not going to drop keys unlike what the comments above suggest. because those keys are covered in the subsequent split and the last split have an empty right key, iterating till the end of the tables being compacted.

## Solution
Changed the timestamp used for split key range right key from `math.MaxUint64` to `0`. Updated the comments above it.

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -1073,14 +1073,13 @@ func (s *levelsController) addSplits(cd *compactDef) {
 			return
 		}
 		if i%width == width-1 {
-			// Right should always have ts=maxUint64 otherwise we'll lose keys
-			// in subcompaction. Consider the following.
+			// Right is assigned ts=0. The encoding ts bytes take MaxUint64-ts,
+			// so, those with smaller TS will be considered larger for the same key.
+			// Consider the following.
 			// Top table is [A1...C3(deleted)]
 			// bot table is [B1....C2]
-			// This will generate splits like [A1 ... C2] . Notice that we
-			// dropped the C3 which is the last key of the top table.
-			// See TestCompaction/with_split test.
-			right := y.KeyWithTs(y.ParseKey(t.Biggest()), math.MaxUint64)
+			// It will generate a split [A1 ... C0], including any records of Key C.
+			right := y.KeyWithTs(y.ParseKey(t.Biggest()), 0)
 			addRange(right)
 		}
 	}


mangalaman93:
Hi @hugy718, thank you for the PR. Do you think it is possible to add a test for this?

hugy718:
`TestCompaction/with_split` test in `level_test.go` covers the correctness of using `addSplits` to run multiple subcompactions

mangalaman93:
@hugy718 I am wondering whether it is easy to add a test that shows that things could go wrong without this change and this change fixes the issue? I can also look into it myself if you think it's not easy to come up with a test.

hugy718:
Actually, nothing would go wrong without this change. It only improves the implementation such that keys in the same bottom table will not span two splits, which is what the old code comments there suggested.

mangalaman93:
yes, you are right. I will let @harshil-goel review it and merge it. Thanks for the PR again.

harshil-goel:
@hugy718 Thanks a lot for the PR. I went through the logic, and the old PR: https://github.com/dgraph-io/badger/pull/1587/files. From what I understand, what's happening is that
1. We need all the versions of a key, in the same split. This would ensure that a key's delete is handled properly. 
2. Earlier, we were just creating splits by the bottom table right most element. That led to creating two splits, one contained C2, other C3. This leads to C2 being still there, but C3 being deleted. (example from the comments in the earlier code). 
3. With the old fix, we would have still created two splits, [A1 ... B] [C2...C3]. With your change, we would be creating something like, [A1...B...C2..C3].
Is my understanding correct?

hugy718:
C3 is placed before C2, since the encoding of time stamp uses maxuint64 to minus it. 

The old approach is very unlikely give meaningful extra parallelism, since it only affects the records of the same key. However, the old approach places a bottom table into two splits. The old example actually is not so good in explaining the resulting splits, since there is a condition above that return just one split. https://github.com/dgraph-io/badger/blob/9afd0a09352373345d66f73abf5adf4ec877d08f/levels.go#L1071-L1074

If we consider a compaction that truly uses multi-split subcompaction, there are much more bottom tables. We consider the examples in the comment L1049. https://github.com/dgraph-io/badger/blob/9afd0a09352373345d66f73abf5adf4ec877d08f/levels.go#L1049-L1054 The old approach does not create splits that have clean boundary that align with a bot table boundary, so the splits for the tables would be: [t0, t1, t2] [t2, t3, t4, t5] [t5, t6, t7, t8] [t8, t9]. The fix would create the desirable splits of [t0, t1, t2] [t3, t4, t5] [t6, t7, t8][t9].

harshil-goel:
From what I understand, you are saying that a bottom table is only in one subcompaction job. However, with this change, if the key is split into multiple tables (C2 in table 3, C3 in table 4, C4 in top level), then the table 4 would be in 2 different subcompaction job right?
It doesn't make sense to merge the change if it doesn't provide any performance / correctness benefit. However, this diff made us go deeper into some topics. You have done a really great work in understanding badger. We would be happy if you just want to change the comment to make it more understandable / correct.

hugy718:
All the tables mentioned in the second example are bottom tables because `addsplits` only consider the bottom tables then the iterator will iterate through all records in that key range. A top table can definitely be involved in two splits subcompaction. The case you raise that the different versions of the same key belongs to different tables in a bottom level, however, will never happen, because these tables are generated by compaction and the `addKeys` lambda in `subcompact` ensures that adjacent tables produced by a compaction will never have different versions of the same key. (i.e. it only checks the conditions that break the add key looping whenever the iterator sees a different key.) Therefore, a bottom table will only be found in one split using the fix.

Performance-wise, there is unlikely improvement because block cache is used most of the time and the last block of a bottom table referenced by two splits (processed by two separate goroutine) is likely in-cache when the slower split needs it. But when block cache is disabled, by placing that block into one split, it avoids both goroutine fetching that block from storage, reducing 1 IO, but that would not be significant since a compaction access hundreds of blocks.

If you still feel that it is easier to keep the old way, I will then revert back and only correct those code comments describing the old way.

harshil-goel:
It makes sense, that we don't have keys spilling over to new tables. In that case it would make sense to do the changes. Thanks a lot for the diff, I have accepted it.

PR #1934, opened 2023-05-03
fix(PageBufferReader): PageBufferReader should conform to Reader interface

billprovince:
## Problem
 Fixes DGRAPHCORE-121
The CI build was showing sporadic failures of TestPagebufferReader2 with the error message:
          Received unexpected error:  EOF
 The result is sporadic because the test case relied on randomized behavior. In particular, it would
generate a read-buffer of some random length, which could occasionally have a length of 0.
When the length is 0, we would encounter this error.

The cause is from the PageBufferReader having incorrect behavior for the Read(p []byte) function: In particular, when p is empty, based on the expected behavior of the Reader interface, this should return 0, nil. However, this is currently returning 0, EOF.
 

## Solution
First:
I added a unit test to make sure we consider the empty buffer input in every case (instead of just randomly every once in a while). This doesn't actually provide a solution, but it makes sure that we don't end up with a regression in this area.
Second:
At the point where we currently detect that we read 0 bytes, before just returning 0, EOF, I add a check to make sure that the value p has non-zero length.


mangalaman93:
@billprovince I think you need to branch off of our latest main. Could you do that?

billprovince:
> @billprovince I think you need to branch off of our latest main. Could you do that?

Kind of odd: I did a git pull of main branch prior to this. Highly unclear how this got the wrong set of changes. Perhaps I inadvertently pulled some other changes???  I'm going to kill this PR, redo my git clone of badger (i.e., completely wipe out my local install version and rebuild from github), redo this PR, then try again.

billprovince:
Killing this PR (replaced with fixPageBufferReaderTry2).

PR #1936, opened 2023-05-04
fix(test): The new test case PagebufferReader5 introduced an error.

billprovince:
## Problem
Fixes DGRAPHCORE-199
The intended fix for the CI issue for randomly getting EOF failures in PagebufferReader2 test introduced the test case FixPagebufferReader5 -- intended to avoid regressions. However, it did something that is not expected by the PageBuffer struct: it created a new PageBufferReader based on an empty PageBuffer. It seems that this is not allowed.
 
## Solution
Fix the test case for PagebufferReader5 by no longer attempting to create a new PageBufferReader based on an empty PageBuffer.

### Note
The CI build will still fail to allow this to pass because it will continue to use the prior version of the tests without the fix. However, locally tested, I get this result:
~/gitspace/github.com/dgraph-io/badger/y$ go test -run PagebufferReader5
PASS
ok  	github.com/dgraph-io/badger/v4/y	0.335s

### DIFF
--- a/.github/workflows/ci-badger-tests.yml
+++ b/.github/workflows/ci-badger-tests.yml
@@ -4,7 +4,7 @@ on:
     branches:
       - main
       - 'release/v*'
-  pull_request_target:
+  pull_request:
     branches:
       - main
       - 'release/v*'
--- a/.github/workflows/ci-badger-tests.yml
+++ b/.github/workflows/ci-badger-tests.yml
@@ -4,7 +4,7 @@ on:
     branches:
       - main
       - 'release/v*'
-  pull_request:
+  pull_request_target:
     branches:
       - main
       - 'release/v*'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-badger-tests.yml
```diff
@@ -4,7 +4,7 @@ on:
     branches:
       - main
       - 'release/v*'
-  pull_request_target:
+  pull_request:
```
skrdgraph, line 7: hey @billprovince I made this fix to your PR. 

There are 2 issues with our CI here:
- `pull_request_target` is what is causing this issue, and this was added per recommendation from coveralls to support coverage
- `cache` cleanup was also done on GitHub Actions (as we suspected an issue there).

cc: @adityasadalage and @joshua-goldstein ... lets do a check on these items tomorrow.

joshua-goldstein (dismissed):

.github/workflows/ci-badger-tests.yml
joshua-goldstein: Thanks for catching this and diagnosing the problem...  I suggest we merge the CI fix in https://github.com/dgraph-io/badger/pull/1944 first, and then we can merge Bill's PR (and hopefully CI will run as expected on here)

joshua-goldstein (approved):

PR #1944, opened 2023-05-05
chore(ci): split off coverage workflow

joshua-goldstein:
In https://github.com/dgraph-io/badger/pull/1936 we observed that because we are using `pull_request_target` without a specific reference, we are not in fact checking out the PR branch.  This PR splits the workflow into two, one only to run badger CI tests, and the other to run tests & report coverage results (using pr_target with the appropriate commit).  This has the benefit that any issues with coverage are isolated to one workflow, which we could disable later if problems arise.  This does create duplication of compute (all tests run x2) but we are using github hosted runners so we do not pay for this cost.  This also aligns with our dgraph repo workflow setup.

### DIFF
--- a/.github/workflows/ci-badger-tests-coverage.yml
+++ b/.github/workflows/ci-badger-tests-coverage.yml
@@ -1,20 +1,16 @@
-name: ci-badger-tests
+name: ci-badger-tests-coverage
 on:
-  push:
+  pull_request_target:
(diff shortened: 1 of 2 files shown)


skrdgraph (dismissed):

skrdgraph (approved):
Looks good 👍

PR #1949, opened 2023-05-16
chore(build): dockerfile and devcontainer

Mr-Destructive:
## Problem
Some of the common problems for the development of the package might be:

- Need to work with the badger repository/code base, but don't have a golang environment setup. 
- No container available for the badger package.
- Create a Web-based or Cloud-based development environment.

 

## Solution
- Added a `Dockerfile` that will create the base image for the badger package/cli.
- Added `.devcontainer` folder for GitHub codespaces, users can click on `Code` and it spins up a web-based editor with the base setup of the package (it can also re-build after making changes).
- Added a `.gitpod.yaml` file for creating a workspace on Gitpod.io, it's a cloud-based development environment.
 

Let me know if these changes are valid. I would love to work further as per your guidance and feedback.
Thank you.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,5 @@
+{
+  "build": {"dockerfile": "../Dockerfile"},
+  "extensions": ["golang.go"],
+  "forwardPorts": [8080]
+}
--- a/.gitpod.yaml
+++ b/.gitpod.yaml
@@ -1,0 +1,19 @@
+image: golang:1.19-alpine
+
+workspace:
+  name: badger-dev
+  path: /app
+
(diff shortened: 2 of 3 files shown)


mangalaman93:
> Containerization can make it easier to upgrade Badger without restarting Dgraph, simplify configuration management, and facilitate running multiple instances of Badger in parallel. This all can make development a bit smoother and more friendly. While it's not a strict requirement, having a Dockerfile for Badger can provide these advantages when needed.

Thanks for the PR. I agree with Joshua here. Because badger is an embedded library, I do not think it would be possible to upgrade badger without restarting Dgraph. Closing the PR for now.

PR #1951, opened 2023-05-17
fix(table): replace unrecoverable panic with error

AveralS:
## Problem

I am using badger to store some temporary transactional data. My database crashed due to power failure and when the app starts again it panics due to the broken badger database. I can't handle this panic, because it is called in a separate goroutine.

## Solution

[initBiggestAndSmallest method](https://github.com/dgraph-io/badger/blob/ef0e55289cf70ade33f1c1dd19dcd8eced85cbbb/table/table.go#L346) should return an error when recovering from underlying panic.


mangalaman93 (dismissed):

mangalaman93:
@AveralS Thanks for the PR. Could you please sign the CLA?

AveralS:
> @AveralS Thanks for the PR. Could you please sign the CLA?

No, i dont see accept button here

image

mangalaman93:
That's pretty odd. You could try a different browser.

table/table.go
```diff
@@ -343,14 +343,14 @@ func OpenInMemoryTable(data []byte, id uint64, opt *Options) (*Table, error) {
 	return t, nil
 }
 
-func (t *Table) initBiggestAndSmallest() error {
+func (t *Table) initBiggestAndSmallest() (err error) {
```
harshil-goel, line 346: Can you remove the named return variable?

Issue #1847, opened 2023-01-18
[BUG]: *DB.Sync should sync memtable in addition to value log


mcfog:
### What version of Badger are you using?

github.com/dgraph-io/badger/v3 v3.2103.5

### Have you tried reproducing the issue with the latest release?

Yes

### What is the hardware spec (RAM, CPU, OS)?

-

### What steps will reproduce the bug?

call *DB.Sync()

### Expected behavior and actual result.

Both memtable('s wal file) and value log should be persisted by Sync call.

mangalaman93:
This seems reasonable to me. If you could raise a PR with a test that proves the issue, that'd be awesome. Otherwise, we will try to get to it as soon as we can. Thanks

PR #1953, opened 2023-05-22
Sync active memtable and value log on Db.Sync (#1847)

SarthakMakhija:
Fixes https://github.com/dgraph-io/badger/issues/1847

Currently, `DB.Sync()` only syncs the value log but not the WAL of the active memtable, however recovery happens from the WAL of the active memtable as a part of the `DB.Open()` method.

This change attempts to sync both the logs, however there can be an issue in syncing one of the logs. This change lists all of the possible cases that can arise during the sync operations. Some of these issues will be taken separately. For example, 
the issue https://github.com/dgraph-io/badger/issues/1954 can arise and shall be handled separately.

This change adds a few tests to ensure that the Sync behavior is not broken.  This change also adds a learning test `(db2_test.go -> TestAssertValueLogIsNotWrittenToOnStartup)` to ensure that value log is only read from (and not written to) during startup. This learning shall be used in the next issue (the one described above (*))

### DIFF
--- a/db.go
+++ b/db.go
@@ -676,7 +676,43 @@ const (
 // Sync syncs database content to disk. This function provides
 // more control to user to sync data whenever required.
 func (db *DB) Sync() error {
-	return db.vlog.sync()
+	/**
+	Make an attempt to sync both the logs, the active memtable's WAL and the vLog (1847).
+	Cases:
+	- All_ok			:: If both the logs sync successfully.
+
+	- Entry_Lost		:: If an entry with a value pointer was present in the active memtable's WAL,
+						:: and the WAL was synced but there was an error in syncing the vLog.
+						:: The entry will be considered lost and this case will need to be handled during recovery.
+
+	- Entries_Lost		:: If there were errors in syncing both the logs, multiple entries would be lost.
+
+	- Entries_Lost      :: If the active memtable's WAL is not synced but the vLog is synced, it will
+						:: result in entries being lost because recovery of the active memtable is done from its WAL.
+						:: Check `UpdateSkipList` in memtable.go.
+
+	- Nothing_lost		:: If an entry with its value was present in the active memtable's WAL, and the WAL was synced,
+						:: but there was an error in syncing the vLog.
+						:: Nothing is lost for this very specific entry because the entry is completely present in the memtable's WAL.
+
+	- Partially_lost    :: If entries were written partially in either of the logs, the logs will be truncated during recovery.
+						:: As a result of truncation, some entries might be lost. Assume that 4KB of data is to be synced and
+					    :: invoking `Sync` results only in syncing 3KB of data and then the machine shuts down or the disk failure happens,
+	                    :: this will result in partial writes. [[This case needs verification]]
+	*/
+
+	var err error
+
+	memtableSyncError := db.mt.SyncWAL()
+	if memtableSyncError != nil {
+		err = errors.Wrap(memtableSyncError, "sync failed for the active memtable WAL")
+	}
+
+	vLogSyncError := db.vlog.sync()
+	if vLogSyncError != nil {
+		err = errors.Wrap(vLogSyncError, "sync failed for the vLog")
+	}
+	return err
 }
 
 // getMemtables returns the current memtables and get references.
--- a/db.go
+++ b/db.go
@@ -695,19 +695,21 @@ func (db *DB) Sync() error {
 						:: but there was an error in syncing the vLog.
 						:: Nothing is lost for this very specific entry because the entry is completely present in the memtable's WAL.
 
-	- Partially_lost    :: If entries were written partially in either of the logs, the logs will be truncated during recovery.
-						:: As a result of truncation, some entries might be lost. Assume that 4KB of data is to be synced and
-					    :: invoking `Sync` results only in syncing 3KB of data and then the machine shuts down or the disk failure happens,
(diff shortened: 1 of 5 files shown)


db.go
```diff
@@ -676,7 +676,43 @@ const (
...
+
+	- Partially_lost    :: If entries were written partially in either of the logs, the logs will be truncated during recovery.
+						:: As a result of truncation, some entries might be lost. Assume that 4KB of data is to be synced and
+					    :: invoking `Sync` results only in syncing 3KB of data and then the machine shuts down or the disk failure happens,
+	                    :: this will result in partial writes. [[This case needs verification]]
+	*/
+
+	var err error
+
+	memtableSyncError := db.mt.SyncWAL()
```
mangalaman93, line 706: we probably should acquire a lock here

db.go
```diff
@@ -676,7 +676,43 @@ const (
...
+						:: Nothing is lost for this very specific entry because the entry is completely present in the memtable's WAL.
+
+	- Partially_lost    :: If entries were written partially in either of the logs, the logs will be truncated during recovery.
+						:: As a result of truncation, some entries might be lost. Assume that 4KB of data is to be synced and
+					    :: invoking `Sync` results only in syncing 3KB of data and then the machine shuts down or the disk failure happens,
+	                    :: this will result in partial writes. [[This case needs verification]]
+	*/
+
+	var err error
+
```
mangalaman93, line 705: don't need a newline here

db.go
```diff
@@ -676,7 +676,43 @@ const (
...
+	memtableSyncError := db.mt.SyncWAL()
+	if memtableSyncError != nil {
+		err = errors.Wrap(memtableSyncError, "sync failed for the active memtable WAL")
+	}
+
+	vLogSyncError := db.vlog.sync()
+	if vLogSyncError != nil {
+		err = errors.Wrap(vLogSyncError, "sync failed for the vLog")
```
mangalaman93, line 713: does it make sense to combine this error with `memtableSyncError`?

mangalaman93 (changes requested):
I just realized that we may need to consider acquiring some lock while accessing the memtable and vlog. Please look into it. Also, address the comments from the CI.

db.go
SarthakMakhija: Sure, but why was the original sync of vlog without locks?

db.go
SarthakMakhija: done

db.go
SarthakMakhija: The idea is to perform a sync of vlog even if there is an error in syncing active memtable WAL. With this approach, I can return only one error.. hence, it is combined with memtableSyncError. 

We have very limited options:

1) If the sync of active memtable fails, do not perform vlog sync. So, no need to combine errors
2) Perform a sync of both the logs, combine errors
3) Perform a sync of both the logs, return a custom error struct

SarthakMakhija:
> I just realized that we may need to consider acquiring some lock while accessing the memtable and vlog. Please look into it. Also, address the comments from the CI. CI also is not happy.

Linting part is ok, I will reduce the line length in the comment. 
Pipeline is failing because of the coverage
```
FAIL
coverage: 64.8% of statements
```
I am not sure if I can cover this part.

db.go
mangalaman93: option 2 is the best. We are still on Go 1.19, on Go 1.20 we could have used multi error.

SarthakMakhija: Yes, I originally used multi error but realized badger is not on golang 1.20. So, used `errors.Wrap` and combined both the errors.

db.go
SarthakMakhija: For now, I have changed the code to use lock (`the write lock`).

mangalaman93: I think for vlog, we acquire the lock inside the sync function. We need to do something for memtable.

mangalaman93:
> I am not sure if I can cover this part.

Don't worry about it for now.

db.go
SarthakMakhija: Ok, wait. Let me change the code to get `RLock` and unlock before I sync vlog

db.go
SarthakMakhija: No, you are right. Corrected the code and added tests to checking error combination.

mangalaman93 (commented):
I am happy with the change now. I will let @joshua-goldstein approve it for merging after review. Thanks for the PR Sarthak.

y/error_test.go
```diff
@@ -0,0 +1,27 @@
+package y
+
+import (
+	"errors"
+	"github.com/stretchr/testify/require"
```
mangalaman93, line 5: this import order has to look like this:

"errors"
"testing"

"github.com/stretchr/testify/require"

SarthakMakhija: Why?

SarthakMakhija: Strange rule to place in lint. I am not even sure what is the value add of the rule.

SarthakMakhija: done

mangalaman93: We do it to separate standard lib imports from external imports. I have always done this for as long as I have been writing Go code. It is a lot nicer and easier to read this way I believe.

SarthakMakhija: Ok, I let goland do this for me.

joshua-goldstein (approved):

mangalaman93 (approved):

PR #1963, opened 2023-06-05
fix(test): fix flakiness of TestPersistLFDiscardStats

billprovince:
## Problem
Fixes DGRAPHCORE-234

The purpose of TestPersistLFDiscardStats is to make sure that if you make changes to the file such that we need to maintain the discardStats in the value log, and if you close the DB, then when you reopen the DB, you remember the
same status.

Note that the DiscardStats are updated when we perform compaction, and in our test cases we have 4 compactors
concurrently running (with ids 0, 1, 2, and 3). Most of the time, we were fine when we captured a single compaction
cycle -- and then closed the DB and then reopened. However, there was a race condition wherein we waited for some
arbitrary amount of time, then captured the current discardStats, then closed the DB. The problem is that between
the time that we capture the discardStatus and close the DB, another compaction cycle may have started!

Hence, every once in a while, we would find that the saved copy of the discard stats would not match what we
picked up later when we reopened the file.

## Solution
As noted in the problem statement, we ended up waiting an "arbitrary amount of time" instead of waiting for specific
events and then reacting to those events. Specifically, we wanted to wait until at least "some stats" had been generated,
and then we waited for compaction to complete. Unfortunately, there did not exist a clear way (previously) to capture the
event of "some stats having been generated", nor was there a way to capture the discardStats upon closure of the
database.

The solution then was to add in two things: first, a test channel in the database where we can log messages to this
channel, but only when the channel has been specified. Second, we add in a means (via options.go) of specifying
an "onCloseDiscardCapture" map. This map will be populated (assuming it was initialized and is not nil) when we
close the db and specifically when we close the valueLog.

We no longer rely on time.Sleep, but instead rely on specific events.


db.go
```diff
@@ -124,6 +124,8 @@ type DB struct {
 	blockCache *ristretto.Cache
 	indexCache *ristretto.Cache
 	allocPool  *z.AllocatorPool
+
+	testChan chan string
```
mangalaman93, line 128: we can choose a better name so that we could use it in future for other tests too

db.go
```diff
@@ -397,6 +400,17 @@ func Open(opt Options) (*DB, error) {
 	return db, nil
 }
 
+// logToTestChan sends a message to the DB's testChan. Note that we expect
+// that the DB *never* closes this channel; the responsibility for
+// allocating and closing the channel belongs to the test module.
+// if db.testChan is nil or has never been initialized, this will
+// be silently ignored.
+func (db *DB) logToTestChan(msg string) {
```
mangalaman93, line 408: This function should have names based on what it does, something notifyUpdateDiscardStats or something

options.go
```diff
@@ -127,6 +127,12 @@ type Options struct {
 	maxBatchSize  int64 // max batch size in bytes
 
 	maxValueThreshold float64
+
+	testChan chan string
```
mangalaman93, line 131: what is this used for compared to onCloseDiscardCapture? Can we not just live with one var?

mangalaman93 (changes requested):

options.go
```diff
@@ -360,6 +366,26 @@ func (opt Options) FromSuperFlag(superflag string) Options {
 	return opt
 }
 
+// withTestChan returns a new Options value with testChan set to given value.
+//
+// testChan is used only for being able to coordinate tests operations.
+// Internally, access to the testChan will be nil-protected. We can safely
+// rely on branch-prediction optimization for performance, so adding this
+// will have negligible performance impact.
+func (opt Options) withTestChan(testChan chan string) Options {
```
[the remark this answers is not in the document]
billprovince: What then would be your counter-proposal to get the desired result?

db.go
billprovince: I suspect that you mean something other than "for other tests", since otherwise, the name "testChan" would suffice. What is your intent here?

On the Zoom call, we discussed the name "syncChannel" (or syncChan to be a bit more succinct).

options.go
billprovince: testChan is used currently for two messages: 
1. The end of vlog.init(db)  (see line 558 of value.go)
2. The end of one cycle of updateDiscardStats (value.go: 1117)

We use this channel here to synchronize between go-routines instead of using time.Sleep (and simply praying for synchronization).

Note that the onCloseDiscardCapture map is *not* something we wait for! Instead, it is simply populated automatically at the time that we invoke valueLog.Close(db). (Called from DB.Close()). Since DB.Close() is synchronous, we don't have to wait for this data via any sort of special channel.

Can we live with just one of these? Theoretically, we could populate a string with the map contents and push it to the testChan. However, I do not recommend this approach. The two additions: the testChan and onCloseDiscardCapture have distinct requirements.

db.go
billprovince: Well, what it literally does is log a message to the test channel.  :)
If we ever wish to extend the usage of testChan (or "syncChan" if you prefer) to cover other use-cases, we should avoid having it be related specifically to updateDiscardStats. Having said that, it is worth noting that this new func is referenced in 2 places in the code: first to update for discardStats and second to notify when we reach the end of initialization of the valueLog (db.vlog). It is *not* actually used to grab the updateDiscardStats! The discardStats are retrieved when we close the database (or more precisely, the valueLog).

options.go
```diff
@@ -360,6 +366,26 @@ func (opt Options) FromSuperFlag(superflag string) Options {
...
+	opt.testChan = testChan
+	return opt
+}
+
+// withOnCloseDiscardCapture makes a shallow copy of the map c to
+// opt.onCloseDiscardCapture so that when we Close the DB, we make sure to
+// copy the contents of the discardStats to the map c.
+// This is strictly for testing purposes.
+func (opt Options) withOnCloseDiscardCapture(c map[uint64]uint64) Options {
```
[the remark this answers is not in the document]
billprovince: Same response

options.go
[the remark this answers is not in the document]
mangalaman93: Can we not access private fields in the test directly without exposing these functions?

value.go
```diff
@@ -554,6 +554,8 @@ func (vlog *valueLog) init(db *DB) {
 	lf, err := InitDiscardStats(vlog.opt)
 	y.Check(err)
 	vlog.discardStats = lf
+	// See TestPersistLFDiscardStats for purpose of statement below.
+	db.logToSyncChan("End: vlog.init(db)")
```
mangalaman93, line 558: would it better to use enum here instead of strings?

test_only.go
```diff
@@ -0,0 +1,97 @@
+/*
```
mangalaman93, line 1: we could rename this file to test_extensions.go

db.go
```diff
@@ -252,6 +254,9 @@ func Open(opt Options) (*DB, error) {
 		bannedNamespaces: &lockedKeys{keys: make(map[uint64]struct{})},
 		threshold:        initVlogThreshold(&opt),
 	}
+
+	db.setSyncChan(opt.syncChan)
```
mangalaman93, line 258: I am thinking if it seems fine to directly do db.syncChan = chan in the test code and skip this code here

test_only.go
```diff
@@ -0,0 +1,97 @@
...
+
+// TODO: Consider using this with specific compilation tags so that it only
+//       shows up when performing testing (e.g., specify build tag=unit).
+//       We are not yet ready to do that, as it may impact customer usage as
+//       well as requiring us to update the CI build flags. Moreover, the
+//       current model does not actually incur any significant cost.
+//       If we do this, we will also want to introduce a parallel file that
+//       overrides some of these structs and functions with empty contents.
+
+// testOnlyOptions specifies an extension to the type Options that we want to
```
mangalaman93, line 30: can we skip this too?

test_util.go
```diff
@@ -0,0 +1,58 @@
+/*
```
mangalaman93, line 1: we can just have all the code in one file instead of two

mangalaman93 (changes requested):
Thanks for the work Bill. This looks a lot cleaner to me and a good balance of what we can achieve for testing. My major concerns are two:
1. I am thinking we don't need to do this dance of having options, with functions, setters and then the actual field to hold test extensions. We can skip all of that and let the test directly set the field that it needs.
2. Instead of using strings for comparison, would it be better to use an enum instead?

value.go
billprovince: I like the idea still of logging strings, because it is actually kind of helpful to be able to have a string-channel for dumping log messages using the waitForMessage function.
I modified this to use string constants instead of just random strings however, which, I believe, will get much the same effect that you are looking for: we want to match on a very specific message.

test_only.go
billprovince: Done.

db.go
billprovince: Sadly, this does not work. We actually want to capture some of the messages that occur during the opening of the DB (i.e., inside of Open(opts)). In particular, the vlog.init message occurs during the Open function, and if we wait until after Open returns the DB, we will have missed the opportunity to capture the message.

test_util.go
billprovince: I highly recommend against that! Notice that util_test.go (was: test_util.go) imports "testing". We do NOT want to import testing in the production DB!  In fact, this is why I renamed from test_util.go to util_test.go: so that the file would be skipped during normal compilation and only included when doing testing.

test_only.go
billprovince: I removed the discardCapture from the Options, but retained the syncChan for the reasons I described earlier; we still need syncChan to be sent in during the Open function.

billprovince:
> Thanks for the work Bill. This looks a lot cleaner to me and a good balance of what we can achieve for testing. My major concerns are two:
> 
> 1. I am thinking we don't need to do this dance of having options, with functions, setters and then the actual field to hold test extensions. We can skip all of that and let the test directly set the field that it needs.
> 2. Instead of using strings for comparison, would it be better to use an enum instead?

Thanks, Aman.
Due to the need to capture messages that occur during `DB* Open(opts)`, we need to specify the syncChan on the options. Waiting unitl after Open returns will have us lose out on the opportunity. However, we could skip the onCloseDiscardCapture option, as this can be updated after Open safely.

With respect to strings vs. enums, I still prefer strings, but I made them string constants instead of just hard-coded strings. This way, the compiler can help us determine if there was an error or not. (Which is one of the main advantages of using enums).

Enums would definitely confer an advantage: however, the advantage of strings is that it is easier to log them and generate messages with them! I actually used the channel to get debug information during testing, which you may be aware is a bit difficult to obtain using standard `go test`. Logged messages don't get logged unless they are hooked up to a testing.T instance. Of course, I removed my ugly debugging messages, but having the ability to add them back when desired is clearly helpful.

test_extensions.go
```diff
@@ -0,0 +1,92 @@
...
+
+// TODO: Consider using this with specific compilation tags so that it only
+//       shows up when performing testing (e.g., specify build tag=unit).
+//       We are not yet ready to do that, as it may impact customer usage as
+//       well as requiring us to update the CI build flags. Moreover, the
+//       current model does not actually incur any significant cost.
+//       If we do this, we will also want to introduce a parallel file that
+//       overrides some of these structs and functions with empty contents.
+
+const updateDiscardStatsMsg = "updateDiscardStats iteration done"
```
mangalaman93, line 30: see if you want to improve this code

test_extensions.go
```diff
@@ -0,0 +1,92 @@
...
+	// syncChan is used to listen for specific messages related to activities
+	// that can occur in a DB instance. Currently, this is only used in
+	// testing activities.
+	syncChan chan string
+}
+
+// withSyncChan returns a new Options value with syncChan set to the given value.
+// If not specified, any operations that would otherwise occur with the syncChan
+// will be silently skipped.
+func (opt Options) withSyncChan(ch chan string) Options {
```
mangalaman93, line 45: remove this func

test_extensions.go
```diff
@@ -0,0 +1,92 @@
...
+	// process of performing the Close operation. Currently, we only consider
+	// using this during testing.
+	onCloseDiscardCapture map[uint64]uint64
+}
+
+// setSyncChan is a trivial setter for db.testOnlyDbExtensions.syncChan.
+// Strictly speaking, this has little value for us, except that it
+// can isolate the test-specific behaviors of a production Badger system
+// to this single file.
+func (db *DB) setSyncChan(ch chan string) {
```
mangalaman93, line 65: this too

util_test.go
```diff
@@ -0,0 +1,58 @@
+/*
```
mangalaman93, line 1: move this code to db_test.go

mangalaman93 (changes requested):

test_extensions.go
billprovince: Done.

test_extensions.go
billprovince: Done.

test_extensions.go
billprovince: Done.

util_test.go
billprovince: Done.

mangalaman93 (approved):
Thanks for addressing all the comments, looks much better now.

PR #1967, opened 2023-06-07
fix(Test): BenchmarkDbGrowth

SpadeA-Tang:
## Problem
 

It seems some parameter settings have been changed in the past. As a result, some tests (at least one) lose the ability to test what they want to test.
For `BenchmarkDbGrowth`, current default value of `ValueThreshold` makes no vlog will be created and current default value of `MemTableSize` makes no memtable will be flushed which means no compaction and thus no vlog gc will be performed.

## Solution
 
Setting  `ValueThreshold` be 1024 and `MemTableSize` be 1MB in this case seems to be fine.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -466,6 +466,8 @@ func BenchmarkDbGrowth(b *testing.B) {
 	opts.NumVersionsToKeep = 1
 	opts.NumLevelZeroTables = 1
 	opts.NumLevelZeroTablesStall = 2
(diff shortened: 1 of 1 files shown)


mangalaman93 (approved):

SpadeA-Tang:
Would you mind helping me to merge this? @mangalaman93

mangalaman93:
Thanks for the PR @SpadeA-Tang

PR #1971, opened 2023-06-16
chore(all): fix staticcheck errors

qdm12:
## Problem

In my VSCode configuration, I have gopls setup with

```json
"gopls": {
    "staticcheck": true,
```

and therefore staticcheck is enabled even without golangci-lint, and also ignores `//nolint:staticcheck` commands. It might also be a different staticcheck since it complained about issues not `//nolint;staticcheck`-ed in the code.

Anyway, this spotted a few staticcheck related errors which I fixed so they won't show if someone has the same config as I do.

## Solution

- do not end error strings with punctuation or new line
- do not start error strings with uppercase letters
- do not use yoda conditions
- do not import package twice in the same file
- Remove empty branches and mention `Else` at the start of a comment

### DIFF
--- a/badger/cmd/bank.go
+++ b/badger/cmd/bank.go
@@ -148,7 +148,7 @@ func min(a, b uint64) uint64 {
 	return b
 }
 
-var errAbandoned = errors.New("Transaction abandonded due to insufficient balance")
+var errAbandoned = errors.New("transaction abandonded due to insufficient balance")
 
 func moveMoney(db *badger.DB, from, to int) error {
 	return db.Update(func(txn *badger.Txn) error {
@@ -597,5 +597,5 @@ func runTest(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 	log.Println("Test FAILED")
-	return fmt.Errorf("Test FAILED")
+	return fmt.Errorf("test FAILED")
 }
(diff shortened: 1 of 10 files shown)


mangalaman93 (changes requested):
Hi, thanks for the PR and fixing some of the static check issues. I noticed that you have modified some of the error messages that we return as part of our API. I would avoid any of the API changes to make a tool happy for now. Could you rework your PR to not change any returned error messages? Any changes to test files are acceptable, I will review those.

qdm12:
> I noticed that you have modified some of the error messages that we return as part of our API. I would avoid any of the API changes to make a tool happy for now.

Are you sure your error strings should be considered as part of your API? Go 1.13 released error wrappings and `errors.Is` and since then encourages to use errors.Is and not error strings for consumers to test for errors programmatically. Although this error wrapping doesn't look it is used in the repository, so maybe you have consumers checking error strings programmatically? 🤔

mangalaman93:
> Are you sure your error strings should be considered as part of your API? Go 1.13 released error wrappings and `errors.Is` and since then encourages to use errors.Is and not error strings for consumers to test for errors programmatically. Although this error wrapping doesn't look it is used in the repository, so maybe you have consumers checking error strings programmatically? thinking

Badger has been around longer than Go 1.13 and I can't claim with 100% confidence that all error strings are used for comparing the errors but I know we do that in some tests sometimes. I would rather make a big change at some point later to standardize these errors and call it out in the release notes while doing a major release instead of breaking our users' applications just in case. I hope that makes sense. You could break this PR into two PRs, one with the errors change and one with the rest. We can include/use the errors change when we standardize these errors. You could also help us do that.

mangalaman93 (approved):

mangalaman93:
@qdm12 Thanks for the PR. Could you resolve the conflicts please?

PR #1973, opened 2023-06-18
give correct publisher metadata

mYmNeo:
## Problem
When using publisher to replay KV, the publisher doesn't give correct `userMeta` and missing `meta` fields.

## Solution

Set correct `userMeta` field and provide `meta` field

### DIFF
--- a/publisher.go
+++ b/publisher.go
@@ -97,7 +97,8 @@ func (p *publisher) publishUpdates(reqs requests) {
 			kv := &pb.KV{
 				Key:       y.ParseKey(k),
 				Value:     y.SafeCopy(nil, e.Value),
-				Meta:      []byte{e.UserMeta},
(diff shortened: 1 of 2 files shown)


harshil-goel:
The idea is that meta field is internal to Badger, and should not be published. UserMeta is the one provided by user, and that is the one that we publish. Can you describe your use case and why you require the internal badger meta?

mYmNeo:
> The idea is that meta field is internal to Badger, and should not be published. UserMeta is the one provided by user, and that is the one that we publish. Can you describe your use case and why you require the internal badger meta?

There are two different problems. One is that the `Meta` in proto file is not `userMeta`, so this commit fix it. Second is that I want to cache kv in memory in my app. But publisher only reports added key even though it's a deleted key, that will cause a infinity increasing of memory

PR #1976, opened 2023-07-04
chore: remove refs to deprecated io/ioutil

testwill:
## Problem
 
"io/ioutil" has been deprecated since Go 1.19: As of Go 1.16
## Solution
 
remove refs to deprecated io/ioutil

### DIFF
--- a/badger/cmd/info.go
+++ b/badger/cmd/info.go
@@ -20,7 +20,7 @@ import (
 	"bytes"
 	"encoding/hex"
 	"fmt"
-	"io/ioutil"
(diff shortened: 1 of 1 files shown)


mangalaman93:
Hi @testwill, Thanks for the PR. Will it make sense to create a function to do this instead of doing this again and again?

mangalaman93 (changes requested):
One more suggestion, this way it looks cleaner.

mangalaman93 (approved):

PR #1979, opened 2023-07-11
chore: pkg imported more than once

testwill:
## Problem
package "github.com/dgraph-io/badger/v4/pb" is being imported more than once

## Solution
Packages are imported only once


mangalaman93 (approved):
Thanks for the PR

PR #1980, opened 2023-07-17
fix(Logging): Fix invalid direct access to logger

PapaCharlie:
## Problem
Because the logger provided in the Options can be nil, directly interacting with it can cause panics. 

## Solution
This change fixes this issue by using the various logging methods provided by Options that wrap the Logger field.


mangalaman93 (approved):

mangalaman93:
Thank you for the PR.

PapaCharlie:
My pleasure!

Issue #1983, opened 2023-08-01
[BUG]: os.Exit on datastore close from block.ref going bellow zero


fredcarle:
### What version of Badger are you using?

v4.1.0

### What version of Go are you using?

1.20

### Have you tried reproducing the issue with the latest release?

Yes

### What is the hardware spec (RAM, CPU, OS)?

16GB, M1, macOS

### What steps will reproduce the bug?

We are using Badger as the datastore for DefraDB and while doing integration test and running our change detector, we run tests from both our develop branch and PR branch. The develop tests each use a unique set of Badger storage files but those files are then reused for the same test in the PR branch. This means that we close the Badger instance in after the tests in develop and run it again in the PR branch. On that second close, we often get an asset Failure on close because the `block.decrRef` method tries to decrease the reference even though it's already zero. https://github.com/sourcenetwork/defradb/actions/runs/5628956063/job/15253408371

### Expected behavior and actual result.

We expect the badger datastore to close without error.

PR #1984, opened 2023-08-01
fix(table): Prevent unnecessary os.Exit

fredcarle:
## Problem
We are using Badger as the datastore for DefraDB and while doing integration test and running our change detector, we run tests from both our develop branch and PR branch. The develop tests each use a unique set of Badger storage files but those files are then reused for the same test in the PR branch. This means that we close the Badger instance in after the tests in develop and run it again in the PR branch. On that second close, we often get an asset Failure on close because the block.decRef method tries to decrease the reference even though it's already zero. https://github.com/sourcenetwork/defradb/actions/runs/5628956063/job/15253408371

Fixes: #1983 

## Solution
What seems to solve the issue is to stop `block.decrRef` from bringing `block.ref` bellow zero. This may not be preferred solution of the Badger team as it might have broader implications but I thought we could use this as a starting point.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -222,7 +222,7 @@ func (b *block) incrRef() bool {
 	}
 }
 func (b *block) decrRef() {
-	if b == nil {
+	if b == nil || b.ref.Load() == 0 {
 		return
 	}
 
@@ -239,7 +239,6 @@ func (b *block) decrRef() {
 		NumBlocks.Add(-1)
 		// blockPool.Put(&b.data)
 	}
-	y.AssertTrue(b.ref.Load() >= 0)
 }
 func (b *block) size() int64 {
 	return int64(3*intSize /* Size of the offset, entriesIndexStart and chkLen */ +


mangalaman93:
@fredcarle Thanks for filling a detailed bug and the possible fix for it. As you said, this seems like a bug somewhere in either DefraDB's code or badger's (or maybe ristretto's). This is an invariant that we should maintain and the assert ensures that we do not violate this invariant. Please check where exactly the references are increased and decreased. If it is possible to provide steps to reproduce, I can spend some time as well looking into it.

fredcarle:
> @fredcarle Thanks for filling a detailed bug and the possible fix for it. As you said, this seems like a bug somewhere in either DefraDB's code or badger's (or maybe ristretto's). This is an invariant that we should maintain and the assert ensures that we do not violate this invariant. Please check where exactly the references are increased and decreased. If it is possible to provide steps to reproduce, I can spend some time as well looking into it.

I've been trying to see exactly where it's happening  but there isn't anything conclusive for me at this point. I don't know Badger enough to be certain. 

I've printed out where the `decrRef` method is being called and, when we see the failure, we usually have this pattern:
```
close block iterator
decrRef 0x140004e4150, 3
close block iterator
decrRef 0x140004e4150, 2
close block iterator
decrRef 0x140004e4150, 1
setblock
decrRef 0x140004e4150, 0
```
The `decrRef` call on setblock with a `block.ref` of zero is the odd one. The close call is done right after this so thats why we get the os.Exit issue.

To reproduce, you could clone the DefraDB repo and run the following:
```
env DEFRA_DETECT_DATABASE_CHANGES=true gotestsum -- ./tests/integration/query/one_to_many_to_one -shuffle=on -p 1
```

fredcarle:
We've merged a change that uses this temporary relief so the test above won't help anymore unfortunately.

fredcarle:
Closing this PR as I can't find any other way to reproduce this os.Exit situation.

PR #1990, opened 2023-08-06
Added get multiple api for badger

ghost:
Added get multiple api in badger.
Able to get upto 60% performance improvement in ldbc s10 dataset queries.
4.69 seconds -> 2.61 seconds

### DIFF
--- a/db.go
+++ b/db.go
@@ -746,6 +746,8 @@ func (db *DB) getMemTables() ([]*memTable, func()) {
 // get returns the value in memtable or disk for given key.
 // Note that value will include meta byte.
 //
+// getBatch would return the values of list of keys in order
+//
 // IMPORTANT: We should never write an entry with an older timestamp for the same key, We need to
 // maintain this invariant to search for the latest value of a key, or else we need to search in all
 // tables and find the max version among them.  To maintain this invariant, we also need to ensure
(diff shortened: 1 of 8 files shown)


db.go
```diff
@@ -756,6 +756,42 @@ func (db *DB) getMemTables() ([]*memTable, func()) {
...
+			continue
+		}
+		version := y.ParseTs(key)
+		for i := 0; i < len(tables); i++ {
+			vs := tables[i].sl.Get(key)
+			y.NumMemtableGetsAdd(db.opt.MetricsEnabled, 1)
+			if vs.Meta == 0 && vs.Value == nil {
+				continue
+			}
+			// Found the required version of the key, return immediately.
```
kevinmingtarja, line 780: This comment should be refactored, since we don't return immediately unlike in `db.get()`

discard_test.go
```diff
@@ -17,12 +17,171 @@
...
+}
+
+// bytesLessThan compares two byte arrays lexicographically.
+func bytesLessThan(a, b []byte) bool {
+	return bytes.Compare(a, b) >= 0
+}
+
+func TestReadC(t *testing.T) {
+	t.Skip()
+	allKeysF, err := os.Open("/home/harshil/all_keys_2")
```
tauraamui, line 77: Bit weird to have an explicit user path, make it a temp path or something?

harshil-goel: This s just a temp test for my local validation. Hence the t.Skip() before the test starts. I can't commit this as the directory being used is quite big.

level_handler.go
```diff
@@ -273,6 +273,113 @@ func (s *levelHandler) getTableForKey(key []byte) ([]*table.Table, func() error)
...
+			if done[i] {
+				continue
+			}
+			results[i], err = s.get(key)
+			if err != nil {
+				return results, err
+			}
+		}
+		return results, nil
+	} else {
```
sanjayk-github-dev, line 344: @harshil-goel I need to understand this code block ... will sync up with you offline.

level_handler.go
```diff
@@ -273,6 +273,113 @@ func (s *levelHandler) getTableForKey(key []byte) ([]*table.Table, func() error)
 	return []*table.Table{tbl}, tbl.DecrRef
 }
 
+func (s *levelHandler) getBatch(keys [][]byte, done []bool) ([]y.ValueStruct, error) {
+	// Find the table for which the key is in, and then seek it
+	getForKey := func(key []byte) (y.ValueStruct, func() error, []*table.Iterator) {
+		tables, decr := s.getTableForKey(key)
+		keyNoTs := y.ParseKey(key)
```
mangalaman93, line 280: we should just parse the key once, we are doing it again and again

level_handler.go
```diff
@@ -273,6 +273,113 @@ func (s *levelHandler) getTableForKey(key []byte) ([]*table.Table, func() error)
...
+	}
+
+	// Use old results from getForKey and find in those tables.
+	findInIter := func(key []byte, itrs []*table.Iterator) y.ValueStruct {
+		var maxVs y.ValueStruct
+
+		for _, it := range itrs {
+			it.Seek(key)
+			if !it.Valid() {
+				continue
```
mangalaman93, line 317: break?

mangalaman93 (changes requested):
This PR needs lot more refactoring, testing and cleanup. Only two comments for now but once it is cleaned up, can review again.

PR #2011, opened 2023-09-26
chore: update readme

joshua-goldstein:
README still links to master in places.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Please consult the [Changelog] for more detailed information on releases.
 
 For more details on our version naming schema please read [Choosing a version](#choosing-a-version).
 
-[Changelog]:https://github.com/dgraph-io/badger/blob/master/CHANGELOG.md
+[Changelog]:https://github.com/dgraph-io/badger/blob/main/CHANGELOG.md
 
 ## Table of Contents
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,7 +39,7 @@ Please consult the [Changelog] for more detailed information on releases.
 
 For more details on our version naming schema please read [Choosing a version](#choosing-a-version).
 
-[Changelog]:https://github.com/dgraph-io/badger/blob/master/CHANGELOG.md
+[Changelog]:https://github.com/dgraph-io/badger/blob/main/CHANGELOG.md
```
Jorropo, line 42: This indeed is what got me.

Jorropo (approved):
Thx
I think the README needs improvement, mainly adding a line about v4 in `Project Status`. But that would have been good enough for me (given I checked out the changelog from README and would have found the v4 entries).

joshua-goldstein:
> Thx
> I think the README needs improvement, mainly adding a line about v4 in `Project Status`. But that would have been good enough for me (given I checked out the changelog from README and would have found the v4 entries).

Agreed, I will look into making the README more clear.  The switch to v4 at least deserves a mention there.

PR #2028, opened 2023-11-08
fix(level-handler): level handler - key with zero version can never be found

agfn:
## Problem
 

After importing database from protobuf (with `StreamWriter`), I can't find keys with zero version.


mangalaman93:
@agfn could you rebase please?

PR #2031, opened 2023-11-25
fix(Transaction): discard empty transactions on CommitWith 

Wondertan:
## Problems
* Transactions with empty `pendingWrites` were never discarded(and marked as done) for `CommitWith`
	* This is similar to https://github.com/dgraph-io/badger/pull/2018, which solved the same problem for `Commit`, but not for `CommitWith`
	* The `CommitWith` is used by `WriteBatch`, so flushing an empty batch never discarded the inner shadowed transaction, causing a multitude of issues.

## Solution
Make sure `Discard` is called for `CommitWith` when `pendingWrites` are empty. The existing unit test is updated to assert that.

### DIFF
--- a/txn.go
+++ b/txn.go
@@ -718,6 +718,8 @@ func (txn *Txn) CommitWith(cb func(error)) {
 		// callback might be acquiring the same locks. Instead run the callback
 		// from another goroutine.
 		go runTxnCallback(&txnCb{user: cb, err: nil})
+		// Discard the transaction so that the read is marked done.
+		txn.Discard()
 		return
 	}
 
--- a/batch_test.go
+++ b/batch_test.go
@@ -103,8 +103,11 @@ func TestEmptyWriteBatch(t *testing.T) {
 			wb = db.NewWriteBatch()
 			require.NoError(t, wb.Flush())
 			wb = db.NewWriteBatch()
+			// Flush commits inner txn and sets a new one instead.
+			// Thus we need to save it to check if it was discarded.
 			txn := wb.txn
 			require.NoError(t, wb.Flush())
+			// check that flushed txn was discarded and marked as read.
 			require.True(t, txn.discarded)
 		})
 	})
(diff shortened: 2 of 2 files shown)


Wondertan:
cc @billprovince @joshua-goldstein

txn.go
```diff
@@ -712,6 +711,7 @@ func (txn *Txn) CommitWith(cb func(error)) {
 	if cb == nil {
 		panic("Nil callback provided to CommitWith")
 	}
+	defer txn.Discard()
```
harshil-goel, line 714: Why two defer for the sam ething?

Wondertan: I don't understand the question

Wondertan: Could you rephrase it?

Wondertan: Yeah, I get it, but what's the same thing? If that's about `Commit` and `CommitWith`, then it's not the same. They are two different ways of committing data and both have to Discard.

Also note that Discard protection from multiple calls, and even if a user calls both there won't be any side effects.

txn_test.go
```diff
@@ -761,7 +761,6 @@ func TestManagedDB(t *testing.T) {
 		for i := 0; i <= 3; i++ {
 			require.NoError(t, txn.SetEntry(NewEntry(key(i), val(i))))
 		}
-		require.Error(t, txn.Commit())
```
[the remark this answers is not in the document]
Wondertan: To avoid committing twice.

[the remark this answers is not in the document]
Wondertan: There was no good reason, so I took it back to the original state.

Wondertan:
Ok, so I think I misunderstood the purpose of `precommitCheck`, and transactions should not be discarded when it errors. I still think the code in Commit funcs is rather confusing and could be simplified to have a single defer, but that's definitely not part of this PR, so I shortened the solution to a one-liner, as well updated unit tests that checks for the bug

Wondertan:
Ping again

mangalaman93 (approved):

PR #2042, opened 2024-01-16
fix: sentinel errors should not have stack traces

mitar:
## Problem

While debugging a program which is using badger I noticed that stack traces returned from my code (which uses errors with stack traces) are useless. I found the reason: badger uses sentinel errors (great!) but it creates them with stack traces at the "global init time" (e.g., `var ErrFoo = errors.New(...)`) and not at "return from function time". Because of that I was seeing stack traces like:

```
github.com/dgraph-io/badger/v4.init
	/go/pkg/mod/github.com/dgraph-io/badger/v4@v4.2.0/compaction.go:41
runtime.doInit1
	/usr/local/go/src/runtime/proc.go:6740
runtime.doInit
	/usr/local/go/src/runtime/proc.go:6707
runtime.main
	/usr/local/go/src/runtime/proc.go:249
runtime.goexit
	/usr/local/go/src/runtime/asm_amd64.s:1650
```

Instead of a stack trace which would show me where the call which is returning sentinel error is made.

## Solution

Ideally, one would create all sentinel errors without stack traces (standard errors package works great for that) and then when the error happens in a function, you return `errors.WithStack(ErrFoobar)` instead of just `ErrFoobar`. This means that sentinel error is then annotated with a stack trace at the place where the error happens. But if you do that, then you have to check for sentinel errors with `errors.Is` instead of just comparing errors with equality (`==`). That might be a breaking change for people who are not using `errors.Is` as they should. Because badger's own codebase uses equality instead of `errors.Is` I decided to not do this in this PR, but instead:

Just make sentinel errors without stack traces. This allows me to annotate with a stack trace inside my program (otherwise my program does not do so because if finds an existing stack trace and assumes it is a deeper one) and I get at least that part of the stack trace. Equality still works as usual.

I suggest this is merged this way and in v5 of badger `errors.WithStack` and required `errors.Is` are introduced.

## Side note

Some time ago [I made this errors package for Go](https://gitlab.com/tozd/go/errors) which fixes various issues in `github.com/pkg/errors` and also has direct support for sentinel (in that package called "base") errors. It also has some features badger has in its `y` package for dealing with errors. You could consider adopting that package to have a comprehensive errors handling in badger.

### DIFF
--- a/badger/cmd/bank.go
+++ b/badger/cmd/bank.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	stderrors "errors"
 	"fmt"
 	"log"
 	"math"
@@ -148,7 +149,7 @@ func min(a, b uint64) uint64 {
 	return b
 }
 
-var errAbandoned = errors.New("Transaction abandonded due to insufficient balance")
+var errAbandoned = stderrors.New("Transaction abandonded due to insufficient balance")
 
 func moveMoney(db *badger.DB, from, to int) error {
 	return db.Update(func(txn *badger.Txn) error {
--- a/db.go
+++ b/db.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	stderrors "errors"
 	"expvar"
 	"fmt"
 	"math"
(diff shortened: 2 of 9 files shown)


mitar:
Still relevant.

mitar:
Still relevant.

mangalaman93 (approved):

PR #2048, opened 2024-02-21
Fix build for GOARCH=wasm with GOOS=js or GOOS=wasip1

paralin:
Fixes the following build failures:

GOOS=js GOARCH=wasm go build ./...
GOOS=wasip1 GOARCH=wasm go build ./...

Depends on: https://github.com/dgraph-io/ristretto/pull/375


paralin:
Still relevant

mangalaman93:
@paralin could you rebase the PR on latest `main` branch? We have merged the ristretto PR and badger has the latest changes now.

paralin:
@mangalaman93 I rebased it. The tests do not pass since mmap fails on wasm. I think there would need to be conditional logic to not use mmap on wasm / js for this to actually be usable. But at least it compiles on these platforms now, with stubs.

paralin:
Rebased again. I think this is still good to merge, even though js does not work properly, at least it builds cleanly.

mangalaman93 (approved):

dir_other.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, y.Wrapf(err, "cannot open directory %q", dirPath)
+	}
+
+	// NOTE: Here is where we would normally call flock.
+	// This is not supported in js / wasm, so skip it.
+
+	if !readOnly {
+		// Yes, we happily overwrite a pre-existing pid file.  We're the
+		// only read-write badger process using this directory.
+		err = os.WriteFile(absPidFilePath, []byte(fmt.Sprintf("%d\n", os.Getpid())), 0666)
```
mangalaman93, line 63: It is possible there maybe other processes using the same badger?

mangalaman93, line 63: It is possible there maybe other processes using the same badger?

dir_other.go
```diff
@@ -0,0 +1,109 @@
+//go:build js || wasip1
+// +build js wasip1
+
+/*
+ * Copyright 2017 Dgraph Labs, Inc. and Contributors
```
mangalaman93, line 5: 2024

mangalaman93 (changes requested):
Thank you for the PR, just two minor comments.

dir_other.go
paralin: Wasm only has one process at a time, and we have no way to check process IDs anyway in that environment. But I'm not sure if the comment is accurate fully. Maybe we are accessing a shared filesystem.

mangalaman93: yeah, we made this choice before for other archs as well

mangalaman93 (approved):

PR #2063, opened 2024-05-28
Feature: Allow stream custom maxsize per batch

simon28082:
## Problem
 
I'm trying to read the data using badger and use grpc for node sync, but I can't change the max cut size of stream because it's restricted hardcode

```go
// maxStreamSize is the maximum allowed size of a stream batch. This is a soft limit
// as a single list that is still over the limit will have to be sent as is since it
// cannot be split further. This limit prevents the framework from creating batches
// so big that sending them causes issues (e.g running into the max size gRPC limit).
var maxStreamSize = uint64(100 << 20) // 100MB
```
## Solution
 
Add a property to Stream to allow custom slice size

```go
type Stream struct {
	// Prefix to only iterate over certain range of keys. If set to nil (default), Stream would
	// iterate over the entire DB.
	Prefix []byte

	// Number of goroutines to use for iterating over key ranges. Defaults to 8.
	NumGo int

	// Badger would produce log entries in Infof to indicate the progress of Stream. LogPrefix can
	// be used to help differentiate them from other activities. Default is "Badger.Stream".
	LogPrefix string

	// ChooseKey is invoked each time a new key is encountered. Note that this is not called
	// on every version of the value, only the first encountered version (i.e. the highest version
	// of the value a key has). ChooseKey can be left nil to select all keys.
	//
	// Note: Calls to ChooseKey are concurrent.
	ChooseKey func(item *Item) bool

	// MaxSize is the maximum allowed size of a stream batch
	MaxSize uint64
}
```

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -323,7 +323,7 @@ func (st *Stream) streamKVs(ctx context.Context) error {
 			// Send the batch immediately if it already exceeds the maximum allowed size.
 			// If the size of the batch exceeds maxStreamSize, break from the loop to
 			// avoid creating a batch that is so big that certain limits are reached.
-			if batch.LenNoPadding() > uint64(st.MaxSize) {
+			if uint64(batch.LenNoPadding()) > st.MaxSize {
 				break loop
 			}
 			select {
--- a/stream.go
+++ b/stream.go
@@ -323,7 +323,7 @@ func (st *Stream) streamKVs(ctx context.Context) error {
 			// Send the batch immediately if it already exceeds the maximum allowed size.
 			// If the size of the batch exceeds maxStreamSize, break from the loop to
 			// avoid creating a batch that is so big that certain limits are reached.
-			if batch.LenNoPadding() > int(st.MaxSize) {
+			if batch.LenNoPadding() > uint64(st.MaxSize) {
 				break loop
 			}
 			select {
--- a/stream.go
+++ b/stream.go
@@ -62,6 +62,14 @@ type Stream struct {
 	// Note: Calls to ChooseKey are concurrent.
 	ChooseKey func(item *Item) bool
 
+	// MaxSize is the maximum allowed size of a stream batch. This is a soft limit
+	// as a single list that is still over the limit will have to be sent as is since it
+	// cannot be split further. This limit prevents the framework from creating batches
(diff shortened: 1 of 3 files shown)


mangalaman93:
@simon28082 the linter is not happy. Could you check that please? and also pull in latest changes from the `main` branch?

simon28082:
@mangalaman93 will do it

simon28082:
@mangalaman93 I done it, please review

stream.go
```diff
@@ -315,7 +323,7 @@ func (st *Stream) streamKVs(ctx context.Context) error {
 			// Send the batch immediately if it already exceeds the maximum allowed size.
 			// If the size of the batch exceeds maxStreamSize, break from the loop to
 			// avoid creating a batch that is so big that certain limits are reached.
-			if batch.LenNoPadding() > int(maxStreamSize) {
+			if batch.LenNoPadding() > uint64(st.MaxSize) {
```
mangalaman93, line 326: I think what you should do here is this:
```
if uint64(batch.LenNoPadding()) > st.MaxSize {
```

mangalaman93 (changes requested):

stream.go
simon28082: modified

mangalaman93 (dismissed):

stream.go
```diff
@@ -315,7 +323,7 @@ func (st *Stream) streamKVs(ctx context.Context) error {
 			// Send the batch immediately if it already exceeds the maximum allowed size.
 			// If the size of the batch exceeds maxStreamSize, break from the loop to
 			// avoid creating a batch that is so big that certain limits are reached.
-			if batch.LenNoPadding() > int(maxStreamSize) {
```
mangalaman93, line 318: could you try upgrading the action version in .github/workflows/ci-golang-lint.yml:29 to
```
uses: golangci/golangci-lint-action@v6.1.1
```

stream_test.go
```diff
@@ -171,6 +171,68 @@ func TestStream(t *testing.T) {
 	require.NoError(t, db.Close())
 }
 
+
+func TestStreamMaxSize(t *testing.T) {
```
mangalaman93, line 175: This file is not go formatted, please check

Issue #2035, opened 2023-12-12
[BUG]: not usable with `GOOS=aix` (Unix)


mohammed90:
### What version of Badger are you using?

- github.com/dgraph-io/badger v1.6.2
- github.com/dgraph-io/badger/v2 v2.2007.4

### What version of Go are you using?

Go1.21.5

### Have you tried reproducing the issue with the latest release?

Yes

### What is the hardware spec (RAM, CPU, OS)?

RAM & CPU: Not sure, whatever Github Actions uses (reproduced [here](https://github.com/caddyserver/caddy/actions/runs/7167726695/job/19514386912#step:5:13)).

OS: AIX

### What steps will reproduce the bug?

Build with: `GOOS=aix GOARCH=ppc64 go build`

### Expected behavior and actual result.

**Expectation:**

Successful compilation

**Actual result:**

```
# github.com/dgraph-io/badger
dir_unix.go:63:13: undefined: unix.Flock
```

```
# github.com/dgraph-io/badger/v2
dir_unix.go:62:13: undefined: unix.Flock
```

### Additional information

The issue is that Badger uses `unix.Flock` from the `golang.org/x/sys/unix`, but the referenced function is not available in the referenced dependency for AIX. The syscall exists in AIX, but not exposed/available in the `x/sys/unix` package. This has been reported upstream (golang/go#64669). 

Badger v1 and v2 is an [indirect dependency for Caddy](https://github.com/caddyserver/caddy/blob/7d919af01b31aff6eb1086e87784bf59c52419bb/go.mod#L84-L85), which is how we come across this issue. This was reported by one of the users who was attempting to compile Caddy for AIX (see caddyserver/caddy#5970). This isn't directly an error of badger. We've already reported it upstream (see golang/go#64669). I don't know if you're interested in figuring out a workaround (haven't personally tried any) or prefer to just wait for the Go ticket to be resolved by upstream.

mohammed90:
Keep. This issue is easily solvable and affects downstream.

mohammed90:
Keep

matthewmcneely:
Fixed in #2115

PR #2115, opened 2024-09-26
fix(aix): add aix directory synchronization support

pmur:
AIX doesn't support a proper flock like linux, but it seems to have enough support for process level file locking using fcntl.

For #2035

## Problem
GOOS=aix does not build. AIX does not support a linux-like flock, and does not export `unix.Flock`.

## Solution
Create an AIX specific directory-locking implementation using the AIX version of flock.

### DIFF
--- a/dir_aix.go
+++ b/dir_aix.go
@@ -138,7 +138,7 @@ func openDir(path string) (*os.File, error) { return os.Open(path) }
 // or see https://github.com/coreos/etcd/issues/6368 for an example.)
 func syncDir(dir string) error {
 	var err error
-	// TODO: AIX does not support fsync on a directory. Doing a full file system sync may be
-	// undesirable too (e.x os.Sync()).
+	// AIX does not support fsync on a directory.
+	// Data durability on crash may be affected.
 	return err
 }
--- a/dir_aix.go
+++ b/dir_aix.go
@@ -1,0 +1,144 @@
+//go:build aix
+// +build aix
+
+/*
+ * SPDX-FileCopyrightText: © 2017-2026 Istari Digital, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package badger
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+
+	"golang.org/x/sys/unix"
+
+	"github.com/dgraph-io/badger/v4/y"
+)
+
+// AIX flock locks files, not descriptors. So, multiple descriptors cannot
+// be used in the same file. The first to close removals the lock on the
+// file.
+type directoryLockGuard struct {
+	// The absolute path to our pid file.
+	path string
+	// Was this a shared lock for a read-only database?
+	readOnly bool
+}
+
+// AIX flocking is file x process, not fd x file x process like linux. We can
+// only hold one descriptor with a lock open at any given time.
+type aixFlock struct {
+	file     *os.File
+	count    int
+	readOnly bool
+}
+
+// Keep a map of locks synchronized by a mutex.
+var aixFlockMap = map[string]*aixFlock{}
+var aixFlockMapLock sync.Mutex
+
+// acquireDirectoryLock gets a lock on the directory (using flock). If
+// this is not read-only, it will also write our pid to
+// dirPath/pidFileName for convenience.
+func acquireDirectoryLock(dirPath string, pidFileName string, readOnly bool) (
+	*directoryLockGuard, error) {
+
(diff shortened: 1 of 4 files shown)


pmur:
Hi, is there something I need to fix in the CL, or is this a CI issue?

pmur:
This is still useful. Ping.

ryanfoxtyler:
@pmur tests are passing now, though I'm concerned about the maintenance cost of this when we don't have an environment to test on–is that something you're able/willing to help with?

dmitshur:
In case it's helpful, I'll point out that you can get some signal from compile-only testing via cross-compilation. For example, `GOOS=aix GOARCH=ppc64 go test -c -o=/dev/null` can be run on any platform you have access to, and at least check for build errors. (The -c flag tells the go command to compile the test binary but do not run it, while the -o flag tells it to discard the result, since you only care about the exit code.)

pmur:
I can provide some help in a limited capacity. Note, I have been using publicly available resources to implement test this (the aix golang CI instance, and the aix gcc compiler farm instance).

I hope @ayappanec might be able to provide more official help. He is part of the IBM team maintaining OSS software on aix.

ayappanec:
@pmur Thanks for tagging me here. 
We are currently in the process of creating new AIX VMs in OSU lab which we give it to various Opensource communities to enable AIX support. I will try to get a VM for this.

pmur:
ping

pmur:
This has been collected dust for awhile. Is this still blocking the transition of the Go AIX CI instance?

dir_aix.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	aixFlockMapLock.Lock()
+	defer aixFlockMapLock.Unlock()
+
+	lg := &directoryLockGuard{absPidFilePath, readOnly}
+
+	if lock, fnd := aixFlockMap[absPidFilePath]; fnd {
+		if !readOnly || lock.readOnly != readOnly {
+			fmt.Printf("error\n")
```
matthewmcneely: Please remove debugging line.

dir_aix.go
```diff
@@ -0,0 +1,157 @@
...
+// only hold one descriptor with a lock open at any given time.
+type aixFlock struct {
+	file     *os.File
+	count    int
+	readOnly bool
+}
+
+// Keep a map of locks synchronized by a mutex.
+var aixFlockMap = map[string]*aixFlock{}
+var aixFlockMapLock sync.Mutex
```
matthewmcneely: Advisory/No change needed: contention could be optimized by conversion to a RWMutex and changing the acquire/release code below as appropriate

dir_aix.go
```diff
@@ -0,0 +1,157 @@
...
+// or see https://github.com/coreos/etcd/issues/6368 for an example.)
+func syncDir(dir string) error {
+	f, err := openDir(dir)
+	if err != nil {
+		return y.Wrapf(err, "While opening directory: %s.", dir)
+	}
+
+	err = f.Sync()
+	closeErr := f.Close()
+	if err != nil {
```
matthewmcneely: OK, but a question: does the AIX sync actually work?

matthewmcneely (changes requested):
@pmur Sorry for the delay. If we get this through, I'll add some aix-specific build commands to our CD to validation that _at least_ it compiles without error.

pmur:
Ping

ayappanec:
@pmur Sure. Let me check and get back.

pmur:
I ran the tests on gcc119 with something like `GOMAXPROCS=8 go test -v -timeout=1h`.  You could try unsetting `tags` in `test.sh` for aix and running it too.

As for jemalloc, @ayappanec you may know better. I tried building the tip of jemalloc on gcc119, it failed with a nonsensical: 
```
configure: error: Unsupported pointer size: 0
```

ayappanec:
```
# lparstat

System configuration: type=Shared mode=Uncapped smt=2 lcpu=30 mem=40960MB psize=17 ent=1.50 

%user  %sys  %wait  %idle physc %entc  lbusy  vcsw phint  %nsp  %utcyc
----- ----- ------ ------ ----- ----- ------ ----- ----- -----  ------
  0.1   0.3    0.0   99.7  0.01   0.5    3.3 3900039994     0   101   0.78 
```

It has 15 virtual cpus (shared mode) and smt is 2. So logical cpus is 2x15 = 30. It's a POWER8 machine
Memory is 40GB.

francislavoie:
https://github.com/dgraph-io/badger/issues/2035 is still open, was this meant to close it?

PR #2169, opened 2025-02-13
chore(deps): Migrate from OpenCensus to OpenTelemetry

jsoriano:
**Description**

Migrate from OpenCensus to OpenTelemetry, doing some adjustements where there is no 1:1 migration path:
* `zpages` handler explicitly added to `http.DefaultServeMux`, as `zpages.Handle()` is not available anymore.
* Annotations replaced by Attributes.

I am also adjusting some uses of spans in compactions, so their annotations are not overwritten on every call to `runCompactDef` in cases where it is called multiple times.

Not using latest version of OpenTelemetry because it requires Go 1.22, and badger still supports 1.21.

Fix https://github.com/hypermodeinc/badger/issues/2155.

**Checklist**

- [x] For all _code_ changes, an entry added to the `CHANGELOG.md` file describing and linking to
      this PR
- [ ] Tests added for new functionality, or regression tests for bug fixes added as applicable
- [ ] For public APIs, new features, etc., PR on [docs repo](https://github.com/hypermodeinc/docs) staged and linked here


jsoriano:
Not sure how to address the issue with protobuf, could it be that `protoc` has a different version in CI? :thinking:

mauri870:
About the Go version, I see https://github.com/hypermodeinc/badger/pull/2125 downgraded it from 1.23 to 1.21. Even if the repo goes with the last N-3 versions as opposed to the Go release policy of N-2 that would still be valid to bump it to 1.22 now. Let's wait for a maintainer word on this.

mangalaman93:
We have upgraded the Min Go version to 1.22 here https://github.com/hypermodeinc/badger/pull/2171. Please upgrade the opentelemetry to the latest and we would be happy to merge the PR. Thank you for creating the PR.

jsoriano:
OTel dependencies upgraded.

levels.go
```diff
@@ -312,27 +313,31 @@ func (s *levelsController) dropPrefixes(prefixes [][]byte) error {
...
+			span.SetAttributes(attribute.String("Drop Prefixes", fmt.Sprintf("%v", prefixes)))
 			cd.t.baseLevel = l.level
 			if err := s.runCompactDef(-1, l.level, cd); err != nil {
 				opt.Warningf("While running compact def: %+v. Error: %v", cd, err)
+				span.End()
 				return err
 			}
+			span.SetAttributes(
+				attribute.Int("Top tables count", len(cd.top)),
+				attribute.Int("Bottom tables clount", len(cd.bot)))
```
mangalaman93, line 337: Count*

levels.go
```diff
@@ -1543,13 +1542,17 @@ func (s *levelsController) doCompact(id int, p compactionPriority) error {
...
+	span.SetAttributes(attribute.String("Compaction", fmt.Sprintf("%+v", cd)))
 	if err := s.runCompactDef(id, l, cd); err != nil {
 		// This compaction couldn't be done successfully.
 		s.kv.opt.Warningf("[Compactor: %d] LOG Compact FAILED with error: %+v: %+v", id, err, cd)
 		return err
 	}
 
+	span.SetAttributes(
+		attribute.Int("Top tables count", len(cd.top)),
+		attribute.Int("Bottom tables clount", len(cd.bot)))
```
mangalaman93, line 1554: Same

mangalaman93 (commented):
Minor spelling mistake! LGTM otherwise. @harshil-goel for final review before merge.

levels.go
```diff
@@ -1512,12 +1512,11 @@ func (s *levelsController) doCompact(id int, p compactionPriority) error {
 		p.t = s.levelTargets()
 	}
 
-	_, span := otrace.StartSpan(context.Background(), "Badger.Compaction")
+	_, span := otel.Tracer("Badger").Start(context.TODO(), "Compaction")
```
mangalaman93, line 1515: For backward compatibility, let's use the same name as before:
```
otel.Tracer("").Start(context.Background(), "Badger.Compaction")
```

value.go
```diff
@@ -1052,8 +1053,8 @@ func discardEntry(e Entry, vs y.ValueStruct, db *DB) bool {
 }
 
 func (vlog *valueLog) doRunGC(lf *logFile) error {
-	_, span := otrace.StartSpan(context.Background(), "Badger.GC")
-	span.Annotatef(nil, "GC rewrite for: %v", lf.path)
+	_, span := otel.Tracer("Badger").Start(context.TODO(), "GC")
```
mangalaman93, line 1056: same here, otel.Tracer("") so that the default one gets picked up
and "GC" --> "Badger.GC"

mangalaman93 (changes requested):
two more comments, we are good to merge otherwise! And resolve the conflicts as well. Thank you.

levels.go
jsoriano: Changed to the proposed names, thanks!

mangalaman93 (approved):

PR #2173, opened 2025-02-18
Added get multiple api for badger

ghost:


db.go
```diff
@@ -747,7 +749,54 @@ func (db *DB) getMemTables() ([]*memTable, func()) {
...
+
+	maxVs := make([]y.ValueStruct, len(keys))
+
+	y.NumGetsAdd(db.opt.MetricsEnabled, 1)
+	// For memtable, we need to check every memtable each time
+	for j, key := range keys {
+		if keysRead[j] {
+			continue
+		}
+		version := y.ParseTs(key)
```
mangalaman93, line 767: this can be an argument to the function

mangalaman93 (changes requested):
1. move the common functions out and use them in getBatch and get
2. Avoid allocating vss and seeks slices
3. define a function for converting ValueStruct to Item type
4. Consider taking the []*Item as an argument instead of constructing it inside GetBatch

PR #2197, opened 2025-05-08
feat: add ChecksumAlgorithm option to decides which algorithm calculate checksums.

LiuQhahah:
**Description**

add ChecksumAlgorithm option to decides which algorithm calculate checksums.
**Checklist**

**Instructions**

- The PR title should follow the [Conventional Commits](https://www.conventionalcommits.org/)
  syntax, leading with `fix:`, `feat:`, `chore:`, `ci:`, etc.
- The description should briefly explain what the PR is about. In the case of a bugfix, describe or
  link to the bug.
- In the checklist section, check the boxes in that are applicable, using `[x]` syntax.
  - If not applicable, remove the entire line. Only leave the box unchecked if you intend to come
    back and check the box later.
- Delete the `Instructions` line and everything below it, to indicate you have read and are
  following these instructions. 🙂

Thank you for your contribution to Badger!


mangalaman93:
@LiuQhahah Thank you for the PR. Could you help us understand the rationale behind the PR? I will review the PR as soon as I can. Thanks

LiuQhahah:
Hi @mangalaman93 
I saw in the code that there was a TODO tag and looked at this feature that hadn't been implemented yet, so I decided to implement it.
Below is the related code : https://github.com/hypermodeinc/badger/blob/main/table/builder.go#L457

Thanks
Qiang

badger/cmd/info.go
```diff
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"encoding/hex"
 	"fmt"
+	"github.com/dgraph-io/badger/v4/pb"
```
mangalaman93, line 12: Can we move it down along with other badger imports?

badger/cmd/info.go
```diff
@@ -69,6 +71,7 @@ func init() {
 	infoCmd.Flags().StringVar(&opt.encryptionKey, "enc-key", "", "Use the provided encryption key")
 	infoCmd.Flags().StringVar(&opt.checksumVerificationMode, "cv-mode", "none",
 		"[none, table, block, tableAndBlock] Specifies when the db should verify checksum for SST.")
+	infoCmd.Flags().StringVar(&opt.checksumAlgorithm, "ct", "crc32c", "[crc32c,xxhash64] Specifies the checksum algorithm for SST.")
```
mangalaman93, line 74: we need to break it down into two lines

badger/cmd/info.go
```diff
@@ -515,6 +520,18 @@ func pluralFiles(count int) string {
...
+		return pb.Checksum_CRC32C
+	case "xxhash64":
+		return pb.Checksum_XXHash64
+	default:
+		fmt.Printf("Invalid checksum algorithm : %s\n", ct)
+		os.Exit(1)
+	}
+	return pb.Checksum_CRC32C
+}
 func checksumVerificationMode(cvMode string) options.ChecksumVerificationMode {
```
mangalaman93, line 535: nit: a newline before this

options.go
```diff
@@ -7,6 +7,7 @@ package badger
 
 import (
 	"fmt"
+	"github.com/dgraph-io/badger/v4/pb"
```
mangalaman93, line 10: same here, please move it down along with other badger imports

options.go
```diff
@@ -93,6 +94,9 @@ type Options struct {
 	// ChecksumVerificationMode decides when db should verify checksums for SSTable blocks.
 	ChecksumVerificationMode options.ChecksumVerificationMode
 
+	//ChecksumAlgorithm decides which algorithm calculate checksums
```
mangalaman93, line 97: nit: space after //

badger/cmd/info.go
```diff
@@ -515,6 +520,18 @@ func pluralFiles(count int) string {
 	return "files"
 }
 
+func checksumAlgorithm(ct string) pb.Checksum_Algorithm {
```
mangalaman93, line 523: please rename this function to strToChecksumAlgorithm

badger/cmd/info.go
```diff
@@ -515,6 +520,18 @@ func pluralFiles(count int) string {
...
+	switch ct {
+	case "crc32c":
+		return pb.Checksum_CRC32C
+	case "xxhash64":
+		return pb.Checksum_XXHash64
+	default:
+		fmt.Printf("Invalid checksum algorithm : %s\n", ct)
+		os.Exit(1)
```
mangalaman93, line 531: can we use y.Error instead like everywhere else?

table/table.go
```diff
@@ -51,6 +51,9 @@ type Options struct {
 	// ChkMode is the checksum verification mode for Table.
 	ChkMode options.ChecksumVerificationMode
 
+	//ChkAlgo is the checksum algorithm mode for Table.
```
mangalaman93, line 54: nit: space after //

mangalaman93 (changes requested):

badger/cmd/info.go
```diff
@@ -8,6 +8,7 @@ package cmd
 import (
 	"bytes"
 	"encoding/hex"
+	stderrors "errors"
```
mangalaman93, line 11: there is no need to use the alias here

badger/cmd/info.go
```diff
@@ -40,10 +42,14 @@ type flagOptions struct {
...
 	discard                  bool
 	externalMagicVersion     uint16
+	checksumAlgorithm        string
 }
 
 var (
 	opt flagOptions
+
+	// ErrInvalidChecksumAlgorithm is returned if the checksum algorithm is invalid.
+	ErrInvalidChecksumAlgorithm = stderrors.New("Invalid checksum algorithm. Supported values: crc32c, xxhash64.")
```
mangalaman93, line 52: nit: the first letter of the error message should not be capitalized

badger/cmd/info.go
```diff
@@ -89,14 +97,20 @@ to the Dgraph team.
 
 func handleInfo(cmd *cobra.Command, args []string) error {
 	cvMode := checksumVerificationMode(opt.checksumVerificationMode)
+	ct, err := strToChecksumAlgorithm(opt.checksumAlgorithm)
+	if err != nil {
```
mangalaman93, line 101: there is no need for the if here, you could directly call y.Check(err) and it internally fails only when err is not nil.

badger/cmd/info.go
```diff
@@ -515,6 +529,19 @@ func pluralFiles(count int) string {
 	return "files"
 }
 
+// When the checkSum Algorithm is invalid, func strToChecksumAlgorithm will return the default checkSum Algorithm
```
mangalaman93, line 532: What is benefit of returning the default checksum when we are returning the error along with it? I think we should return default checksum with nil error

badger/cmd/info_test.go
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"fmt"
+	"github.com/dgraph-io/badger/v4/pb"
```
mangalaman93, line 21: please separate stdlib imports with non-stdlib imports like in other files

mangalaman93 (changes requested):
@LiuQhahah There is one issue I see here. That is, if I update the checksum algorithm on an existing badger directory, it is going to fail in that case. Can we handle that case so that for existing data we use the right checksum algorithm whereas for all the new data we use the new checksum algorithm?

LiuQhahah:
> @LiuQhahah There is one issue I see here. That is, if I update the checksum algorithm on an existing badger directory, it is going to fail in that case. Can we handle that case so that for existing data we use the right checksum algorithm whereas for all the new data we use the new checksum algorithm?

Following the current design, I'm unsure about how to proceed.

Currently, the options only provide a single value for checksumAlgorithm. If we want to support multiple checksum algorithms, it might mislead users since they wouldn’t know which algorithm was actually used.

I suggest sticking with the current design: one Badger directory should use only one checksum algorithm. If users need multi-checksum support, we can discuss it separately.

What do you think?

Thanks
Qiang

mangalaman93:
> > @LiuQhahah There is one issue I see here. That is, if I update the checksum algorithm on an existing badger directory, it is going to fail in that case. Can we handle that case so that for existing data we use the right checksum algorithm whereas for all the new data we use the new checksum algorithm?
> 
> Following the current design, I'm unsure about how to proceed.
> 
> Currently, the options only provide a single value for checksumAlgorithm. If we want to support multiple checksum algorithms, it might mislead users since they wouldn’t know which algorithm was actually used.
> 
> I suggest sticking with the current design: one Badger directory should use only one checksum algorithm. If users need multi-checksum support, we can discuss it separately.
> 
> What do you think?
> 
> Thanks Qiang

Seems like we already store checksum algorithm in the block. We should use that when reading, whereas when writing, we should use the checksum algorithm provided in the option.

We should also add a test for this scenario where we open badger with a algo1 for checksum, write and read some data, close badger. Open again with algo2 for checksum, read the old data, write some more new data and read new data. This ensures that we are using the right checksum algorithm.

LiuQhahah:
> > > @LiuQhahah There is one issue I see here. That is, if I update the checksum algorithm on an existing badger directory, it is going to fail in that case. Can we handle that case so that for existing data we use the right checksum algorithm whereas for all the new data we use the new checksum algorithm?
> > 
> > 
> > Following the current design, I'm unsure about how to proceed.
> > Currently, the options only provide a single value for checksumAlgorithm. If we want to support multiple checksum algorithms, it might mislead users since they wouldn’t know which algorithm was actually used.
> > I suggest sticking with the current design: one Badger directory should use only one checksum algorithm. If users need multi-checksum support, we can discuss it separately.
> > What do you think?
> > Thanks Qiang
> 
> Seems like we already store checksum algorithm in the block. We should use that when reading, whereas when writing, we should use the checksum algorithm provided in the option.
> 
> We should also add a test for this scenario where we open badger with a algo1 for checksum, write and read some data, close badger. Open again with algo2 for checksum, read the old data, write some more new data and read new data. This ensures that we are using the right checksum algorithm.

Thank you for your information.
I agree with you.

I will follow this task.
Thanks,
Qiang

PR #2198, opened 2025-05-11
fix crash loop on missing manifest tables

RJKeevil:
Please see https://discuss.hypermode.com/t/manifest-removes-non-existing-table/19882 

W are experiencing an error where dgraph tries to remove a table that is missing. I assume that given  it has already been removed, it should not trigger a crash loop in compaction. Instead i sense check that the table is indeed removed from all levels, and return without an error code. Please advise if you think this is safe or not.

**Checklist**

### DIFF
--- a/db2_test.go
+++ b/db2_test.go
@@ -427,24 +427,24 @@ func TestCompactionFilePicking(t *testing.T) {
 	for i := 1; i <= 10; i++ {
 		// Each table has difference of 1 between smallest and largest key.
 		tab := createTableWithRange(t, db, 2*i-1, 2*i)
-		addToManifest(t, db, tab, 3)
+		addToManifest(t, db, tab, 3, db.opt)
 		require.NoError(t, l3.replaceTables([]*table.Table{}, []*table.Table{tab}))
 	}
 
 	l2 := db.lc.levels[2]
 	// First table has keys 1 and 4.
 	tab := createTableWithRange(t, db, 1, 4)
-	addToManifest(t, db, tab, 2)
+	addToManifest(t, db, tab, 2, db.opt)
 	require.NoError(t, l2.replaceTables([]*table.Table{}, []*table.Table{tab}))
 
 	// Second table has keys 5 and 12.
 	tab = createTableWithRange(t, db, 5, 12)
-	addToManifest(t, db, tab, 2)
+	addToManifest(t, db, tab, 2, db.opt)
 	require.NoError(t, l2.replaceTables([]*table.Table{}, []*table.Table{tab}))
 
 	// Third table has keys 13 and 18.
 	tab = createTableWithRange(t, db, 13, 18)
-	addToManifest(t, db, tab, 2)
+	addToManifest(t, db, tab, 2, db.opt)
 	require.NoError(t, l2.replaceTables([]*table.Table{}, []*table.Table{tab}))
 
 	cdef := &compactDef{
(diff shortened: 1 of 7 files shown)


manifest.go
```diff
@@ -445,10 +445,13 @@ func applyManifestChange(build *Manifest, tc *pb.ManifestChange) error {
 	case pb.ManifestChange_DELETE:
 		tm, ok := build.Tables[tc.Id]
 		if !ok {
-			return fmt.Errorf("MANIFEST removes non-existing table %d", tc.Id)
+			for _, level := range build.Levels {
```
mangalaman93, line 448: Could we add a log here for FYI?

RJKeevil: @mangalaman93 of course, I would have done this but was a bit unsure of how you use the logger in badger.  Looks like all current warnings are thrown from an options object, should i cange the function signatures to pass the options object down to this function?

mangalaman93: yes, that seems alright I think.

RJKeevil: Ok I had an attempt, ive added it as a commit on top so you can check it separately, had to do some plumbing to get the options object in the right places. The most questionable chenge there is in manifest_test.go where i had to init a db object, hopefully that is correct?

db2_test.go
```diff
@@ -24,11 +24,12 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/dgraph-io/ristretto/v2/z"
+
```
mangalaman93, line 28: nit: no need for this change

levels.go
```diff
@@ -23,10 +23,11 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 
+	"github.com/dgraph-io/ristretto/v2/z"
```
mangalaman93, line 26: nit: same here

manifest.go
mangalaman93: @harshil-goel could you take a look at this as well. Looks fine to me.

manifest_test.go
```diff
@@ -198,7 +198,14 @@ func TestManifestRewrite(t *testing.T) {
 	require.NoError(t, err)
 	defer removeDir(dir)
 	deletionsThreshold := 10
-	mf, m, err := helpOpenOrCreateManifestFile(dir, false, 0, deletionsThreshold)
+
+	db, err := Open(DefaultOptions(dir))
+	require.NoError(t, err, "error while opening db")
+	defer func() {
```
mangalaman93, line 204: we can use t.Cleanup now too

mangalaman93:
> @mangalaman93 the test error seems related to the use of t.Cleanup, can you advise?

```
        defer removeDir(dir)
	deletionsThreshold := 10
	mf, m, err := helpOpenOrCreateManifestFile(dir, false, 0, deletionsThreshold)

	db, err := Open(DefaultOptions(dir))
	require.NoError(t, err, "error while opening db")

	t.Cleanup(func() {
		require.NoError(t, db.Close())
	})
```

it seems that defer cleaned up the directory and db.Close is not happy about it. I think we should use either `t.Cleanup`  everywhere in this test or defer.

RJKeevil:
I went back to defer for now to have control over their order, lets see if the tests now pass

RJKeevil:
and again, with the remove explicitly in the same defer as close db (since they are coupled anyway)

mangalaman93:
Thank you for the PR @RJKeevil

RJKeevil:
Thanks for the help @mangalaman93 !

PR #2211, opened 2025-07-15
Update stream framework with new alternate keyToList function

harshil-goel:


stream.go
```diff
@@ -75,6 +75,14 @@ type Stream struct {
 	//
 	// Note: Calls to KeyToList are concurrent.
 	KeyToList func(key []byte, itr *Iterator) (*pb.KVList, error)
+	// UseKeyToListWithThreadId is used to indicate that KeyToListWithThreadId should be used
```
mangalaman93, line 78: This comment needs to be simplified.

stream.go
```diff
@@ -252,6 +266,23 @@ func (st *Stream) produceKVs(ctx context.Context, threadId int) error {
 				}
 			}
 		}
+
+		if st.UseKeyToListWithThreadId {
+			if kvs, err := st.FinishThread(threadId); err != nil {
+				return err
+			} else {
+				for _, kv := range kvs.Kv {
```
mangalaman93, line 274: we should define this as a default function and assign it to FinishThread

stream.go
```diff
@@ -233,7 +241,13 @@ func (st *Stream) produceKVs(ctx context.Context, threadId int) error {
 
 			// Now convert to key value.
 			itr.Alloc.Reset()
-			list, err := st.KeyToList(item.KeyCopy(nil), itr)
+			var list *pb.KVList
+			var err error
+			if st.UseKeyToListWithThreadId {
+				list, err = st.KeyToListWithThreadId(item.KeyCopy(nil), itr, threadId)
```
mangalaman93, line 247: what's the benefit of passing the threadId?

mangalaman93 (commented):
We should add a test as well and considering getting rid of the boolean field that decides what function to use. We can simply check whether a field is nil.

PR #2215, opened 2025-07-18
fix: Subscribe API Meta and UserMeta fields wrong

fovecifer:
**Description**

I am using Subscribe API to listening the key changes, I found the Meta and UserMeta in pb.KV were wrong

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/dgraph-io/badger/v4
 
 go 1.23.0
 
-toolchain go1.24.4
+toolchain go1.24.5
 
 require (
 	github.com/cespare/xxhash/v2 v2.3.0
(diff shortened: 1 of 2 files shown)


mangalaman93:
Thank you for the PR. I do not think this is a bug. This is expected because the internal meta is not useful to the caller and hence, we write the UserMeta into meta before publishing.

fovecifer:
> Thank you for the PR. I do not think this is a bug. This is expected because the internal meta is not useful to the caller and hence, we write the UserMeta into meta before publishing.

Why not write the UserMeta into UserMeta? as a caller of the Subscribe API, I was confused where to get my UserMeta

fovecifer:
> Do you mind explaining why you are using a fork? Is this PR the reason for it? Thank you patiently answering my questions.

I want to implement a "Redis replication" like feature to the badger.
I am using "Backup" API to do a full synchronization when the mater/replica connected,
then I will broadcast the changes of master node keys to the replica nodes, so I am using the Subscribe API 
to obtain the changes of master node.
At first I found the UserMeta of "*badger.KVList" in the Subscribe was missing, it will lost some data of my 
master node. 
And I also need to ignore the changes of some keys, I found the bitDiscardEarlierVersions of the key meta, 
the Meta field of "*pb.KV" is begin with uppercase letter, so it could access outside the badger package.
I felt confused when I found the source code of Subscribe API, the UserMeta and Meta fields are both 
begin with uppercase letter, I think the design of badger library is providing all informations of the k-v pair
to all the callers.

PR #2265, opened 2026-03-20
Fixed nil dereference.

Aleksei-Badyaev:
Fixed a critical null pointer dereference vulnerability.

After having been compared to a nil value at stream_writer.go:274, pointer '(**sw.db).orc' is passed as implicit 'this' parameter in call to function 'github.com/dgraph-io/badger/v4.oracle.readTs' at stream_writer.go:278, where it is dereferenced at txn.go:79.


matthewmcneely:
@Aleksei-Badyaev Can you check the formatting failure in the trunk check run?

PR #2268, opened 2026-03-23
fix : read only user should be able to open db on ReadOnly mode 

### LINKED ISSUE #2234: Support opening database with read-only file permissions
## Is your feature request related to a problem? Please describe

When opening a database directory with `WithReadOnly`: https://pkg.go.dev/github.com/dgraph-io/badger/v4#Options.WithReadOnly
and when the user running badger only has read permissions for the database files,
then badger produces the following error:
```
err: while opening file: DISCARD
err: open db/DISCARD: permission denied
unable to open: db/DISCARD
```

## Describe the solution you'd like

Badger shouldn't try to write to (or create) the `DISCARD` file: https://github.com/dgraph-io/badger/blob/v4.7.0/discard.go#L29-L35 when in read-only mode.

Using `DISCARD` seems to be unavoidable currently, except for in-memory mode: https://github.com/dgraph-io/badger/blob/v4.7.0/value.go#L538-L544

## Describe alternatives you've considered

I'm currently working around this by loading a copy of the database from a writeable directory.
But ideally, I'd like to drop the workaround.

(If needed I can elaborate on why I consume a read-only database in the first place)

## Additional context

Found the PR that introduced the read-only mode:

> This change assumes that there will be no need to write anything during read-only operations, no need to do garbage collection or compaction or revising vlogs or anything.
> https://github.com/dgraph-io/badger/pull/432

And the commit that introduced "discard stats" as separate file:
> - Discard stats are now stored in a separate file, instead of within the LSM tree.
> - GC only picks up value logs based off discard stats.
> - GC no longer does sampling, it uses discard stats to inform when a value log needs to be GCed.
>
> https://github.com/dgraph-io/badger/commit/e3a0d29f75d9978d8230956c39f0dbdde2f1cb63

I might be wrong, but to me it looks like discard stats & GC weren't intended to be active during read-only mode.
And if they weren't, it seems like it would solve the issue I'm facing?

Just my rough guess - I didn't spend enough time with badger's internals to properly evaluate that.
I'd appreciate any help or input! :)

Naelpuissant:
**Description**

Should fix #2234 
Help appreciated :smile: 
Step to reproduce : 
- Open a db and fill it with some data
- run `chmod -w ./db/*`
- Open it again in `WithReadOnly` mode

**Checklist**

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -1742,8 +1742,6 @@ func TestTestSequence2(t *testing.T) {
 }
 
 func TestReadOnly(t *testing.T) {
-	t.Skipf("TODO: ReadOnly needs truncation, so this fails")
-
 	dir, err := os.MkdirTemp("", "badger-test")
 	require.NoError(t, err)
 	defer removeDir(dir)
--- a/db_test.go
+++ b/db_test.go
@@ -11,6 +11,7 @@ import (
 	"encoding/binary"
 	"flag"
 	"fmt"
+	"io/fs"
 	"math"
 	"math/rand"
 	"os"
@@ -1769,12 +1770,10 @@ func TestReadOnly(t *testing.T) {
 	opts.ReadOnly = true
 	kv1, err := Open(opts)
 	require.NoError(t, err)
-	defer kv1.Close()
 
 	// Open another read-only
 	kv2, err := Open(opts)
 	require.NoError(t, err)
-	defer kv2.Close()
 
 	// Attempt a read-write open while it's open for read-only
 	opts.ReadOnly = false
@@ -1790,6 +1789,10 @@ func TestReadOnly(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, b1, []byte("value1"))
 	err = txn1.Commit()
+	require.NoError(t, err)
+	err = kv1.RunValueLogGC(0.5)
+	require.Error(t, err, ErrGCInReadOnlyMode)
+	err = kv1.Sync()
 	require.NoError(t, err)
 
 	// Get a thing from the DB via the other connection
@@ -1809,6 +1812,32 @@ func TestReadOnly(t *testing.T) {
 	require.Contains(t, err.Error(), "No sets or deletes are allowed in a read-only transaction")
 	err = txn.Commit()
 	require.NoError(t, err)
+
+	// Close
+	require.NoError(t, kv1.Close())
+	require.NoError(t, kv2.Close())
+
+	// Test os permission read-only open
+	err = filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		require.NoError(t, err)
+		if path == dir {
+			return os.Chmod(path, 0o500)
+		}
+		return os.Chmod(path, 0o400)
+	})
+	require.NoError(t, err)
+
+	opts.ReadOnly = true
+	kv3, err := Open(opts)
+	require.NoError(t, err)
+	txn3 := kv3.NewTransaction(true)
+	_, err = txn3.Get([]byte("key1"))
+	require.NoError(t, err)
+	require.NoError(t, txn3.Commit())
+	require.NoError(t, kv3.Close())
+
+	// Restore permissions for cleanup
+	require.NoError(t, os.Chmod(dir, 0o700))
 }
 
 func TestLSMOnly(t *testing.T) {
@@ -1843,7 +1872,6 @@ func TestLSMOnly(t *testing.T) {
 		txnSet(t, db, []byte(fmt.Sprintf("key%d", i)), value, 0x00)
 	}
 	require.NoError(t, db.Close())
-
 }
 
 // This test function is doing some intricate sorcery.
--- a/db.go
+++ b/db.go
@@ -378,7 +378,7 @@ func Open(opt Options) (*DB, error) {
 	db.closers.writes = z.NewCloser(1)
 	go db.doWrites(db.closers.writes)
 
-	if !db.opt.InMemory {
+	if !db.opt.InMemory && !db.opt.ReadOnly {
 		db.closers.valueGC = z.NewCloser(1)
 		go db.vlog.waitOnGC(db.closers.valueGC)
 	}
@@ -538,7 +538,7 @@ func (db *DB) close() (err error) {
 	db.blockWrites.Store(1)
 	db.isClosed.Store(1)
 
-	if !db.opt.InMemory && !db.opt.ReadOnly {
+	if db.closers.valueGC != nil {
 		// Stop value GC first.
 		db.closers.valueGC.SignalAndWait()
 	}
@@ -671,8 +671,8 @@ const (
 // Sync syncs database content to disk. This function provides
 // more control to user to sync data whenever required.
 func (db *DB) Sync() error {
-	if db.opt.InMemory {
-		// InMemory mode does not use WAL/vlog files, so Sync is a no-op.
+	if db.opt.InMemory || db.opt.ReadOnly {
+		// InMemory and read-only modes do not use WAL/vlog files, so Sync is a no-op.
 		return nil
 	}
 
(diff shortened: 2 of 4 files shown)


matthewmcneely (changes requested):
## Critical: goroutine leak on ReadOnly close

[`db.go:381-384`](https://github.com/Naelpuissant/badger/blob/fix-read-only/db.go#L381-L384) is **unchanged** by the PR:

```go
if !db.opt.InMemory {
    db.closers.valueGC = z.NewCloser(1)
    go db.vlog.waitOnGC(db.closers.valueGC)
}
```

So when `ReadOnly=true && InMemory=false`, the GC closer **and goroutine still get created**, but with the new [`db.go:541`](https://github.com/Naelpuissant/badger/blob/fix-read-only/db.go#L541) guard the close path now skips `SignalAndWait()`. The goroutine is therefore leaked on every `db.Close()`.

Worse, [`value.go:1080`](https://github.com/Naelpuissant/badger/blob/fix-read-only/value.go#L1080) does `vlog.garbageCh <- struct{}{}` after the close signal — and the PR makes `garbageCh` nil in ReadOnly via the `init` early‑return. Even if the closer were signaled, the goroutine would deadlock forever on a nil‑channel send.

The same hazard exists in the failure path: [`db.go:484-486`](https://github.com/Naelpuissant/badger/blob/fix-read-only/db.go#L484-L486) `cleanup()` calls `valueGC.Signal()`, which would unblock `waitOnGC` and then hang on the nil `garbageCh`.

The commit message claims to "avoid unused GC initialization", but the un‑skipped goroutine spawn contradicts that intent. The cleaner fix is to gate the goroutine spawn itself:

```go
if !db.opt.InMemory && !db.opt.ReadOnly {
    db.closers.valueGC = z.NewCloser(1)
    go db.vlog.waitOnGC(db.closers.valueGC)
}
```

…and then the new guard in `db.close()` becomes redundant (the existing `cleanup()` nil‑check at [`db.go:484`](https://github.com/Naelpuissant/badger/blob/fix-read-only/db.go#L484) is already safe). This makes the three code sites (`init`, spawn, close) consistent: ReadOnly = no GC anywhere.

## Other issues

- **`RunValueLogGC` on a ReadOnly DB** ([`db.go:1234`](https://github.com/Naelpuissant/badger/blob/fix-read-only/db.go#L1234)) doesn't reject early. With this PR, calling it bottoms out in [`value.go:1085`](https://github.com/dgraph-io/badger/blob/fix-read-only/value.go#L1085):

  ```go
  select {
  case vlog.garbageCh <- struct{}{}:
      ...
  default:
      return ErrRejected
  }
  ```

  Sending to a nil channel in a `select` with `default` falls to `default`, so it accidentally returns `ErrRejected` rather than panicking. It's correct by luck, not design. Worth either adding `if db.opt.ReadOnly { return ErrRejected }` (or a dedicated error) at the top of `RunValueLogGC`, or documenting the invariant.

- **`updateDiscardStats`** ([`value.go:1101`](https://github.com/Naelpuissant/badger/blob/fix-read-only/value.go#L1101)) only short‑circuits on `InMemory`, then dereferences `vlog.discardStats` — which is now nil under ReadOnly. In practice it isn't reached because compactions don't run in ReadOnly mode (gated at [`db.go:343`](https://github.com/Naelpuissant/badger/blob/fix-read-only/db.go#L343)), but a defensive `|| vlog.opt.ReadOnly` here would match the symmetric guards added elsewhere and remove a footgun.

- **Pre‑existing, related**: [`db.go:706`](https://github.com/Naelpuissant/badger/blob/fix-read-only/db.go#L706) `db.Sync()` calls `db.mt.SyncWAL()` and `db.mt` is nil in ReadOnly per [`db.go:329-333`](https://github.com/Naelpuissant/badger/blob/fix-read-only/db.go#L329-L333). Not in this PR's scope, but un‑skipping `TestReadOnly` increases the chance someone exercises this. Worth a follow‑up similar to [#2264](https://github.com/dgraph-io/badger/pull/2264) (the InMemory NPE‑on‑sync fix).

## Test coverage

[`TestReadOnly`](https://github.com/Naelpuissant/badger/blob/fix-read-only/db_test.go#L1744) exercises the happy path (concurrent RO opens, rejected RW‑while‑RO open, attempted write) but does not cover:

- `RunValueLogGC` on a ReadOnly DB (would have caught the goroutine/nil‑channel hazard above)
- `Sync()` on a ReadOnly DB
- Goroutine leak verification (no `goleak` check, so the leak is invisible in CI)

Adding at least the `RunValueLogGC` case would have flagged the inconsistency in the GC initialization gating.

## Recommended changes before merge

1. Gate the `valueGC` closer/goroutine creation at [`db.go:381`](https://github.com/Naelpuissant/badger/blob/fix-read-only/db.go#L381) on `!ReadOnly` as well, and revert the [`db.go:541`](https://github.com/dgraph-io/badger/blob/fix-read-only/db.go#L541) change as redundant.
2. Add an early `ErrRejected` (or a dedicated `ErrReadOnly`) in `RunValueLogGC` for ReadOnly DBs.
3. *Optional*: extend `TestReadOnly` to call `db.RunValueLogGC(0.5)` and `db.Sync()` and assert they don't panic.

Naelpuissant:
@matthewmcneely Thanks for your reply, I'm working on it !

> the new guard in db.close() becomes redundant (the existing cleanup() nil‑check at [db.go:484](https://github.com/Naelpuissant/badger/blob/fix-read-only/db.go#L484) is already safe).

Since cleanup is only called on Open, how the close guard would be redundant ?

matthewmcneely:
> @matthewmcneely Thanks for your reply, I'm working on it !
> 
> > the new guard in db.close() becomes redundant (the existing cleanup() nil‑check at [db.go:484](https://github.com/Naelpuissant/badger/blob/fix-read-only/db.go#L484) is already safe).
> 
> Since cleanup is only called on Open, how the close guard would be redundant ?

Whoops, I mistakenly thought cleanup() and db.close() were the same thing. If you gate the spawn with:

```
- if !db.opt.InMemory {
+ if !db.opt.InMemory && !db.opt.ReadOnly {
      db.closers.valueGC = z.NewCloser(1)
      go db.vlog.waitOnGC(db.closers.valueGC)
  }
```

and then replace your `!ReadOnly` line 541 with a nil‑check of valeuGC:

```
- if !db.opt.InMemory && !db.opt.ReadOnly {
-     // Stop value GC first.
-     db.closers.valueGC.SignalAndWait()
+ if db.closers.valueGC != nil {
+     // Stop value GC first.
+     db.closers.valueGC.SignalAndWait()
  }

```

Then I think that seems to be cleanest.

Naelpuissant:
@matthewmcneely Ok I see, it should be fine now. 
I've also added more guards on Flatten, GetSequence and BanNamespace since those methods might mutate the DB state.

db_test.go
```diff
@@ -1811,6 +1811,26 @@ func TestReadOnly(t *testing.T) {
...
+	err = os.Chmod(dir, 0o500)
+	require.NoError(t, err)
+	opts.ReadOnly = true
+	kv3, err := Open(opts)
+	require.NoError(t, err)
+	txn3 := kv3.NewTransaction(true)
+	_, err = txn3.Get([]byte("key1"))
+	require.NoError(t, err)
+	require.NoError(t, kv3.Close())
+	require.NoError(t, txn3.Commit())
```
matthewmcneely, line 1836: Swap 1830 with 1829, at present commit is called after close, which I guess works because txn3 performed no writes.

db_test.go
```diff
@@ -1811,6 +1811,26 @@ func TestReadOnly(t *testing.T) {
 	require.Contains(t, err.Error(), "No sets or deletes are allowed in a read-only transaction")
 	err = txn.Commit()
 	require.NoError(t, err)
+
+	// Close
+	require.NoError(t, kv1.Close())
+	require.NoError(t, kv2.Close())
+
+	// Test os permission read-only open
+	// We don't directly chmod the directory to still be able to aquire the lock
```
matthewmcneely: Typo: "acquire"

matthewmcneely (changes requested):
The PR description's repro is chmod -w ./db/* (strip write from the files), but the test instead does chmod 0o500 on the directory. That blocks file create/delete but leaves the existing data files writable, so the new O_RDONLY flag at [value.go:580-582] isn't actually exercised — O_RDWR would still succeed on those files. The test does cover the empty-vlog-deletion guard and the init() early-return (both need the dir to be unwritable), but a follow-up chmod 0o400 on each *.vlog would directly exercise the O_RDONLY change.

PR #2269, opened 2026-03-24
perf: skip lsm lookup for expired entries during value log rewrite

lamb007:
## PR Description

This PR optimizes value log GC rewrite by avoiding the `db.get(e.Key)` lookup when the value log entry is already deleted or expired.

Currently, `rewrite` always reads the current value from the LSM tree before calling `discardEntry`. For entries that are already deleted or expired in the value log, that lookup is unnecessary:

- If the same version is still returned from the LSM tree, it will also be considered deleted or expired, so no read path will need to resolve this value through the value log and the entry can be discarded.
- If the LSM tree no longer contains that version, the value log entry is unreferenced and can be discarded directly.y.

To avoid this unnecessary work, this change moves the deleted/expired check earlier in rewrite, before `db.get(e.Key)`, by switching the check from the LSM-returned `ValueStruct` to the value log Entry itself, and removes the corresponding deleted/expired check from `discardEntry`.

### Benchmark

A standalone benchmark was also added to measure the impact of this optimization on an expired-heavy rewrite workload. The benchmark builds the same fixed dataset for both implementations and measures a single `vlog.rewrite(lf)` call on the oldest expired-only vlog file.

Benchmark command:

```bash
go test ./... -run '^$' -bench 'BenchmarkValueGCRewriteExpiredOnlyFile$' -benchmem -benchtime=5x -count=1
```

Benchmark results on my machine:

- Optimized:
  - `10938658 ns/op`
  - `20301009 B/op`
  - `90138 allocs/op`
- Original:
  - `21713735 ns/op`
  - `24000905 B/op`
  - `172563 allocs/op`

In this workload, the optimized version reduces rewrite time by about 2x and also lowers allocation count.

### DIFF
--- a/value_gc_rewrite_bench_test.go
+++ b/value_gc_rewrite_bench_test.go
@@ -1,0 +1,150 @@
+/*
+ * SPDX-FileCopyrightText: © 2017-2025 Istari Digital, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package badger
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	benchmarkRewriteExpiredKeyCount         = 8192
+	benchmarkRewriteLiveKeyCount            = 4096
+	benchmarkRewriteValueSize               = 2 << 10
+	benchmarkRewriteValueLogFileSize        = 8 << 20
+	benchmarkRewriteValueThreshold          = 128
+	benchmarkRewriteEntryExpiry      uint64 = 1
+)
+
+func benchmarkValueGCRewriteOptions(dir string) Options {
+	opt := getTestOptions(dir)
+	opt.ValueLogFileSize = benchmarkRewriteValueLogFileSize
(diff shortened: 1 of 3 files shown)


value_gc_rewrite_bench_test.go
```diff
@@ -0,0 +1,145 @@
+package badger
```
matthewmcneely, line 6: Could you add the header?

```
/*
 * SPDX-FileCopyrightText: © 2017-2025 Istari Digital, Inc.
 * SPDX-License-Identifier: Apache-2.0
 */
```

value_test.go
```diff
@@ -896,6 +898,71 @@ func (th *testHelper) writeRange(from, to int) {
 	}
 }
 
+func TestValueGCRewriteSkipsLSMGetForExpiredVlogEntry(t *testing.T) {
```
matthewmcneely: It might be good to add a vlog file containing a mix of expired **_and_** live entries — i.e. the case where the new early-return branch fires for some entries while db.get(e.Key) + discardEntry(...) fires for others in the same file--just for completeness.

matthewmcneely (changes requested):

PR #2278, opened 2026-05-22
fix: fix an issue where the compactor causes previously deleted data to reappear

yangchun0821:
**Description**

During Level 0 compaction, the engine may skip intermediate levels and compact Level 0 tables directly to the bottommost level. This can leave stale data in the skipped intermediate levels. As a result, previously deleted keys may reappear from the user's perspective after compaction completes.

**Checklist**

### DIFF
--- a/levels.go
+++ b/levels.go
@@ -385,10 +385,26 @@ func (s *levelsController) levelTargets() targets {
 	for i := len(s.levels) - 1; i > 0; i-- {
 		ltarget := adjust(dbSize)
 		t.targetSz[i] = ltarget
-		if t.baseLevel == 0 && ltarget <= s.kv.opt.BaseLevelSize {
-			t.baseLevel = i
-		}
 		dbSize /= int64(s.kv.opt.LevelSizeMultiplier)
+	}
+
+	// Bring the base level down to the last empty level.
+	t.baseLevel = 1
+	for i := 1; i < len(s.levels)-1; i++ {
+		if s.levels[i].getTotalSize() > 0 {
+			break
+		}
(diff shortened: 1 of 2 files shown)


shiva-istari (changes requested):
Hi @yangchun0821 
Thanks for the PR and the description!

  The current TestLevelTargets sub-tests only assert which baseLevel is chosen they don't verify what actually happens to the data. Could you add tests that set up the full compaction scenario and assert the
  deleted key stays gone? These tests should fail on main (old code routes tombstone to wrong level, key reappears) and pass on this branch (fix routes tombstone correctly, key stays gone).

yangchun0821:
@shiva-istari 
Thank you for your review. I’ve addressed your comments by adding the new test case you recommended.

matthewmcneely:
Verified the regression test empirically. I ported `levels_test.go` from this branch onto `main` (keeping `main`'s `levels.go`) and ran it:

| | `TestDeletedKeyReappears` |
|---|---|
| this branch | passes |
| `main` | fails: `foo` reappears, `Get` returns `nil` instead of `ErrKeyNotFound` |

That is exactly the fails-on-`main` / passes-on-branch behavior that was asked for. The test now asserts the user-visible outcome (the deleted key stays gone) rather than just which `baseLevel` is chosen, so the review concern is addressed.

One housekeeping note: the review is still marked `CHANGES_REQUESTED` from before these commits landed, so it just needs a re-review to clear the block.

matthewmcneely (approved):

shiva-istari (approved):

PR #2295, opened 2026-06-05
fix(in-memory): resolve orphaned skiplist arena leak on Close

### LINKED ISSUE #2205: In memory databases leak memory on Close/DropAll due to orphaned skiplists
## SkipLists leak on various destructive operations primarily used in testing

When a user like myself creates an in-memory badgerDB for use in test the underlying Open call creates a skl.NewSkipList with a rather large skl.newArena that is impossible for the garbage collector to find and remove.  The underlying code that "cleans up" the skiplist when the reference count becomes 0 simply drops the head node, but this is a skiplist with each node retaining reference to other nodes (sometimes multiple references).  I reproduced and profiled multiple cases in our code where running tests with `go test -count=1000 ./` these lost skipLists quickly accumulated and caused the test to crash.  Adding a DropAll only made things worse, causing an additional lost skiplist.  I also confirmed that adding a runtime.GC() call to the end of our tests did nothing to find and remove this leaked memory. 

## To Reproduce

1. Create a loop that creates an in-memory db, writes to it once, closes it
2. profile with pprof
3. observe the leak

## Expected behavior

The only way to assure that the skiplist can be dereferenced completely is to walk the list destrying and unreferencing all adjacent nodes in the list (eg walk the list and delete everything).  Simply setting `head = nil` will not suffice. 

## Screenshots

pprof output from a test that does the above 1000 times (no example code, this is unfortunately part of our proprietary system).  If you need me to create an example code snippet I can circle back.

Image

## Environment

- OS: Tested on Fedora 6.14.8-300.fc42.x86_64

parthbhanti22:
**Description**

Fixes #2205.

**Root Cause:**
When Badger is configured with `opt.InMemory = true`, the `db.close()` function triggers an early return. This prevents the cleanup logic from iterating over the immutable memtables slice (`db.imm`). Because `mmap` allocations bypass Go's GC, failing to explicitly call `DecrRef()` on these memtables leaves the `skl.Arena` allocations orphaned in memory indefinitely.

**The Fix:**
Added an explicit iteration over `db.imm` to decrement reference counts on all immutable memtables prior to the early return in the in-memory shutdown sequence. 

**Verification:**
Included `issue_2205_test.go` to aggressively cycle in-memory database instantiation. 
* **Before:** `pprof` `inuse_space` showed ~83GB of retained `skl.newArena` allocations across 1000 iterations.
* **After:** Retained arenas dropped to 0 bytes. 

All tests in `./test.sh` pass locally.

**Checklist**

ppof1
ppof2

### DIFF
--- a/db.go
+++ b/db.go
@@ -625,6 +625,9 @@ func (db *DB) close() (err error) {
 	db.threshold.close()
 
 	if db.opt.InMemory {
+		for _, mem := range db.imm{
+			mem.DecrRef()
+		}
 		return
 	}
 
--- a/issue_2205_test.go
+++ b/issue_2205_test.go
@@ -1,0 +1,34 @@
+package badger
+
+import (
+	"testing"
+)
+
+func TestIssue2205_MemoryLeak(t *testing.T) {
+	// Loop 1000 times to force the leak to accumulate
+	for i := 0; i < 1000; i++ {
+		// Initialize an in-memory database configuration
+		opt := DefaultOptions("").WithInMemory(true)
+		opt.Logger = nil // Turn off logging so our terminal doesn't get spammed
+
+		db, err := Open(opt)
+		if err != nil {
+			t.Fatalf("Failed to open DB on iteration %d: %v", i, err)
+		}
+
+		// Perform a single write transaction
+		err = db.Update(func(txn *Txn) error {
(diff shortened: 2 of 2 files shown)


parthbhanti22:
@DocSavage  @deepakjois  @fjl 
Pl review

PR #2297, opened 2026-06-08
feat(txn): add Txn.MultiGet for batched all-versions point reads

shaunpatterson:
## What

Adds `Txn.MultiGet(keys [][]byte) ([]KeyResult, error)` — a batched, all-versions point read.

For each requested key it returns the full version chain (commit ts ≤ the txn's read ts, newest-first) — exactly what a per-key `NewKeyIterator(AllVersions)` + `Seek` + walk yields today — but **amortizes iterator construction** (`getMemTables`, per-level table iterators, the merge iterator) across the whole batch instead of paying it once per key. Keys are visited in sorted order so the shared iterator only seeks forward and adjacent keys reuse decoded SSTable blocks; input order is restored before returning.

It returns *all* versions (not just the newest value like `Txn.Get`) so callers can fold delta records on top of a complete record — which is exactly why the cheap point-get path is unusable for them.

## Why

Point-read-heavy, fan-out workloads issue many independent single-key reads per operation, strictly serially, each paying full iterator setup. The motivating case is dgraph's HNSW vector search: a query touches many posting-list keys (neighbor vectors/edges), and a candidate's sibling reads are independent and batchable.

## Semantics

Match the existing iterator path: read-ts filtering, badger-internal/banned-key skipping, and (for a read-write txn) read-set tracking for conflict detection — just as `NewKeyIterator(key).Seek(key)` would. `KeyResult.Versions` is empty for an absent key (no separate per-key not-found error). `ItemVersion` mirrors `Item` (`IsDeletedOrExpired`, `DiscardEarlierVersions`); values are materialized/copied so they stay valid after the call and after the txn is discarded.

## Testing

- **Differential test** (`TestMultiGetMatchesKeyIterator`, memtable + on-disk): asserts `MultiGet` returns byte-identical version chains to per-key `NewKeyIterator` over randomized data — sets, deletes, TTLs, discard-earlier markers, value-log-sized values, and absent/empty keys.
- `TestMultiGetReadTs`, `TestMultiGetEmpty`.
- **Benchmark** `BenchmarkMultiGetVsKeyIterator` (frontier of K dense keys), benchstat n=6:

| K | time | bytes | allocs |
|---|------|-------|--------|
| 16  | −5.4%  | −13.0% | −35.1% |
| 64  | −12.4% | −16.1% | −38.8% |
| 256 | −16.0% | −16.6% | −39.7% |

Additive change — no existing path is modified.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -311,6 +311,13 @@ type IteratorOptions struct {
 	Reverse        bool // Direction of iteration. False is forward, true is backward.
 	AllVersions    bool // Fetch all valid versions of the same key.
 	InternalAccess bool // Used to allow internal access to badger keys.
+	// NoReadTracking disables the per-Item/Seek conflict-detection read
+	// tracking that read-write transactions use to detect concurrent
+	// writes. Set this when the caller knows exactly which keys to record
+	// (e.g. Txn.MultiGet, which walks arbitrary intermediate keys but
+	// only cares about the requested keys). Has no effect on read-only
+	// transactions, where addReadKey is already a no-op.
+	NoReadTracking bool
 
 	// The following option is used to narrow down the SSTables that iterator
 	// picks up. If Prefix is specified, only tables which could have this
@@ -503,6 +510,139 @@ func (txn *Txn) NewKeyIterator(key []byte, opt IteratorOptions) *Iterator {
 	return txn.NewIterator(opt)
 }
 
+// ItemVersion is a single MVCC version of a key, as returned by Txn.MultiGet.
+//
+// It is a flattened, self-contained snapshot of one badger.Item: Value has
+// already been materialized and copied (so it stays valid after MultiGet
+// returns and after the transaction is discarded), and the metadata mirrors
+// the accessor methods on badger.Item.
+type ItemVersion struct {
+	// Value is a copy of the version's value bytes (value-log reads are
+	// resolved during MultiGet). It is owned by the caller.
+	Value []byte
+	// Version is the commit timestamp of this version.
+	Version uint64
+	// ExpiresAt is the Unix expiry time (0 if the entry never expires).
+	ExpiresAt uint64
+	// UserMeta is the user-supplied meta byte (e.g. dgraph posting-list bits).
+	UserMeta byte
+	// meta holds badger's internal meta bits (bitDelete, etc.). Unexported
+	// on purpose; use IsDeletedOrExpired / DiscardEarlierVersions to
+	// interpret it.
+	meta byte
+}
+
+// IsDeletedOrExpired mirrors (*Item).IsDeletedOrExpired for a MultiGet version.
+func (iv *ItemVersion) IsDeletedOrExpired() bool {
+	if iv.meta&bitDelete != 0 {
+		return true
+	}
+	if iv.ExpiresAt == 0 {
+		return false
+	}
+	return iv.ExpiresAt <= uint64(time.Now().Unix())
+}
+
+// DiscardEarlierVersions mirrors (*Item).DiscardEarlierVersions for a version.
+func (iv *ItemVersion) DiscardEarlierVersions() bool {
+	return iv.meta&bitDiscardEarlierVersions != 0
+}
+
+// KeyResult holds all versions (commit ts <= the transaction's read ts) of
+// one requested key, ordered newest-first — the same order an AllVersions
+// iterator yields. Versions is empty if the key does not exist (or has no
+// version visible at the read ts); there is no separate not-found error
+// per key.
+type KeyResult struct {
+	Versions []ItemVersion
+}
+
+// MultiGet reads many keys in a single iterator pass at the transaction's
+// read timestamp. For each requested key it returns the full version chain
+// (commit ts <= read ts), newest-first — exactly what an AllVersions
+// NewKeyIterator(key) + Seek + walk would yield, but amortizing iterator
+// construction (getMemTables, per-level table iterators, the merge iterator)
+// across the whole batch instead of paying it once per key.
+//
+// Intended for point-read-heavy, fan-out workloads such as dgraph's HNSW
+// vector search, where a query touches many independent posting-list keys
+// and the per-key NewKeyIterator setup dominates. Returning every version
+// (rather than the single newest value, as Txn.Get does) is required so
+// callers can fold delta postings on top of a complete posting, the reason
+// dgraph cannot use the cheap point-get path.
+//
+// The result slice is in the same order as keys. Internally the keys are
+// visited in sorted order so the shared iterator only seeks forward, which
+// lets adjacent keys reuse decoded SSTable blocks; the original order is
+// restored before returning.
+//
+// Semantics match the existing iterator path: versions newer than the read
+// ts are skipped, badger-internal keys and banned-namespace keys are
+// filtered. For a read-write transaction, only the requested keys are added
+// to the read set for conflict detection — exactly as NewKeyIterator(key)
+// would do — even though the shared iterator walks arbitrary intermediate
+// keys; NoReadTracking suppresses per-Item/Seek tracking and the outer loop
+// records each requested key once.
+func (txn *Txn) MultiGet(keys [][]byte) ([]KeyResult, error) {
+	if txn.discarded {
+		return nil, ErrDiscardedTxn
+	}
+	res := make([]KeyResult, len(keys))
+	if len(keys) == 0 {
+		return res, nil
+	}
+
+	// Visit keys in sorted order so the shared iterator only seeks forward.
+	order := make([]int, len(keys))
+	for i := range order {
+		order[i] = i
+	}
+	sort.Slice(order, func(a, b int) bool {
+		return bytes.Compare(keys[order[a]], keys[order[b]]) < 0
+	})
+
+	opt := DefaultIteratorOptions
+	opt.AllVersions = true
+	opt.PrefetchValues = false
+	opt.NoReadTracking = true
+	it := txn.NewIterator(opt)
+	defer it.Close()
+
+	for _, idx := range order {
+		key := keys[idx]
+		if len(key) == 0 {
+			continue
+		}
+		// Record the requested key for conflict detection. The iterator
+		// itself does not track reads because NoReadTracking is set; it
+		// would otherwise record every walked key.
+		txn.addReadKey(key)
+		var versions []ItemVersion
+		for it.Seek(key); it.Valid(); it.Next() {
+			item := it.Item()
+			if !bytes.Equal(item.Key(), key) {
+				break
+			}
+			iv := ItemVersion{
+				Version:   item.version,
+				ExpiresAt: item.expiresAt,
+				UserMeta:  item.userMeta,
+				meta:      item.meta,
+			}
+			// Materialize the value now; the iterator (PrefetchValues=false)
+			// would otherwise read it lazily and recycle the item on Next.
+			val, err := item.ValueCopy(nil)
+			if err != nil {
+				return nil, err
+			}
+			iv.Value = val
+			versions = append(versions, iv)
+		}
+		res[idx].Versions = versions
+	}
+	return res, nil
+}
+
 func (it *Iterator) newItem() *Item {
 	item := it.waste.pop()
 	if item == nil {
@@ -515,7 +655,9 @@ func (it *Iterator) newItem() *Item {
 // This item is only valid until it.Next() gets called.
 func (it *Iterator) Item() *Item {
 	tx := it.txn
-	tx.addReadKey(it.item.Key())
+	if !it.opt.NoReadTracking {
+		tx.addReadKey(it.item.Key())
+	}
 	return it.item
 }
 
@@ -750,7 +892,7 @@ func (it *Iterator) Seek(key []byte) {
 	if it.iitr == nil {
 		return
 	}
-	if len(key) > 0 {
+	if len(key) > 0 && !it.opt.NoReadTracking {
 		it.txn.addReadKey(key)
 	}
 	for i := it.data.pop(); i != nil; i = it.data.pop() {
--- a/iterator.go
+++ b/iterator.go
@@ -311,14 +311,6 @@ type IteratorOptions struct {
 	Reverse        bool // Direction of iteration. False is forward, true is backward.
 	AllVersions    bool // Fetch all valid versions of the same key.
 	InternalAccess bool // Used to allow internal access to badger keys.
-	// noReadTracking disables the per-Item/Seek conflict-detection read
-	// tracking that read-write transactions use to detect concurrent
-	// writes. Set this when the caller knows exactly which keys to record
-	// (e.g. Txn.MultiGet, which walks arbitrary intermediate keys but
-	// only cares about the requested keys). Has no effect on read-only
-	// transactions, where addReadKey is already a no-op.
-	noReadTracking bool
-
 	// The following option is used to narrow down the SSTables that iterator
 	// picks up. If Prefix is specified, only tables which could have this
 	// prefix are picked based on their range of keys.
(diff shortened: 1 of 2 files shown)


iterator.go
```diff
@@ -311,6 +311,13 @@ type IteratorOptions struct {
 	Reverse        bool // Direction of iteration. False is forward, true is backward.
 	AllVersions    bool // Fetch all valid versions of the same key.
 	InternalAccess bool // Used to allow internal access to badger keys.
+	// NoReadTracking disables the per-Item/Seek conflict-detection read
+	// tracking that read-write transactions use to detect concurrent
+	// writes. Set this when the caller knows exactly which keys to record
+	// (e.g. Txn.MultiGet, which walks arbitrary intermediate keys but
+	// only cares about the requested keys). Has no effect on read-only
+	// transactions, where addReadKey is already a no-op.
+	NoReadTracking bool
```
matthewmcneely: Can this be unexported?

As a public field, the only thing `NoReadTracking` does is silently disable SSI conflict detection for a read-write txn. Set it on a plain `NewIterator` and you get lost updates with no error — which is the same class of bug the commit message describes fixing here. There's no way for a caller to discover they've done it wrong.

`MultiGet` is in package `badger`, so `noReadTracking` works identically, and `prefixIsKey` a few lines below is the precedent for an unexported knob in this exact struct. Exported, it's part of the v4 surface forever.

iterator.go
```diff
@@ -503,6 +510,139 @@ func (txn *Txn) NewKeyIterator(key []byte, opt IteratorOptions) *Iterator {
...
+	opt := DefaultIteratorOptions
+	opt.AllVersions = true
+	opt.PrefetchValues = false
+	opt.NoReadTracking = true
+	it := txn.NewIterator(opt)
+	defer it.Close()
+
+	for _, idx := range order {
+		key := keys[idx]
+		if len(key) == 0 {
```
matthewmcneely: This silently diverges from `Txn.Get` on two error cases, and the name puts `MultiGet` squarely in the `Get` family, so callers will expect `Get`'s behavior.

- Empty key: `Txn.Get` returns `ErrEmptyKey` (txn.go:435). Here it's skipped and the caller gets an empty `Versions`, indistinguishable from absent. A caller with a bug that puts an empty key in a batch sees a plausible-looking not-found instead of an error.
- Banned namespace: `Txn.Get` returns `ErrBannedKey` (txn.go:441). Here it's filtered down in `parseItem` (iterator.go:778), so it fails closed — nothing leaks — but the caller can't tell banned from absent.

The banned case is the one that matters for dgraph. It branches on `badger.ErrBannedKey` explicitly in `posting/index.go:789`, `worker/draft.go:678`, `worker/draft.go:984`, and `schema/schema.go:540`. Under `MultiGet` those branches go unreachable and a banned namespace reads back as an empty posting list.

An explicit `isBanned` + empty-key check in this loop would restore parity. Whichever way you go, the docstring should say what happens to an empty key — the skip itself is necessary, since `it.Seek(nil)` would rewind.

iterator.go
```diff
@@ -503,6 +510,139 @@ func (txn *Txn) NewKeyIterator(key []byte, opt IteratorOptions) *Iterator {
...
+			}
+			iv := ItemVersion{
+				Version:   item.version,
+				ExpiresAt: item.expiresAt,
+				UserMeta:  item.userMeta,
+				meta:      item.meta,
+			}
+			// Materialize the value now; the iterator (PrefetchValues=false)
+			// would otherwise read it lazily and recycle the item on Next.
+			val, err := item.ValueCopy(nil)
```
matthewmcneely, line 676: This materializes and copies *every* version <= readTs, with no way for the caller to stop early. Dgraph's `ReadPostingList` (posting/mvcc.go:694-730) does the opposite — it breaks on `IsDeletedOrExpired`, returns on `BitEmptyPosting` without touching the value at all, and returns on `BitCompletePosting`.

So for a key with a complete posting sitting on a delta chain — the shape that trips incremental rollup at `deltaCount > 500` — today's path reads one item and this reads and vlog-resolves all of them. That's the workload the PR is aimed at, and it's the case where `MultiGet` is slower.

The benchmark won't surface it: uniform 3 versions, 64-byte values, no vlog resolution. Could you add an arm with a deep chain and vlog-sized values so the crossover shows up? And some way to bound the walk — a per-version predicate, or a `MaxVersionsPerKey`.

Separately, and I think inherent rather than fixable: dgraph unmarshals deltas via `item.Value(cb)` with no copy, whereas `[]ItemVersion` mandates a copy per version because the item is recycled on `Next`. Worth noting that the benchmark's `KeyIterator` baseline also calls `ValueCopy`, so it isn't quite the code being replaced.

iterator.go
```diff
@@ -503,6 +510,139 @@ func (txn *Txn) NewKeyIterator(key []byte, opt IteratorOptions) *Iterator {
...
+func (iv *ItemVersion) DiscardEarlierVersions() bool {
+	return iv.meta&bitDiscardEarlierVersions != 0
+}
+
+// KeyResult holds all versions (commit ts <= the transaction's read ts) of
+// one requested key, ordered newest-first — the same order an AllVersions
+// iterator yields. Versions is empty if the key does not exist (or has no
+// version visible at the read ts); there is no separate not-found error
+// per key.
+type KeyResult struct {
```
matthewmcneely, line 568: Now that `Key` is dropped, `KeyResult` is a single-field wrapper over `[]ItemVersion`. Fine if you want room to extend without breaking callers, but it's permanent public API either way, so worth making it a deliberate choice rather than a leftover.

multiget_test.go
```diff
@@ -0,0 +1,342 @@
+/*
+ * SPDX-FileCopyrightText: © 2017-2025 Dgraph Labs, Inc.
```
matthewmcneely: Header should be `© 2017-2026 Istari Digital, Inc.` — that's what the other 98 Go files in the repo use.

multiget_test.go
```diff
@@ -0,0 +1,342 @@
...
+// TestMultiGetReadSet is the regression test for the read-write conflict
+// tracking contract. Without NoReadTracking on the iterator, MultiGet would
+// record every walked key between successive Seeks as a read, causing
+// spurious conflicts in dgraph's HNSW-style fan-out.
+//
+// What we assert:
+//   - Only the requested keys end up in the txn's read set (len(txn.reads)
+//     equals the unique-key count, not the walked-key count).
+//   - The read set contains exactly the requested keys, with the requested
+//     duplicates collapsed.
```
matthewmcneely: `requested` has no duplicates, so nothing here exercises the collapsing claim. Duplicate keys do append twice to `txn.reads`; the assertion below only passes because it hashes into a map first. Either add a duplicate to `requested` or drop this line from the comment.

multiget_test.go
```diff
@@ -0,0 +1,342 @@
...
+		"read set should contain exactly the requested keys; got %d entries", len(readSet))
+	require.True(t, readSet[z.MemHash(mk(0))], "mk(0) should be in read set")
+	require.True(t, readSet[z.MemHash(mk(49))], "mk(49) should be in read set")
+}
+
+// loadFrontierDB builds a managed DB shaped like a dgraph predicate: dense
+// keys, a few MVCC versions each, flushed to disk.
+func loadFrontierDB(b *testing.B, nKeys, versions int) (*DB, [][]byte) {
+	b.Helper()
+	dir, err := os.MkdirTemp(".", "badger-mget")
```
matthewmcneely: `b.TempDir()` instead? `os.MkdirTemp(".", ...)` puts the dirs in the package directory, and they survive if the process dies before `b.Cleanup` runs.

matthewmcneely (changes requested):

shaunpatterson:
Thanks for the thorough review — all 7 points addressed in `29015a6` (pushed to `sp/badger_multiget`). Per-point:

1. **`NoReadTracking` exported → unexported.** Renamed to `noReadTracking` everywhere (struct field + the two callsites in `Item()` and `Seek()`). Doc comment now spells out the silent-SSI-bypass hazard. Matches the `prefixIsKey` precedent.

2. **`Txn.Get` parity for empty/banned keys.** Added explicit `ErrEmptyKey` / `ErrBannedKey` checks at the top of `MultiGetWithOptions`, so an empty or banned key in the request slice fails the whole call instead of silently looking absent. `TestMultiGetRejectsBadKeys` covers both. Doc comment updated to document the behavior.

3. **Deep-chain cost + `MaxVersionsPerKey`.** Added `IteratorOptions.MaxVersionsPerKey` (zero = unlimited, matching today) and a new `MultiGetOptions` / `MultiGetWithOptions` entry point with `WithMaxVersionsPerKey(n)` so callers can bound the per-key version walk. Doc comment calls out the dgraph `ReadPostingList` shape and notes the versions are returned newest-first so the cap trims the oldest tail.

   New `BenchmarkMultiGetDeepChainVsKeyIterator` (200 versions/key, 4 KiB values above `ValueThreshold=64` so they land in the vlog, K=64) on M4 Max:
   ```
   KeyIterator-16     286    8184789 ns/op   52.7 MB/op   26564 allocs/op
   MultiGet-16        249    9593842 ns/op   54.7 MB/op   26514 allocs/op   ← uncapped, slower (as you predicted)
   MultiGetCap1-16  51040     47982 ns/op    286 KB/op     528 allocs/op   ← ~200x speedup with cap=1
   ```
   So uncapped `MultiGet` is genuinely slower than per-key `KeyIterator` on deep chains — confirming your concern. The cap turns it into the right tool for the dgraph HNSW shape (read newest only). For the shallow-frontier workload the PR originally benchmarked (3 versions/key, 64-byte values) the wins still hold.

   Re: the `KeyIterator` baseline in the original benchmark calling `ValueCopy` — yes, that's a known approximation; I kept it as-is since it's the only thing measurable against `MultiGet`'s own copy. The deep-chain benchmark uses the same baseline so the comparison is apples-to-apples within each pair.

4. **`KeyResult` deliberate wrapper.** Doc comment now says it's a wrapper so future per-key fields (the requested key, a per-key error, found/not-found) can be added without breaking the result slice shape.

5. **Copyright bumped** to `© 2017-2026`.

6. **Duplicate-key claim in `TestMultiGetReadSet`** — added `mk(0)` a second time to `requested` (`{mk(0), mk(49), mk(0)}`), assert `len(got) == 3`, and assert `mk(0)`'s fingerprint appears exactly once in `txn.reads`. Doc comment updated.

7. **`b.TempDir()`** in `loadFrontierDB`; dropped the `os` import and the manual `removeDir` cleanup.

Full `./... -short` suite passes (258s) on the patched branch. Ready for re-review when you have time.

shaunpatterson:
Another round of fixes after a thermo-nuclear code-quality review of `29015a6`. Pushed as `86e6669`. Per-finding:

**Code-judo simplifications (deletions, not rearrangements):**

- **Deleted `MultiGetOptions` / `DefaultMultiGetOptions` / `WithMaxVersionsPerKey` / `MultiGetWithOptions`.** The whole options struct existed to plumb `MaxVersionsPerKey` through one method. Collapsed to two methods: `MultiGet(keys)` for unbounded, `MultiGetN(keys, maxVersionsPerKey int)` for capped. No functional-options pattern, no wrapper struct, no default-constructor boilerplate. The cap is just a parameter.

- **Moved `noReadTracking` off `IteratorOptions`.** Unexporting it on a public struct of genuine user-facing knobs (`Prefix`, `Reverse`, `AllVersions`, etc.) was a no-op for the API surface concern that motivated the unexport. Moved the field onto `Iterator` itself, next to the `Item()` and `Seek()` methods that read it. `MultiGet` sets it directly on the constructed iterator. Now there's no way to set it from outside the package.

- **Added `Key []byte` to `KeyResult`** so the wrapper earns its keep — callers no longer need a parallel array to associate each result with its requested key. Dropped the "deliberate wrapper for future fields" doc; the field is load-bearing now. Test asserts the echo.

- **Folded the empty/banned pre-flight loop into the main sorted walk.** Two-pass scan was only there to give "first invalid key in input order" precedence. Switched to sorted-order precedence (one branch in the loop instead of a separate O(n) pass), which is what callers actually want — deterministic in the key set, not in input order. Doc updated. New test case in `TestMultiGetRejectsBadKeys` exercises banned-before-good.

- **Collapsed `loadFrontierDB` + `loadDeepChainDB` into one `loadBenchDB`** helper with a `tweak func(*Options)` callback. -30 lines.

- **Dropped cross-package doc references** (`BitCompletePosting` etc. are dgraph's `UserMeta` bits, not badger's); described the use case generically.

- **Dropped the `mopts` indirection**: `maxVersions` is the local var `MultiGetN` passes straight through to `opt.MaxVersionsPerKey`. No round-trip through a struct field.

**Net:** iterator.go 989 → 970 lines, multiget_test.go 527 → 497. **-49 lines** while gaining `KeyResult.Key`, `MultiGetN`, sorted-order failure semantics, `KeyResult.Key` assertions, and the banned-before-good test case.

**Benchmarks still hold:** `BenchmarkMultiGetDeepChainVsKeyIterator` on M4 Max with 200 versions/key, 4 KiB values:
```
KeyIterator-16     21.0 ms/op   90.5 MB/op   99710 allocs/op
MultiGet-16        21.6 ms/op   94.3 MB/op  107972 allocs/op   ← uncapped, as before
MultiGetCap1-16   234    µs/op   818 KB/op     791 allocs/op   ← ~92× speedup with cap=1
```

**Tests:** full `./... -short` suite passes (257s). All 6 `TestMultiGet*` tests green, including the new banned-before-good assertion and `KeyResult.Key` echo assertion.

shaunpatterson:
Closing this. The consumer this was built for — dgraph-io/dgraph#9732, batching HNSW neighbor-vector reads — is also being closed, because measurement showed the batching is a regression rather than a win. The cause is in this API's design, so it applies here too.

## `MultiGet` forfeits per-key table pruning

`Txn.MultiGet` builds its shared iterator with no `Prefix`:

```go
opt := DefaultIteratorOptions
opt.AllVersions = true
opt.PrefetchValues = false
opt.MaxVersionsPerKey = maxVersions
it := txn.NewIterator(opt)
```

`NewKeyIterator` sets `prefixIsKey`, which lets `pickTables` reject tables by bloom filter (`t.DoesNotHave(y.Hash(opt.Prefix))`) and bound the table range. Without it the merge iterator spans every table at every level and each `Seek` descends into all of them. Amortizing iterator construction across the batch does not pay for that.

Measured on a store sized like a real embedding corpus — 200k keys x 1536 B values (384-d float32), snappy, 256 MB block cache, medians of 6 samples:

| frontier | `NewKeyIterator` loop | shared txn, per-key `NewKeyIterator` | shared prefix-less iterator | prefix-scoped shared iterator | `Txn.MultiGet` |
|---|---|---|---|---|---|
| scattered K=8 | 35.5 us | **+4%** | +67% | +76% | **+71%** |
| scattered K=16 | 77.4 us | **-11%** | +101% | +63% | **+53%** |
| scattered K=64 | 291 us | **+1%** | +56% | +53% | **+53%** |
| scattered K=150 | 686 us | **-0%** | +74% | +54% | **+56%** |
| contiguous K=64 | 198 us | **-19%** | +88% | +54% | **+50%** |

The second column is the key result: sharing the transaction while keeping per-key bloom-pruned iterators is **free** (±0%). So `NewKeyIterator` setup — the cost this PR set out to amortize — is not a meaningful part of the total. Scoping the shared iterator to a common key prefix recovers range pruning but not per-key bloom rejection, so it stays +50...+76%.

Sorting the keys, which `multiGet` already does, doesn't rescue it: the contiguous rows regress just as much as the scattered ones.

## Why `BenchmarkMultiGetVsKeyIterator` reported a win

That benchmark uses 64-byte values with `Flatten(2)` and draws a **contiguous** frontier (`keys[base+i]`), so the whole working set is a handful of resident blocks — the regime where there is nothing to prune and amortization is the only variable. Sweeping store size at the dgraph layer shows the sign flip clearly, and it tracks total bytes against the block cache rather than key count:

| store | bytes | batched vs per-key |
|---|---|---|
| n=500, dim=384 | ~0.8 MB | **-13%** |
| n=5 000, dim=384 | ~7.7 MB | **-13%** |
| n=50 000, dim=384 | ~77 MB | **-1%** |
| n=200 000, dim=64 | ~51 MB | **-14%** |
| n=200 000, dim=384 | ~307 MB | **+52%** |
| n=200 000, dim=1536 | ~1.2 GB | **+72%** |

Note `n=200 000/dim=64` (51 MB) wins while `n=50 000/dim=384` (77 MB) is already flat — it's resident bytes, not key count. Batched point reads help only while the working set stays inside the block cache, which is the opposite of when a batched read API is worth having.

## Secondary concern, if this is ever revisited

`multiGet` materializes every version with `item.ValueCopy(nil)`, a heap allocation per MVCC version. dgraph's `ReadPostingList` avoids that by using the in-place `item.Value(func(val []byte) error)` callback. For vlog-sized values that copy is a real per-version regression on top of the pruning loss.

For the batched-point-read use case, the promising direction isn't a shared iterator — it's keeping per-key bloom-pruned lookups and letting the caller issue them concurrently. The reads are independent; funnelling them through one iterator is what gives up the pruning that makes point reads fast.

Thanks for the review time on this one.

PR #2298, opened 2026-06-11
Add Go fuzz tests for OSS-Fuzz integration

canolgun:
Adds Go native fuzz targets covering critical code paths. Part of OSS-Fuzz integration (google/oss-fuzz#15661). Verified: go test -fuzz=. -fuzztime=30s


matthewmcneely:
This PR is against the `master` branch which has not been the release branch for some time.

PR #2330, opened 2026-08-03
fix(compaction): restore size-ratio base level selection, clamped to the first non-empty level

### LINKED ISSUE #2327: v4.9.4 regression: base level collapses to L1 on large bulk imports, ~3x slower than v4.9.2
## Describe the bug

(Human flagged, AI Prepared, human reviewed)

Upgrading from **v4.9.2 → v4.9.4** makes a large bulk import **~3x slower** on the
exact same workload, hardware, and options (only the badger version changed).

Analysis of badger's own compaction logs points at the `levelsController.levelTargets()`
rewrite that shipped in **v4.9.4 (#2278, "fix an issue where the compactor causes
previously deleted data to reappear")**. In v4.9.2 the base level is chosen from a
size-ratio calculation, which keeps the base level deep (L3–L5) for a large, mostly
bottom-heavy tree. In v4.9.4 the base level is chosen by scanning up from L1 and stopping
at the first non-empty level, so a single stray table at a shallow intermediate level
collapses the base level toward L1. Once the base level is shallow, L0 flushes land near
the top of the tree and the data has to cascade down level by level
(`1->2->3->4->5->6`) instead of being compacted directly toward the bottom
(`0->5`/`0->6`, then `5->6`). That extra write-amplification is what we believe accounts
for the slowdown.

We are reporting this as a probable regression and are happy to run further experiments
(see "How to confirm" below).

## To Reproduce

We bulk-load a large batch of key/value pairs into a freshly-created, empty DB opened with
`badger.Open(badger.DefaultOptions(dir))` — **no options overridden** (Snappy, MemTableSize
64 MiB, BaseLevelSize 10 MiB, LevelSizeMultiplier 10, MaxLevels 7, NumCompactors 4).

Write pattern (technically):

- Entries are buffered and flushed in batches of **5000**.
- Each batch is a single `db.Update(func(txn *badger.Txn) error { ... })` transaction that
  calls `txn.SetEntry(badger.NewEntry(key, value))` once per entry in the batch.
- Keys are short byte slices built from **several distinct top-level prefixes**, so the
  keyspace is spread across many prefixes rather than inserted in a single sorted range.
- Values are small byte slices. No TTL / `ExpiresAt` is set.
- The whole load runs against an initially empty DB and inserts on the order of **300M+
  entries** (~8–10 GB on disk).

Minimal standalone reproducer (no external data needed) — attached as `main.go`, with
captured logs `repro_v4.9.2.log` / `repro_v4.9.4.log`. It inserts ~302M entries the same
way: batches of 5000 committed via `db.Update` + `txn.SetEntry`, keys spread across 2
top-level prefixes.

```
go run . ./data 500000
```

Steps:

1. Run the reproducer against **v4.9.2** on an empty dir → base level stays deep (L3–L5),
   no L0 stalls.
2. Run the same reproducer against **v4.9.4** on a fresh empty dir → base level collapses
   to L1 partway through and stays there, L0 stalling appears, sustained cascade
   compactions.
3. Compare badger's compaction log lines (`LOG Compact <from>-><to>`) and the
   `L0 was stalled for` lines between the two runs.

**Production numbers** (identical input, identical hardware, only the badger version
changed):

|                                             | v4.9.2                 | v4.9.4                    |
| ------------------------------------------- | ---------------------- | ------------------------- |
| Load wall time (application-measured)       | 2h36m42s               | 7h45m16s — **2.97x**      |
| `1->2` + `2->3` + `3->4` cascade compactions | 9                      | 681                       |
| `0->1` direct compactions                   | 0                      | 23                        |
| Total L0-stall time                         | ~59.7 min              | ~82.9 min (1.4x)          |
| Dominant pattern                            | `5->6` direct, occasional `0->5`/`0->3` | `5->6` direct **plus** sustained `1->2->3->4->5` cascade |

**Reproducer numbers** (~302M entries, run is only ~5 min so the wall-clock gap is small,
but the mechanism signature is the same):

|                              | v4.9.2  | v4.9.4            |
| ---------------------------- | ------- | ----------------- |
| Lifetime L0 stall            | 0s      | 1m57.5s           |
| Base level at end            | L3/L4   | **L1**            |
| `0->1` direct compactions    | none    | present           |

## Expected behavior

Same import performance on v4.9.4 as on v4.9.2. Specifically, on a large, bottom-heavy
tree the base level should stay deep so that L0 data is compacted toward the bottom
directly, instead of collapsing to L1 and forcing a full level-by-level cascade with the
extra write-amplification (and L0 stalling) that comes with it.

## Screenshots

N/A. Compaction logs from both runs are attached (`repro_v4.9.2.log`,
`repro_v4.9.4.log`, plus the production logs).

## Environment

- OS: Linux (compiled inside a Docker container)
- Language: Go
- Version: **v4.9.4** (regression vs **v4.9.2**)

## Additional context

**Suspected cause — the `levelTargets()` rewrite in v4.9.4 (#2278).**
v4.9.2 sets `baseLevel` in the size-ratio loop (`ltarget <= BaseLevelSize`), which picks a
deep base level for a growing/bottom-heavy tree. v4.9.4 replaces that with
`baseLevel = 1` followed by a scan up from L1 that stops at the first level with any
non-zero size — so a single stray table at a shallow intermediate level immediately
collapses the base level toward L1. The compaction-direction distribution in our logs
(near-zero `1->2`/`2->3`/`3->4` on v4.9.2 vs hundreds on v4.9.4, and `0->1` appearing only
on v4.9.4) is the direct symptom of that collapse.

**Not the L0 backpressure change (#2308).** v4.9.3 replaced the busy-sleep L0 backpressure
with cond signalling. We don't think that is the primary cause: L0-stall time only grows
~1.4x (59.7 → 82.9 min) while total compaction work grows ~4x. If the backpressure change
were the driver we'd expect stall time to scale with the slowdown, not compaction volume.

**Honest caveats / what we have NOT confirmed:**

- The 2.97x (production) and the compaction/stall figures are exact, taken from badger's
  own logs. The mechanism explanation is our reading of the v4.9.2→v4.9.4 diff, not
  something confirmed by a maintainer.
- We attribute the `levelTargets()` rewrite to #2278 based on the v4.9.4 changelog and by
  elimination (v4.9.3's only change is #2308).
- We have **not** yet tested **v4.9.3** in isolation.
- The minimal reproducer demonstrates the qualitative mechanism (L0 stall going from zero
  to substantial, base level collapsing from deep to L1) but the run is short (~5 min), so
  it does not yet reproduce the full 3x magnitude — that only appears at production scale
  over a multi-hour run.

**Workaround:** pinned back to v4.9.2.

[repro_v4.9.4.log](https://github.com/user-attachments/files/30472439/repro_v4.9.4.log)
[repro_v4.9.2.log](https://github.com/user-attachments/files/30472440/repro_v4.9.2.log)
[main.go.txt](https://github.com/user-attachments/files/30472449/main.go.txt)

shiva-istari:
Fixes #2327

## Problem

v4.9.4 introduced a severe ingestion regression: on large `db.Update`-based imports
(300M+ entries, default options), the base level collapses to L1 partway through the
load and never recovers. Every byte must then cascade through every intermediate level
(`1→2→3→4→5→6`) instead of being compacted directly toward the bottom (`0→L4/L5`,
`5→6`). The reporter measured **2.97x slower** wall time in production (2h36m → 7h45m),
with cascade compactions going from 9 to 681 and sustained L0 write stalls.

## Background: how the two previous versions chose Lbase

`levelTargets()` decides which level L0 gets compacted into (Lbase). The two released
behaviors each got one thing right and one thing wrong:

**v4.9.2 — fast, but unsafe.** Lbase came from a pure size-ratio computation
(dynamic level sizing, as in RocksDB): the deepest level whose target size is within
`BaseLevelSize`, extended further down through empty levels. For a bottom-heavy tree
this keeps Lbase deep (L3–L5) and write amplification low. However, the selection never
looked at the levels **above** the chosen base. If a stray table held data at a
shallower level (possible after the tree shrinks, or historical states), L0's newest
entries — including delete tombstones — were compacted **below** older versions of the
same keys. Reads stop at the shallowest version, so previously deleted data reappeared
and fresh updates read stale. That corruption is what #2278 fixed.

**v4.9.4 (#2278) — safe, but collapses.** The size-ratio selection was replaced with an
emptiness scan: start at L1 and walk down through *empty* levels, stopping at the first
non-empty one; Lbase is the empty level just above it. This makes the corruption
structurally impossible (nothing can sit above Lbase), but the rule is self-destructive
under write load: dumping L0 onto that empty level makes it non-empty, so the next
selection stops one level higher. Each dump burns a level — L4 → L3 → L2 → L1 — and
once Lbase reaches L1, `0→1` compactions keep L1 permanently non-empty, pinning the
base there forever. The result is the full per-level cascade, ~3x write amplification,
and L0 stalls.

## The fix

Combine the two, mirroring RocksDB's `CalculateBaseBytes()`:

1. **Restore the v4.9.2 size-ratio selection** (deepest level with target ≤
   `BaseLevelSize`, plus the existing walk down through empty levels and the
   step-onto-a-tiny-next-level tweak). This is the *performance floor*: occupying the
   base level does not move the next round's choice, so there is no ratchet.
2. **Add a safety clamp**: if any level *above* the chosen base holds data, Lbase
   becomes the **shallowest non-empty level itself** — never a level below existing
   data, and never an empty level above existing data. Compacting L0 *into* the
   occupied level merges versions in the correct order, so #2278's invariant is fully
   preserved (its runtime guard in `runCompactDef` is untouched and still enforces
   this at compaction time).

The clamp also makes bad states self-healing: an L0 dump merging into a stray level
pushes it past its size target, normal compaction drains it, and Lbase returns to its
natural depth — typically within one compaction cycle. Trees already collapsed by
v4.9.4 recover the same way under write load (their L1–L3 remnants sit far above the
10MB targets, so they drain at high priority).

## Why this does not reintroduce the #2278 corruption

The corruption required Lbase to be *deeper* than a level holding data. After the
clamp, `baseLevel ≤ first non-empty level` holds on every path, so every level above
Lbase is empty by construction. All regression tests added by #2278 pass, including
the end-to-end `TestDeletedKeyReappears` and the non-empty-level-above-base runtime
guard test. One #2278 unit expectation is deliberately updated:
`TestLevelTargets/level size exactly at BaseLevelSize boundary` now expects Lbase = L5
(the first non-empty level, whose size-ratio target is exactly `BaseLevelSize`) instead
of the empty L4 above it — compacting into L5 is safe (nothing above it) and choosing
empty L4 is precisely the ratchet mechanism this PR removes. The test comment explains
the reasoning.

## Results

Reporter's standalone reproducer from #2327 (302M entries via `db.Update`, batches of
5000, default options), same machine, three-way comparison:

|                              | v4.9.2      | v4.9.4              | this PR       |
| ---------------------------- | ----------- | ------------------- | ------------- |
| Wall time                    | 4m08s       | 4m26s               | **3m56–4m00s** (4 runs) |
| Lifetime L0 stall            | 0s          | **57.7s** (11 events) | **0s**      |
| Base level over the run      | L4          | collapses to **L1** at ~170M entries, never recovers | **L3/L4 throughout** |
| L1–L2 at end                 | empty       | 98MB + 61MB stuck   | empty         |
| Compaction pattern           | 9× `5→6`    | `5→6` + `0→1` + cascades | 2–5× `5→6` only |

(The wall-clock gap at this scale is small; at the reporter's production scale the
collapse compounds to ~3x. The structural signals — stalls, base level, cascade
compactions — are the meaningful metrics, and they match v4.9.2 exactly.)

Full test suite passes. New `TestLevelTargetsStrayLevels` covers the three selection
scenarios this PR is about: clamping to a stray level, a fully collapsed tree, and
recovery once shallow levels drain.

## Note for users downgrading as a workaround

Pinning back to v4.9.2 on a data directory that v4.9.4 left in the collapsed shape
(data stranded in L1–L3) re-exposes the #2278 corruption: v4.9.2 will compact L0 below
those levels. The safe options are staying on v4.9.4 or upgrading to a release with
this fix.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci-badger-lsm-bench.yml
+++ b/.github/workflows/ci-badger-lsm-bench.yml
@@ -1,0 +1,122 @@
+name: ci-badger-lsm-bench
+
+# Bulk-load LSM performance benchmark (issue #2327 regression gate).
+#
+# Builds the same harness (integration/lsmbench) against the PR's base and the
+# PR's head, runs both on this runner, and fails if:
+#   - the head run shows the structural #2327 signature (base level collapsed
+#     to L1/L2 on a >2GB tree, tables stranded in L1/L2 at any checkpoint,
+#     or lifetime L0 stall > 30s), or
+#   - the head wall clock is >25% slower than the base run on the same
+#     hardware (self-baselining: no hardcoded absolute time, so the gate is
+#     meaningful regardless of runner speed).
+#
+# Runs independently of the unit tests and badger-bank; ~2x a single bulk
+# load of ~302M entries (~10GB on disk per run, cleaned between runs).
+
+on:
+  workflow_dispatch: # allows manual trigger from GitHub
+  pull_request:
+    branches:
+      - main
+      - release/v*
+
+permissions:
+  contents: read
+
+env:
+  LSMBENCH_ROWS: 500000 # ~302M entries, ~10GB LSM
+
+jobs:
+  # Detect whether any non-doc files changed, mirroring the other CI
+  # workflows: the job always reports a status, and short-circuits to success
+  # for docs-only PRs.
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      code: ${{ steps.filter.outputs.code }}
+    steps:
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          predicate-quantifier: every
+          filters: |
+            code:
+              - '!**/*.md'
+              - '!docs/**'
+              - '!images/**'
+
+  lsm-bench:
+    needs: changes
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    steps:
+      - if: needs.changes.outputs.code == 'true'
+        uses: actions/checkout@v5
+      - if: needs.changes.outputs.code == 'true'
+        name: Checkout baseline (PR base)
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ github.event.pull_request.base.sha || 'main' }}
+          path: base-checkout
+      - if: needs.changes.outputs.code == 'true'
+        name: Setup Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+      - if: needs.changes.outputs.code == 'true'
+        name: Build harnesses (head + baseline)
+        run: |
+          #!/bin/bash
+          set -euo pipefail
+          # Head: harness compiled against this checkout.
+          go build -o /tmp/lsmbench-head ./integration/lsmbench/
+          # Baseline: same harness source compiled against the base checkout,
+          # via a scratch module so the import resolves to the old tree.
+          mkdir /tmp/base-mod && cp integration/lsmbench/main.go /tmp/base-mod/
+          cd /tmp/base-mod
+          go mod init lsmbench-base
+          go mod edit -require=github.com/dgraph-io/badger/v4@v4.0.0 \
+            -replace=github.com/dgraph-io/badger/v4="$GITHUB_WORKSPACE/base-checkout"
+          go mod tidy
+          go build -o /tmp/lsmbench-base .
+      - if: needs.changes.outputs.code == 'true'
+        name: Run baseline benchmark
+        run: |
+          #!/bin/bash
+          set -euo pipefail
+          # No gates on the baseline run: it only provides the wall-clock
+          # reference, and the base commit may itself carry a known regression.
+          /tmp/lsmbench-base -dir /tmp/bench-base -rows "$LSMBENCH_ROWS" | tee /tmp/base.out
+          rm -rf /tmp/bench-base
+      - if: needs.changes.outputs.code == 'true'
+        name: Run head benchmark (structural gates enforced)
+        run: |
+          #!/bin/bash
+          set -euo pipefail
+          /tmp/lsmbench-head -dir /tmp/bench-head -rows "$LSMBENCH_ROWS" \
+            -min-base 3 -max-stall 30s -max-shallow-occupied 0 | tee /tmp/head.out
+          rm -rf /tmp/bench-head
+      - if: needs.changes.outputs.code == 'true'
+        name: Compare wall clock (head vs baseline)
+        run: |
+          #!/bin/bash
+          set -euo pipefail
+          base_ms=$(grep LSMBENCH_RESULT /tmp/base.out | grep -o 'wall_ms=[0-9]*' | cut -d= -f2)
+          head_ms=$(grep LSMBENCH_RESULT /tmp/head.out | grep -o 'wall_ms=[0-9]*' | cut -d= -f2)
+          limit_ms=$(( base_ms * 125 / 100 ))
+          {
+            echo "### LSM bulk-load benchmark (${LSMBENCH_ROWS} rows)"
+            echo ""
+            echo "| run | result |"
+            echo "|---|---|"
+            echo "| baseline | $(grep LSMBENCH_RESULT /tmp/base.out) |"
+            echo "| head | $(grep LSMBENCH_RESULT /tmp/head.out) |"
+            echo ""
+            echo "Wall clock: head ${head_ms}ms vs baseline ${base_ms}ms (limit: ${limit_ms}ms = 125% of baseline)"
+          } >> "$GITHUB_STEP_SUMMARY"
+          echo "baseline=${base_ms}ms head=${head_ms}ms limit=${limit_ms}ms"
+          if [ "$head_ms" -gt "$limit_ms" ]; then
+            echo "::error::head wall clock ${head_ms}ms is more than 25% slower than baseline ${base_ms}ms"
+            exit 1
+          fi
--- a/integration/lsmbench/main.go
+++ b/integration/lsmbench/main.go
@@ -1,0 +1,317 @@
+/*
+ * SPDX-FileCopyrightText: © Hypermode Inc. <hello@hypermode.com>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// lsmbench is a bulk-load LSM benchmark used to detect compaction
+// performance regressions (issue #2327). It recreates the reproducer from
+// that issue: ~300M small entries written through db.Update in batches of
+// 5000 against default options, keys spread across several top-level
+// prefixes.
+//
+// It intentionally uses only badger's public API so the same harness can be
+// compiled against any badger version (the module's replace directive points
+// at the working tree; CI repoints it at a baseline checkout to compare the
+// two on identical hardware).
+//
+// Beyond wall time, it tracks the structural signature of the #2327
+// regression, which is hardware-independent:
+//
+//   - the base level must not collapse to L1/L2 once the tree is large
+//     (healthy: L3+; broken: pinned at L1 from ~170M entries onward), and
+//   - lifetime L0 write stalls must stay near zero (healthy: 0s; broken:
+//     ~1-2 minutes at this scale).
+//
+// Exit code is non-zero if any enabled gate fails. The final line of output
+// is machine-readable:
+//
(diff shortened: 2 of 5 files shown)


matthewmcneely:
@shiva-istari Can we add an integration test for this recreating the 4m wall clock / 300m test? Tag it so it doesn't get caught up in normal unit tests. Then add an independent CI workflow for it? The badger-bank takes over 20m so this shouldn't hold things up. Maybe hardcode a 4m wall clock threshold, if the test falls outside, fail the run. Might be a good foundation for a few other select benchmarks too that can be added in the future.

shiva-istari:
> @shiva-istari Can we add an integration test for this recreating the 4m wall clock / 300m test? Tag it so it doesn't get caught up in normal unit tests. Then add an independent CI workflow for it? The badger-bank takes over 20m so this shouldn't hold things up. Maybe hardcode a 4m wall clock threshold, if the test falls outside, fail the run. Might be a good foundation for a few other select benchmarks too that can be added in the future.


Added in the latest commits:

**`integration/lsmbench`** — a standalone benchmark harness recreating the #2327 reproducer (~302M entries via `db.Update`, default options), plus an integration test behind the **`integration` build tag** (verified: `make test` / plain `go test ./...` never picks it up). Run manually with:

```
go test -v -tags=integration -timeout 60m ./integration/lsmbench/
```

**`ci-badger-lsm-bench.yml`** — independent workflow, doesn't touch the unit-test or badger-bank jobs.

One deliberate change from the suggestion: **no hardcoded 4m wall clock.** Two reasons, both measured during validation:

1. 4m was my 12-core workstation's number — a 4-vCPU runner will be far off it in either direction.
2. Wall time barely discriminates at CI scale anyway: across validation runs, the *broken* (pre-fix) build was only **1.4–3.6% slower** than the fixed one. Any threshold tight enough to catch that fires on runner noise (same code, same machine varies ~7% run-to-run).

Instead the workflow is **self-baselining**: it compiles the same harness against the PR's merge-base and the PR's head (the harness uses only public API, so it builds against any badger version), runs both on the same runner, and fails on any of:

- **base level collapse** — base level below L3 on a >2GB tree (the #2327 signature; binary and hardware-independent),
- **shallow occupancy** — tables stranded in L1/L2 at any entry-indexed checkpoint,
- **L0 stalls** — lifetime stall > 30s,
- **wall clock** — head > 125% of base *on the same runner* (catastrophic-slowdown backstop; runner speed cancels out).

Validated locally at full scale, twice each:

| | fixed (this PR) | pre-fix main |
|---|---|---|
| verdict | ✅ all gates pass | ❌ `base level collapsed: min base on a >2GB tree was L2, want >= L3` |
| min base level | L3 | L1 / L2 (varies by run) |
| collapsed checkpoints | 0 | 7 of 15 |
| wall delta | — | +1.4% / +3.6% (would pass any sane time gate — hence the structural gates) |

Both runs' `LSMBENCH_RESULT` lines land in the workflow's job summary so the numbers are visible per-PR without log digging. The harness takes `-rows`, `-min-base`, `-max-stall`, `-max-shallow-occupied` flags, so adding further benchmarks on this foundation is mostly new workloads + gates.

matthewmcneely (approved):
argus run shows no regression in Dgraph

op          metric              baseline    candidate   delta%   thr%  noise%  verdict
workload    measured_s           709.062     466.819    -34.2      10     7.4  ok
cluster     rss_mb_total        8547.341    8678.421     +1.5      15     2.8  ok
cluster     jemalloc_mb_total    452.252     435.380     -3.7      15     5.4  ok
cluster     cpu_pct_total        409.931     376.056     -8.3      15     9.2  ok
cluster     goroutines_total     400.667     399.167     -0.4      15     0.1  ok
cluster     queries_per_s         38.330      49.806    +29.9      10     8.3  ok
bulkload    nquads_per_s       13785.887   16366.716    +18.7      10     4.5  ok
bulkload    batch_p99           2292.729    1837.399    -19.9      10     5.4  ok

PR #2331, opened 2026-08-08
fix: handle empty .mem / .vlog files during Open() instead of failing with z.NewFile (#2207)

### LINKED ISSUE #2207: z.NewFile error on badger.Open()
## Describe the bug

Sometimes a simple `badger.Open()` with default options returns error `z.NewFile`.

## To Reproduce

Keep writing data, closing, reopening, writing data, etc...

A second look gave me the way to reproduce it:
- Just leave the dir state containing `.mem` files.

In this case, `badger.Open()` always return `z.NewFile`.

## Expected behavior

No error. In fact, a second call to `badger.Open()` returns no error.

## Environment

- OS: macOS, Linux
- Language: Go
- Version: v4.7.0

## Additional context

This portion of code [here](https://github.com/hypermodeinc/badger/blob/main/memtable.go#L75-L78) seems to be the root cause. So `openMemTable` can return `mt, z.NewFile` as per [ref](https://github.com/hypermodeinc/badger/blob/main/memtable.go#L130-L132). But `mt` != nil in this case.

Other portions of the code make sense of it, like [here](https://github.com/hypermodeinc/badger/blob/main/table/table.go#L248).

adityeah8969:
**Description**

Fixes [#2207](https://github.com/dgraph-io/badger/issues/2207)

When badger.Open() encounters an empty (0-byte) .mem or .vlog file in the data directory, it fails instead of opening the database. Both can be left behind by a crash between file creation and the first write.

Old Buggy Behaviour

- .mem, read-write: z.OpenMmapFileUsing truncates the 0-byte file to MemTableSize and returns the z.NewFile sentinel, which openMemTables treats as a fatal error → Open() aborts.
- .mem, read-only: the same truncate runs against an O_RDONLY fd and fails with "invalid argument" before the sentinel is returned → Open() aborts.
- .vlog, read-write: same as .mem — the file is truncated to 2 * ValueLogFileSize and the z.NewFile sentinel is treated as fatal → Open() aborts.
- .vlog, read-only: truncate on an O_RDONLY fd fails with "invalid argument" → Open() aborts.

Fix (mem): stat each .mem file before opening and skip any 0-byte file. Stat-and-skip never opens the file, so no truncation is attempted — it works in both read-write and read-only modes and leaves the empty file untouched.

Fix (vlog): same stat-and-skip for .vlog files, removing the skipped fid from vlog.filesMap so Close() doesn't dereference its nil MmapFile, and guarding the max-fid truncation so a skipped max-fid artifact isn't touched.

**Checklist**

---
<a href="https://app.blacksmith.sh/dgraph-io/codesmith/badger/pr/2331"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788779308&installation_model_id=8575&pr_number=2331&repository=dgraph-io%2Fbadger&return_to=https%3A%2F%2Fgithub.com%2Fdgraph-io%2Fbadger%2Fpull%2F2331&signature=5610b4b18b10b4ced9bba90751e23800b7ac701cd9abb8eaaaed67ebf6a2d382"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database recovery by safely skipping empty memory and value-log files left behind after crashes.
  * Preserved existing data and write functionality when opening databases in read-only or read-write modes.
  * Prevented failures when the latest value-log file is empty or unavailable during startup.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -2740,8 +2740,9 @@ func TestCloseDBWhileReading(t *testing.T) {
 
 // TestOpenWithEmptyMemFile verifies that badger.Open handles empty .mem files
 // gracefully. An empty .mem file can be left behind after a crash that occurs
-// between file creation and the first write. Before the fix for #2207,
-// openMemTables would treat z.NewFile as a fatal error and refuse to open the DB.
+// between file creation and the first write. The fix for #2207 stats each .mem
+// file during discovery and skips 0-byte files, which works for both read-write
+// and read-only modes without mutating the file.
 func TestOpenWithEmptyMemFile(t *testing.T) {
 	dir := t.TempDir()
 	opt := getTestOptions(dir)
@@ -2787,6 +2788,13 @@ func TestOpenWithEmptyMemFile(t *testing.T) {
 	// Step 3: Re-open the DB. This must succeed despite the empty .mem file.
 	db2, err := Open(opt)
 	require.NoError(t, err, "Open should succeed with an empty .mem file")
+
+	// The empty file should still exist — the stat-and-skip approach is
+	// intentionally non-mutating so it works in read-only mode as well.
+	require.FileExists(t, emptyFile)
+	fi, err := os.Stat(emptyFile)
+	require.NoError(t, err)
+	require.Equal(t, int64(0), fi.Size(), "empty .mem file should be left untouched")
 
 	// Step 4: Verify existing data is intact.
 	err = db2.View(func(txn *Txn) error {
--- a/db_test.go
+++ b/db_test.go
@@ -2823,4 +2823,155 @@ func TestOpenWithEmptyMemFile(t *testing.T) {
 	require.NoError(t, err)
 
 	require.NoError(t, db2.Close())
+
+	// Step 6: Read-only Open must also succeed with the empty .mem file
+	// present, since the stat-and-skip approach avoids truncation entirely.
+	roOpt := opt
+	roOpt.ReadOnly = true
+	db3, err := Open(roOpt)
+	require.NoError(t, err, "read-only Open should succeed with an empty .mem file")
+	require.NoError(t, db3.Close())
+}
+
+// TestOpenWithEmptyVlogFile verifies that badger.Open handles empty .vlog files
+// gracefully. Like an empty .mem file, a 0-byte .vlog file can be left behind
+// after a crash between file creation and the first write. The fix stats each
+// .vlog file during open and skips 0-byte files, which works for both read-write
+// and read-only modes without mutating the file.
+func TestOpenWithEmptyVlogFile(t *testing.T) {
+	dir := t.TempDir()
+	opt := getTestOptions(dir)
+
+	// Step 1: Create a DB with some data.
+	db, err := Open(opt)
+	require.NoError(t, err)
+	for i := 0; i < 100; i++ {
+		err := db.Update(func(txn *Txn) error {
+			return txn.Set([]byte(fmt.Sprintf("key:%05d", i)),
+				[]byte(fmt.Sprintf("value:%05d", i)))
+		})
+		require.NoError(t, err)
+	}
+	require.NoError(t, db.Close())
+
+	// Step 2: Find the highest file ID and plant an empty .vlog file with a
+	// higher ID. Value log files use %06d.vlog (6-digit zero padding).
+	entries, err := os.ReadDir(dir)
+	require.NoError(t, err)
+	maxFid := 0
+	for _, e := range entries {
+		name := e.Name()
+		for _, ext := range []string{".vlog", ".sst", memFileExt} {
+			if strings.HasSuffix(name, ext) {
+				fidStr := name[:len(name)-len(ext)]
+				if fid, err := strconv.ParseInt(fidStr, 10, 64); err == nil {
+					if int(fid) > maxFid {
+						maxFid = int(fid)
+					}
+				}
+			}
+		}
+	}
+	emptyFid := maxFid + 1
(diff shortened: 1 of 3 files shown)


memtable.go
```diff
@@ -73,6 +73,12 @@ func (db *DB) openMemTables(opt Options) error {
 			flags = os.O_RDONLY
 		}
 		mt, err := db.openMemTable(fid, flags)
+		if err == z.NewFile {
+			// This memtable's WAL was empty (0 bytes), so it was
+			// bootstrapped. There's nothing to recover; discard it.
+			mt.DecrRef()
+			continue
+		}
```
matthewmcneely: This fixes the read-write path, but the same crash artifact still blocks `Open` with `ReadOnly: true`. The `z.NewFile` sentinel is only reachable when the fd is writable: for an existing 0-byte file, `z.OpenMmapFileUsing` calls `fd.Truncate` before it can return `NewFile`, and on an `O_RDONLY` descriptor that fails. I verified on this branch: a read-only open with a leftover empty `.mem` file still errors, now with `truncate ...: invalid argument`.

An alternative that covers both modes: stat the file in `openMemTables` and skip 0-byte files before calling `openMemTable` (skip rather than delete, so read-only stays non-mutating). That would also avoid some churn in the current shape, where the empty file gets truncated to `2*MemTableSize`, bootstrapped with a header, synced, and then immediately deleted via the skiplist `OnClose` callback.

Not a blocker (read-only opens were equally broken before this PR), but if you keep the current approach, please add a comment noting the read-only case is still unhandled.

db_test.go
```diff
@@ -2735,3 +2737,82 @@ func TestCloseDBWhileReading(t *testing.T) {
...
+	}
+	emptyFid := maxFid + 1
+	emptyFile := filepath.Join(dir, fmt.Sprintf("%05d%s", emptyFid, memFileExt))
+	f, err := os.Create(emptyFile)
+	require.NoError(t, err)
+	require.NoError(t, f.Close())
+
+	// Step 3: Re-open the DB. This must succeed despite the empty .mem file.
+	db2, err := Open(opt)
+	require.NoError(t, err, "Open should succeed with an empty .mem file")
```
matthewmcneely, line 2790: Can you add `require.NoFileExists(t, emptyFile)` after this Open? The fix relies on a side effect that would be easy to break silently: `mt.DecrRef()` drops the skiplist refcount to zero, which fires the `OnClose` callback installed in `openMemTable`, and that deletes the WAL file. I confirmed on this branch that the empty file is gone after re-open and a third open is clean. Pinning that here keeps a future refactor of the refcount/`OnClose` wiring from quietly leaving stale `.mem` files behind.

matthewmcneely (changes requested):

memtable.go
adityeah8969: Makes sense, I have opted for "stat and skip" now.

db_test.go
adityeah8969: This might not be relevant after the new approach taken [here](https://github.com/dgraph-io/badger/pull/2331#discussion_r3753417334).

matthewmcneely: Ah right

db_test.go
```diff
@@ -2735,3 +2737,90 @@ func TestCloseDBWhileReading(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Step 5: Write more data to ensure the DB is fully functional.
+	err = db2.Update(func(txn *Txn) error {
+		return txn.Set([]byte("new-key"), []byte("new-value"))
+	})
+	require.NoError(t, err)
+
+	require.NoError(t, db2.Close())
+}
```
matthewmcneely: Could you add a test to actually read-only mode?

```go
// Step 6: The same artifact must not block a read-only open.
roOpt := opt
roOpt.ReadOnly = true
db3, err := Open(roOpt)
require.NoError(t, err, "read-only Open should succeed with an empty .mem file")
require.NoError(t, db3.Close())

```

matthewmcneely (changes requested):
Thanks for riding this out @adityeah8969....

But also note the same bug exists in the value log path (value.go line 611) and #2207 notes "leaving the dir state" after crashes, so this can't close that issue until that's addressed. Should be the same logic and the mem files, right?

db_test.go
adityeah8969: Added

adityeah8969:
@matthewmcneely I have pushed a commit to solve empty vlog file issue as well, which you described [here](https://github.com/dgraph-io/badger/pull/2331#pullrequestreview-4918881800)
 
Change was the same, just needed an extra check to not truncate empty .vlog file later in the method "func (vlog *valueLog) open(db *DB)"

adityeah8969:
@matthewmcneely  All review comments have been addressed, please let me know if we could take this improvement to closure.

PR #2333, opened 2026-08-15
fix(table): return error instead of fatalpanic when opening a corrupt sst

### LINKED ISSUE #2201: fatalpanic in background goroutine due to re-panic in table.(*Table).initBiggestAndSmallest
# What version of Badger are you using?
v4.5.1

# Does this issue reproduce with the latest master?

We have tested this with the latest version. 

# What operating system and version are you using?
OS: Windows 10 Pro 

# What Go version are you using?
Go version : 1.22.12

# What did you do? (Steps to reproduce)
Have a BadgerDB v4 database with a corrupted SSTable. The corruption specifically seems to affect the table index reading process, triggering an initial panic during t.initIndex() or related FlatBuffer parsing.
Attempt to open this database using badger.Open(opts). The opts include a custom logger.
badger.Open() might return successfully without a synchronous error or panic.
Shortly after the database appears to open, a background goroutine initiated by BadgerDB (seemingly from newLevelsController for table loading/initialization) encounters an issue.

# What did you expect to see?
We expected that if BadgerDB encounters severe data corruption in a background task that leads to an internal panic (even a deliberate re-panic for debug purposes), it should:

Catch this panic within its own background goroutine.
Log a detailed error through the provided logger.
Transition the DB instance to an error state or signal the application in a manageable way (e.g., by making subsequent operations return errors, or by closing the DB and making Open fail on next attempt if not already).
The application should not crash due to an unhandled panic in a BadgerDB background goroutine.

The application crashes due to runtime.fatalpanic. The stack trace indicates that an initial panic (P1) during table index processing (e.g., in flatbuffers.GetUint32) was caught by a defer in table.(*Table).initBiggestAndSmallest. This defer then collected debug information and subsequently executed a nested defer which intentionally re-panicked (P2) with this debug information. This P2 was not caught further up in BadgerDB's background goroutine call stack, leading to fatalpanic.

### Key Stack Trace Snippet of the fatal panic:
```
runtime.fatalpanic (panic.go:1217) runtime
runtime.gopanic (panic.go:779) runtime
table.(*Table).initBiggestAndSmallest.func1.1 (table.go:352) github.com/dgraph-io/badger/v4/table // <<< P2 (re-panic with debug info) occurs here
runtime.deferreturn (panic.go:602) runtime                                                  // Part of defer handling for P1
table.(*Table).initBiggestAndSmallest.func1 (table.go:398) github.com/dgraph-io/badger/v4/table // Outer defer in initBiggestAndSmallest that caught P1
runtime.gopanic (panic.go:770) runtime                                                      // This is gopanic for P1
runtime.goPanicIndex (panic.go:114) runtime                                                 // P1 (e.g., index out of bounds)
flatbuffers.GetUint32 (encode.go:47) github.com/google/flatbuffers/go                      // P1 might originate here
flatbuffers.GetUOffsetT (encode.go:121) github.com/google/flatbuffers/go
fb.GetRootAsTableIndex (TableIndex.go:14) github.com/dgraph-io/badger/v4/fb
table.(*Table).readTableIndex (table.go:706) github.com/dgraph-io/badger/v4/table
table.(*Table).initIndex (table.go:463) github.com/dgraph-io/badger/v4/table
table.(*Table).initBiggestAndSmallest (table.go:402) github.com/dgraph-io/badger/v4/table  // Call leading to P1
table.OpenTable (table.go:308) github.com/dgraph-io/badger/v4/table
badger.newLevelsController.func1 (levels.go:151) github.com/dgraph-io/badger/v4            // This is likely running in a new goroutine
badger.newLevelsController.gowrap2 (levels.go:166) github.com/dgraph-io/badger/v4
runtime.goexit (asm_amd64.s:1695) runtime                                                   // Goroutine exit
 - asynchronous stack trace
badger.newLevelsController (levels.go:130) github.com/dgraph-io/badger/v4   
```

# (Optional) Relevant code analysis / Source of the re-panic:

// In (t *Table).initBiggestAndSmallest()
// ...
	defer func() { // D1 - Outer defer
		if r := recover(); r != nil { // Catches P1 from initIndex or other table ops
			var debugBuf bytes.Buffer
			defer func() { // D2 - Nested defer
				// THIS LINE INTENTIONALLY RE-PANICS (P2) WITH DEBUG INFO
				panic(fmt.Sprintf("%s\n== Recovered ==\n", debugBuf.String()))
			}()
			// Code to populate debugBuf with file ID, size, checksums, index info, etc.
			fmt.Fprintf(&debugBuf, "\n== Recovering from initIndex crash ==\n") // Part of the P2 message
            // ... more Fprintf calls ...
		}
	}()
// ...

This deliberate re-panic P2, if unhandled in the background goroutine where initBiggestAndSmallest is executed, causes the fatalpanic.

adityeah8969:
**Description**

Fixes [#2201](https://github.com/dgraph-io/badger/issues/2201) 

`db.Open()` crashes with a `runtime.fatalpanic` when an SSTable's footer is corrupted (e.g. zeroed), instead of returning an error.

Looking at the stack trace of issue description, it looks like following could be the reason for runtime termination,

An SST footer is laid out as `[index][indexLen][checksum][checksumLen]`, with the two lengths stored as 4-byte big-endian values. When the footer is zeroed, `checksumLen` and `indexLen` both read as `0`, so we read empty checksum and index bytes. The empty protobuf checksum unmarshals to `Sum = 0`, and the CRC32C of an empty index is also `0`, so the `0 == 0` verification passes. The code then parses the empty index as flatbuffers and reads an out-of-bounds offset into a slice, which panics. `initBiggestAndSmallest` catches that panic, collects debug info, then deliberately re-panics — and that re-panic escapes the table-loading goroutine in `newLevelsController` (which has no `recover()`), so the runtime terminates the process.

Changes:
- `table/table.go`: `initBiggestAndSmallest` returns an error (instead of re-panicking) after collecting debug info, and `initIndex` validates the footer's `checksumLen`/`indexLen`, rejecting `<= 0` (an empty checksum/index false-passes the `0 == 0` checksum check and then panics parsing empty flatbuffers) and `>= tableSize` (the length can't describe a region as large as the whole file — it would make the footer's read offset negative and slice out of bounds).
- `levels.go`: added a `recover()` in the table-loading goroutine as a safety net for any panic that still escapes the table package.

**Checklist**

### DIFF
--- a/db_corrupt_test.go
+++ b/db_corrupt_test.go
@@ -1,0 +1,107 @@
+/*
+ * SPDX-FileCopyrightText: © 2017-2025 Istari Digital, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package badger
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// writeDBWithSST creates a DB, writes enough keys to flush an on-disk SSTable,
+// closes it, and returns the dir, the options used, and the first .sst path.
+func writeDBWithSST(t *testing.T) (dir string, opts Options, sstPath string) {
+	t.Helper()
+
+	dir, err := os.MkdirTemp("", "badger-test")
+	require.NoError(t, err)
+	opts = getTestOptions(dir)
+
+	db, err := Open(opts)
+	require.NoError(t, err)
+	for i := 0; i < 5000; i++ {
+		err := db.Update(func(txn *Txn) error {
+			return txn.Set(
+				[]byte(fmt.Sprintf("key:%08d", i)),
+				[]byte(fmt.Sprintf("value:%08d", i)))
+		})
+		require.NoError(t, err)
+	}
+	require.NoError(t, db.Close())
+
+	entries, err := os.ReadDir(dir)
+	require.NoError(t, err)
+	for _, e := range entries {
+		if strings.HasSuffix(e.Name(), ".sst") {
+			return dir, opts, filepath.Join(dir, e.Name())
+		}
+	}
+	t.Fatal("expected at least one SST on disk")
+	return dir, opts, ""
+}
+
+// TestOpenWithCorruptedFooterReturnsError mirrors the issue #2201 repro: the
+// footer/index region is zeroed, which must be rejected by the footer
+// validation and surfaced as a graceful Open error (not a fatalpanic).
+func TestOpenWithCorruptedFooterReturnsError(t *testing.T) {
+	dir, opts, sstPath := writeDBWithSST(t)
+	defer removeDir(dir)
+
+	data, err := os.ReadFile(sstPath)
+	require.NoError(t, err)
+	corruptSize := 300
+	if len(data) < corruptSize {
+		corruptSize = len(data)
+	}
+	for i := len(data) - corruptSize; i < len(data); i++ {
+		data[i] = 0
+	}
+	require.NoError(t, os.WriteFile(sstPath, data, 0644))
+
+	_, err = Open(opts)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "Data corrupted")
+}
+
+// TestOpenWithCorruptedBlockDataSucceeds corrupts only block data (the middle
+// of the file), leaving the footer and index intact. Block data is read
+// lazily, so Open must succeed rather than crash or error.
+func TestOpenWithCorruptedBlockDataSucceeds(t *testing.T) {
+	dir, opts, sstPath := writeDBWithSST(t)
+	defer removeDir(dir)
+
+	data, err := os.ReadFile(sstPath)
+	require.NoError(t, err)
+	mid := len(data) / 2
+	for i := mid; i < mid+100 && i < len(data); i++ {
+		data[i] = 0xFF
+	}
+	require.NoError(t, os.WriteFile(sstPath, data, 0644))
+
+	db, err := Open(opts)
+	require.NoError(t, err)
+	require.NoError(t, db.Close())
+}
+
+// TestOpenWithTinySSTableReturnsError truncates an SST below the 4-byte footer.
+// initIndex then reads at a negative offset and panics, and the recover defer's
+// debug collection panics again, so the panic escapes the table package
+// and must be caught by the recover() in the newLevelsController goroutine.
+// The result is a graceful Open error, not a crash.
+func TestOpenWithTinySSTableReturnsError(t *testing.T) {
+	dir, opts, sstPath := writeDBWithSST(t)
+	defer removeDir(dir)
+
+	require.NoError(t, os.Truncate(sstPath, 2))
+
+	_, err := Open(opts)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "opening table")
+}
--- a/db_corrupt_test.go
+++ b/db_corrupt_test.go
@@ -91,10 +91,8 @@ func TestOpenWithCorruptedBlockDataSucceeds(t *testing.T) {
 }
 
 // TestOpenWithTinySSTableReturnsError truncates an SST below the 4-byte footer.
-// initIndex then reads at a negative offset and panics, and the recover defer's
-// debug collection panics again, so the panic escapes the table package
-// and must be caught by the recover() in the newLevelsController goroutine.
-// The result is a graceful Open error, not a crash.
+// initIndex now rejects it with a footer-validation error, which must surface
+// as a graceful Open error (not a crash) through the table package.
 func TestOpenWithTinySSTableReturnsError(t *testing.T) {
 	dir, opts, sstPath := writeDBWithSST(t)
 	defer removeDir(dir)
@@ -103,5 +101,5 @@ func TestOpenWithTinySSTableReturnsError(t *testing.T) {
 
 	_, err := Open(opts)
 	require.Error(t, err)
-	require.Contains(t, err.Error(), "opening table")
+	require.Contains(t, err.Error(), "Data corrupted")
 }
--- a/levels.go
+++ b/levels.go
@@ -14,6 +14,7 @@ import (
 	"math"
 	"math/rand"
 	"os"
+	"runtime/debug"
 	"sort"
 	"strings"
 	"sync"
@@ -120,7 +121,7 @@ func newLevelsController(db *DB, mf *Manifest) (*levelsController, error) {
 			var rerr error
 			defer func() {
 				if r := recover(); r != nil {
-					rerr = fmt.Errorf("error opening table %s: %v", fname, r)
+					rerr = fmt.Errorf("error opening table %s: %v\n%s", fname, r, debug.Stack())
 				}
 				throttle.Done(rerr)
 				numOpened.Add(1)
--- a/levels.go
+++ b/levels.go
@@ -126,6 +126,15 @@ func newLevelsController(db *DB, mf *Manifest) (*levelsController, error) {
 				throttle.Done(rerr)
 				numOpened.Add(1)
 			}()
+			// tables is sized by opt.MaxLevels, and nothing upstream constrains
+			// the level recorded in the manifest, so reject an out-of-range level
+			// here rather than letting the append below panic.
+			if int(tf.Level) >= len(tables) {
+				rerr = fmt.Errorf(
+					"manifest records table %s at level %d, but MaxLevels is %d",
+					fname, tf.Level, len(tables))
+				return
+			}
 			dk, err := db.registry.DataKey(tf.KeyID)
 			if err != nil {
 				rerr = y.Wrapf(err, "Error while reading datakey")
@@ -153,9 +162,11 @@ func newLevelsController(db *DB, mf *Manifest) (*levelsController, error) {
 				return
 			}
 
-			mu.Lock()
-			tables[tf.Level] = append(tables[tf.Level], t)
-			mu.Unlock()
+			func() {
+				mu.Lock()
+				defer mu.Unlock()
+				tables[tf.Level] = append(tables[tf.Level], t)
+			}()
 		}(fname, tf)
 	}
 	if err := throttle.Finish(); err != nil {
--- a/levels.go
+++ b/levels.go
@@ -119,6 +119,9 @@ func newLevelsController(db *DB, mf *Manifest) (*levelsController, error) {
 		go func(fname string, tf TableManifest) {
 			var rerr error
 			defer func() {
+				if r := recover(); r != nil {
+					rerr = fmt.Errorf("error opening table %s: %v", fname, r)
+				}
 				throttle.Done(rerr)
 				numOpened.Add(1)
 			}()
--- a/table/initindex_test.go
+++ b/table/initindex_test.go
@@ -50,30 +50,46 @@ func TestInitIndexFooterValidation(t *testing.T) {
 	require.NoError(t, err)
 	checksumLen := uint32(len(validChecksum))
 
-	// Note: checksumLen/indexLen are read via y.BytesToU32 (uint32) and cast to
-	// int, so on a 64-bit platform they can never be negative. The "< 0" half
-	// of the `<= 0` guard is therefore unreachable from a real file; the "= 0"
-	// cases below exercise the reachable branch of the same condition.
+	// The footer is read by walking backwards from EOF (readPos starts at
+	// tableSize), so each length must fit in the bytes remaining *before*
+	// readPos — not merely be smaller than the whole table. checksumLen == 0 is
+	// legal (a zero checksum marshals to no bytes), so only the bounds are
+	// checked for checksumLen; a zero indexLen is always corruption since a
+	// table always has at least one block.
 	tests := []struct {
 		name    string
 		data    []byte
 		wantErr string
 	}{
 		{
-			name:    "checksumLen=0",
-			data:    buildFooterData(0, nil, 0, nil),
+			name:    "tableSize<4",
+			data:    make([]byte, 2),
+			wantErr: "invalid table size in footer. Data corrupted",
+		},
+		{
+			// 8-byte file since checksumLen and inddexLen are both unit32.
+			// checksumLen field reads 8, exceeding the 4 bytes
+			// remaining before readPos.
+			name:    "checksumLen=tableSize",
+			data:    buildFooterData(0, nil, 8, nil),
 			wantErr: "invalid checksum length in footer. Data corrupted",
 		},
 		{
-			// 8-byte file: [indexLen=0:4][checksumLen=8:4] => checksumLen == tableSize.
-			name:    "checksumLen=tableSize",
-			data:    buildFooterData(0, nil, 8, nil),
+			name:    "checksumLen=tableSize-1",
+			data:    buildFooterData(0, nil, 7, nil),
 			wantErr: "invalid checksum length in footer. Data corrupted",
 		},
 		{
 			name:    "checksumLen>tableSize",
 			data:    buildFooterData(0, nil, 100, nil),
 			wantErr: "invalid checksum length in footer. Data corrupted",
+		},
+		{
+			// checksumLen=0 is legal, so it must fall through to the empty
+			// index length related error.
+			name:    "checksumLen=0 falls through to indexLen",
+			data:    buildFooterData(0, nil, 0, nil),
+			wantErr: "invalid index length in footer. Data corrupted",
 		},
 		{
 			name:    "indexLen=0",
--- a/table/initindex_test.go
+++ b/table/initindex_test.go
@@ -67,7 +67,7 @@ func TestInitIndexFooterValidation(t *testing.T) {
 			wantErr: "invalid table size in footer. Data corrupted",
 		},
 		{
-			// 8-byte file since checksumLen and inddexLen are both unit32.
+			// 8-byte file since checksumLen and indexLen are both uint32.
 			// checksumLen field reads 8, exceeding the 4 bytes
 			// remaining before readPos.
 			name:    "checksumLen=tableSize",
--- a/table/initindex_test.go
+++ b/table/initindex_test.go
@@ -1,0 +1,126 @@
+/*
+ * SPDX-FileCopyrightText: © 2017-2025 Istari Digital, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package table
+
+import (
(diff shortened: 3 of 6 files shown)


adityeah8969:
The PR is **ready**, its been auto pushed to DRAFT as repository does not allow more than 3 open PRs from a contributor.

table/table.go
```diff
@@ -423,8 +420,8 @@ func (t *Table) initIndex() (*fb.BlockOffset, error) {
 	readPos -= 4
 	buf := t.readNoFail(readPos, 4)
 	checksumLen := int(y.BytesToU32(buf))
-	if checksumLen < 0 {
-		return nil, errors.New("checksum length less than zero. Data corrupted")
+	if checksumLen <= 0 || checksumLen >= t.tableSize {
+		return nil, errors.New("invalid checksum length in footer. Data corrupted")
```
matthewmcneely: These bound the lengths against `t.tableSize`, but both reads walk backwards from EOF, so the constraint you actually need is the number of bytes remaining before `readPos`. Both bands are still reachable:

```
checksumLen = tableSize-1                 (passes: 0 < N-1 < N)
  -> readPos = N-4-(N-1) = -3             -> panic: slice bounds out of range [-3:]

tableSize=50, checksumLen=2, indexLen=48  (passes: 48 < 50)
  -> readPos = 50-8-2-48 = -8             -> panic: slice bounds out of range [-8:]
```

I reproduced both by calling `initIndex` directly on a crafted footer. The `indexLen` band at line 439 is the wider of the two: anything above `tableSize-8-checksumLen` overruns, and that sits well below `tableSize`.

What you want is a check that `readPos` stays `>= 0` after each subtraction, so `checksumLen > t.tableSize-4` here, and `t.indexLen > readPos` at line 439.

table/table.go
```diff
@@ -423,8 +420,8 @@ func (t *Table) initIndex() (*fb.BlockOffset, error) {
 	readPos -= 4
 	buf := t.readNoFail(readPos, 4)
 	checksumLen := int(y.BytesToU32(buf))
-	if checksumLen < 0 {
-		return nil, errors.New("checksum length less than zero. Data corrupted")
+	if checksumLen <= 0 || checksumLen >= t.tableSize {
```
matthewmcneely: Separate problem with the same line: `checksumLen == 0` is legal for a healthy table.

`Checksum_CRC32C` is enum value `0`, and the builder writes `pb.Checksum{Sum: crc32c(index), Algo: CRC32C}` (`table/builder.go:456-467`). Under proto3 an all-zero message marshals to nothing:

```
Sum=0 Algo=CRC32C -> len(marshal)=0
Sum=1 Algo=CRC32C -> len(marshal)=2
```

So a table whose index CRC happens to be zero writes `checksumLen == 0`, and this rejects it as corrupt. Around 2^-32 per table, so not urgent, but the failure mode is reporting healthy data as corruption, which is about the worst kind of wrong answer a storage engine can give.

The false pass you're closing is really "an empty index satisfies `0 == 0`", and the `indexLen == 0` check at line 439 covers that on its own, since a real table always has at least one block. I'd drop the `<= 0` half here and keep the bounds fix.

table/table.go
```diff
@@ -383,10 +379,11 @@ func (t *Table) initBiggestAndSmallest() error {
 			t.indexStart = readPos
 			indexData := t.readNoFail(readPos, t.indexLen)
 			fmt.Fprintf(&debugBuf, "index: %v ", indexData)
+
+			err = fmt.Errorf("initIndex crashed: %v\n%s", r, debugBuf.String())
```
matthewmcneely: This assignment is conditional on every debug read above it succeeding, and the messiest corruption is exactly when they don't.

Line 367 re-reads the footer with the unvalidated `checksumLen` straight off disk. Feed it the `checksumLen = tableSize-1` footer from the other thread and that read panics inside this deferred function, so this line never runs and the panic leaves `OpenTable`:

```
panic ESCAPED initBiggestAndSmallest: runtime error: slice bounds out of range [-3:]
```

That is what the comment you deleted was flagging (`// Use defer for printing info because there may be an intermediate panic`). The old code re-panicked deliberately, but the nested defer guaranteed the debug info got attached either way. Setting `err` first, and giving the debug reads their own `recover()`, keeps both properties:

```go
if r := recover(); r != nil {
    var debugBuf bytes.Buffer
    defer func() {
        _ = recover() // intermediate panic while gathering debug info
        err = fmt.Errorf("initIndex crashed: %v\n%s", r, debugBuf.String())
    }()
    ... existing debug reads ...
}
```

Unrelated and pre-existing, while you're in here: line 378 subtracts `t.indexLen` while line 374 reads `indexLen` into a local, so the debug output can describe a different region than the one it prints.

levels.go
```diff
@@ -119,6 +119,9 @@ func newLevelsController(db *DB, mf *Manifest) (*levelsController, error) {
 		go func(fname string, tf TableManifest) {
 			var rerr error
 			defer func() {
+				if r := recover(); r != nil {
+					rerr = fmt.Errorf("error opening table %s: %v", fname, r)
+				}
```
matthewmcneely, line 125: Two things worth tightening.

`%v` on the recovered value gives you one line with no stack. This now swallows every panic in the table-open path, including genuine bugs in `registry.DataKey` or `buildTableOptions`, so a `debug.Stack()` in the message, or an `Errorf` log alongside it, earns its bytes.

The net is also one-sided. `OpenInMemoryTable` runs the same `initBiggestAndSmallest`, and its three callers (`db.go:1136`, `levels.go:905`, `stream_writer.go:486`) have no equivalent guard. All three work on freshly built data, so the risk is low, but it's worth knowing.

db_corrupt_test.go
```diff
@@ -0,0 +1,107 @@
...
+// The result is a graceful Open error, not a crash.
+func TestOpenWithTinySSTableReturnsError(t *testing.T) {
+	dir, opts, sstPath := writeDBWithSST(t)
+	defer removeDir(dir)
+
+	require.NoError(t, os.Truncate(sstPath, 2))
+
+	_, err := Open(opts)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "opening table")
```
matthewmcneely: `"opening table"` only appears in the `levels.go` recover this PR adds, so this asserts that the backstop caught a panic rather than that the table package handled it. The gap from the `initBiggestAndSmallest` thread is showing up here as a test expectation.

Once the debug-collection `recover()` is in, this can assert on the table-package error, and the `levels.go` recover goes back to being a net instead of the mechanism.

table/table.go
```diff
@@ -331,15 +331,11 @@ func OpenInMemoryTable(data []byte, id uint64, opt *Options) (*Table, error) {
 	return t, nil
 }
 
-func (t *Table) initBiggestAndSmallest() error {
+func (t *Table) initBiggestAndSmallest() (err error) {
```
matthewmcneely, line 336: Anchoring here because this signature change is what activates the problem, though the line itself is fine.

Lines 297-299 are the only error return in `OpenTable` that doesn't call `mf.Close(-1)`; the `VerifyChecksum` branch just below does. That was academic while this function panicked instead of returning. It's now the main corrupt-table exit, and fds accumulate across failed `Open` calls:

```
25 failed Opens          -> sst fds leaked: 7
same, with mf.Close(-1)  -> sst fds leaked: 0
```

`closeAllTables` in `levels.go` only closes tables that opened successfully, so nothing downstream picks this up. Of everything I found here this is the one with real Dgraph exposure, since a caller that retries `Open` against a corrupt directory keeps accumulating mappings.

table/table.go
```diff
@@ -439,6 +436,9 @@ func (t *Table) initIndex() (*fb.BlockOffset, error) {
 	readPos -= 4
 	buf = t.readNoFail(readPos, 4)
 	t.indexLen = int(y.BytesToU32(buf))
+	if t.indexLen <= 0 || t.indexLen >= t.tableSize {
+		return nil, errors.New("invalid index length in footer. Data corrupted")
+	}
 
 	// Read index.
 	readPos -= t.indexLen
```
matthewmcneely, line 467: Anchoring here for proximity; the finding is line 472 below, and it's relevant to the PR title.

`y.AssertTrue` and `y.Check` both call `log.Fatalf`, which is `os.Exit`. No `recover()` anywhere catches those, so the net added in `levels.go` has a hole by construction, and `readNoFail` is built on `y.Check`.

`y.AssertTrue(index.Offsets(&bo, 0))` at 472 is still on this path. An index whose checksum is self-consistent and which parses as a valid flatbuffer with no offsets vector takes the process down:

```
2026/08/18 18:24:15 Assert failed
FAIL   github.com/dgraph-io/badger/v4/table
```

Hard to hit accidentally, since the corrupt index has to match its own stored checksum. But it's the same class as #2201, and returning an error instead of asserting is a one-liner. Your call whether that belongs here or in a follow-up.

matthewmcneely (changes requested):

table/table.go
adityeah8969: Addressed!

table/table.go
adityeah8969: Addressed!

I thought of 'checksumLen==0' as invalid earlier with same reasoning of a valid table having at least one block inside. But it does make sense to let this error get reported later while checking index.

table/table.go
adityeah8969: Addressed both defer as well as pre-existing bug.

levels.go
adityeah8969: Addressed!

db_corrupt_test.go
adityeah8969: Addressed!

table/table.go
adityeah8969: Addressed!

table/table.go
adityeah8969: Addressed, I have made changes to return error instead of assert.

adityeah8969:
@matthewmcneely I have addressed all review comments, please take a look at your convenience.

Also thanks a ton for descriptive comments, those were really helpful.

table/table.go
```diff
@@ -420,11 +427,16 @@ func (t *Table) initIndex() (*fb.BlockOffset, error) {
...
+		return nil, errors.New("invalid table size in footer. Data corrupted")
+	}
 	readPos -= 4
 	buf := t.readNoFail(readPos, 4)
 	checksumLen := int(y.BytesToU32(buf))
-	if checksumLen < 0 {
-		return nil, errors.New("checksum length less than zero. Data corrupted")
+	// checksumLen == 0 is legal (a zero checksum marshals to nothing), so only
+	// reject lengths that don't fit in the bytes remaining before readPos.
+	if checksumLen > readPos {
```
matthewmcneely: This needs the `< 0` half back. We're trying to keep 32-bit platforms buildable, and dropping it leaves this as the one footer field that isn't safe there.

`y.BytesToU32` returns a `uint32`, and `int(...)` on a 32-bit build wraps anything at or above 2^31 to a negative `int`. A negative `checksumLen` satisfies `checksumLen > readPos` as false, so it passes this check, and then `readPos -= checksumLen` moves `readPos` *forward*, past EOF. `t.readNoFail(readPos, checksumLen)` then panics slicing at an offset beyond `len(m.Data)`.

`t.indexLen <= 0` kept its zero-and-below clause, so the index field is already covered. This is just restoring the symmetry, and the pre-PR code had it:

```suggestion
	if checksumLen < 0 || checksumLen > readPos {
```

Int width is a live concern in this code rather than a hypothetical: this package already carries `intSize = int(unsafe.Sizeof(int(0)))` at line 37, and `options.go:164` documents a 32-bit overflow workaround.

table/table.go
```diff
@@ -295,6 +295,7 @@ func OpenTable(mf *z.MmapFile, opts Options) (*Table, error) {
 	t.ref.Store(1)
 
 	if err := t.initBiggestAndSmallest(); err != nil {
+		_ = mf.Close(-1)
```
matthewmcneely, line 299: This is right, thanks. One sibling left over: the `BlockSize == 0 && Compression != None` early return at lines 271-273 is now the only error path in `OpenTable` that doesn't release the mapping.

Same leak you just fixed here, and it contradicts the contract you documented over in `builder_test.go` ("OpenTable takes ownership of the mmap it is given"). It's reachable for anyone running compression with a zero `BlockSize`, and it leaks a mapping on every table open. Worth an `mf.Close(-1)` there too while this is open.

Minor, and only because I'm already on the line: the three sibling calls at 276, 283 and 303 are bare `mf.Close(-1)` with no `_ =`. Either form is fine, but matching them saves the next reader a second of wondering why this one is different.

table/initindex_test.go
```diff
@@ -0,0 +1,142 @@
...
+		data    []byte
+		wantErr string
+	}{
+		{
+			name:    "tableSize<4",
+			data:    make([]byte, 2),
+			wantErr: "invalid table size in footer. Data corrupted",
+		},
+		{
+			// 8-byte file since checksumLen and inddexLen are both unit32.
```
matthewmcneely: Two typos here: "inddexLen" and "unit32".

Worth fixing rather than leaving in the tree. This is the comment block explaining the validation rules, so it's the thing someone reads first when they come back to this code, and anyone grepping for `indexLen` won't find it.

```suggestion
			// 8-byte file since checksumLen and indexLen are both uint32.
```

matthewmcneely (changes requested):

table/table.go
adityeah8969: makes sense, Addressed!

table/table.go
adityeah8969: Addressed both concerns!

Please do note not having '_ = ' in 'mf.Close()' threw lint errors as part of trunk checks on git push.    
`272:11  high  Error return value of `mf.Close` is not checked  golangci-lint2/errcheck`

For maintaining consistency with other sibling calls, I have removed '_ ='.

There is another (unrelated to our change in this PR) trunk format error that I am getting after latest rebase.

```
docs/quickstart.md
 -:-  fmt  Incorrect formatting, autoformat by running 'trunk fmt'  prettier
```

I could fix these in next PR.

table/initindex_test.go
adityeah8969: Oops, thanks for pointing these out, Addressed!

adityeah8969:
@matthewmcneely All new comments have been addressed, please take a look.

matthewmcneely:
Hey @adityeah8969 I pushed a few formatting fixes to your branch. All green now. Can you set this PR as "Ready for review"?

matthewmcneely:
@adityeah8969 I understand your PR limit is met. Can you set one of the others to DRAFT so that this one opens up?

adityeah8969:
@matthewmcneely This is marked "read for review" so that it could be merged.

matthewmcneely:
Hey @adityeah8969, thanks for sticking with this one. Five rounds is more than it should have taken, and the last few are on me for surfacing things one layer at a time. Rather than send you back again I've pushed the fix myself in 912f160, so you don't have to context-switch into it. Have a look, and say the word if you'd rather I backed it out and left it to you.

The finding: the `recover()` in the table-opening goroutine can fire while `mu` is held. The only statement in that critical section is `tables[tf.Level] = append(...)`, and `tables` is sized `opt.MaxLevels` while nothing upstream constrains the level the manifest records. So a manifest with a table at a level at or beyond `MaxLevels` panics with index-out-of-range, the recover catches it, and `mu` is never released. Every remaining goroutine then parks on `mu.Lock()`, `throttle.Finish()` never completes, and `db.Open()` blocks forever.

Reproduced with 3 tables moved to level 6 and `MaxLevels(3)`: Open hung indefinitely with 2 goroutines parked in `sync.(*Mutex).Lock`. With 4 tables it does surface the error through `throttle.Do()`, but still leaves 2 goroutines and their open mmaps parked on the leaked mutex for the life of the process.

Why this was worth settling before merge rather than as a follow-up, and it is a Dgraph-shaped reason: on `main` this same input dies immediately with a stack trace naming the line. With the recover in place it hangs silently instead. For a Dgraph alpha those are very different failures. A crash hands an operator a stack in the logs, a non-zero exit, and a visible restart loop. A hang hands them an alpha that never reports ready, with no error, log line, or stack to grep for. That is the difference between a few minutes of diagnosis and an afternoon of it, and it is the one case where this PR would have left an operator worse off than before.

To be clear, this is not a hole in your change. The `recover()` is the right addition, and the footer validation is what makes the whole thing safe. This is a second-order interaction with a pre-existing unguarded critical section, and I read those same three lines in three separate rounds without seeing it.

What I pushed:

- Reject an out-of-range `tf.Level` up front, so a bad manifest level becomes a clear error and no mmap is opened for it.
- Take and release `mu` in a closure with `defer`, so no future panic in that region can leave it held.
- A regression test that hangs and fails without the fix, and passes in 0.2s with it.

Verified locally: full root suite green (249s), `-race` clean, and `trunk check` reports no new issues.

Two smaller things I turned up that I would rather not hold this PR for. Happy to file them as issues, or to take them here if you would rather keep going:

- `table/table.go:736` dereferences `b.offset` inside `if err != nil`, but `t.block()` returns a nil block alongside its error. `OpenTable` reaches it whenever `ChecksumVerificationMode` is `OnTableRead` or `OnTableAndBlockRead`, so the real checksum message gets replaced by a nil deref and the `mf.Close(-1)` on that path never runs. The same line reaches the public `db.VerifyChecksum()`, where nothing recovers it.
- The debug dump at `table/table.go:394` formats the whole index region with `%v`, so the error string runs roughly 4x `indexLen`. A pristine encrypted DB reopened with the default `IndexCacheSize` of 0 gave an 88KB error, which `db.Open`'s own cleanup defer then logs at ERROR. Capping it to the first 64 bytes plus the length would settle it.

Once CI is green on 912f160 I am happy to approve. Thanks again for the patience here, and for the care in the earlier rounds: the footer validation held up against an 80k-case sweep of malformed footers without a single escaping panic, which is exactly what this needed.

table/table.go
matthewmcneely: No worries, I pushed a commit to your branch for these.

matthewmcneely (approved):

adityeah8969:
@matthewmcneely thanks a lot for reviewing, final commit, merge and descriptive comments. It seems like there have been some edge cases that I failed to notice and will keep a note of this for future PRs. 

Regarding your comment [here](https://github.com/dgraph-io/badger/pull/2333#issuecomment-5445993861)
_Have a look, and say the word if you'd rather I backed it out and left it to you._

Absolutely, please do not hesitate to ask for cleanups, I will be more than happy to re-address concerns (or nits) in case earlier rounds of review did not take them into a/c.

